PR #13, opened 2018-07-08
GAE deployment

### LINKED ISSUE #3: GAE deployment and production app management
Ported from @matthewroot's issue https://github.com/Recidiviz/recidiviz.com/issues/23 into this new repo.

https://github.com/Recidiviz/recidiviz.com/issues/22 rebuilt recidiviz.com into a single page application using React. Currently this application is not deployable to GAE and also has no way to manage the application in production.

The app uses `webpackDevServer` to serve the app in development but we must use something more robust in production. [PM2](http://pm2.keymetrics.io/) is one viable option here. As part of this issue, a yarn script should be added to start the app in production, at a minimum.

The app will also need to be configured for deployment to GAE. @jessex can add additional details on recommended approach.

**Acceptance Criteria**
1. Application can be deployed to GAE
1. ~~Able to manage application in production~~
1. ~~Yarn script added for starting application in production~~

matthewroot:
Resolves #3.

This PR adds:
1. Configuration for deploying to GAE production service
1. Configuration for deploying to GAE staging service
1. Yarn scripts for executing deployment to either prod or staging


matthewroot:
This PR does not currently address hooking into our Travis pipeline. Wanted to get your thoughts on whether or not it makes sense to go ahead and do that as part of this PR, @jessex. We don't have an open issue for this in `pulse-dashboard` but would be similar to [#71](https://github.com/Recidiviz/recidiviz-data/issues/71) on `recidiviz-data`. I don't love providing the ability for anyone to be able to deploy to prod and staging although I've tried to make it more straightforward by adding scripts to limit the potential for screwing up.

It would be trivial to configure Travis to deploy to either `pulse-dashboard-staging` or `pulse-dashboard-prod` upon merging to master. My inclination would be for the former and then to have a manual or scheduled job than can execute a production deploy. We could schedule this to whatever interval we want (daily? hourly?) and limit access to a select few users. Alternatively, we could have Travis deploy straight to prod after a merge to master. If we do that, I'm not exactly sure how we'll use `pulse-dashboard-staging` right now.

#discuss

matthewroot:
Oh, also, you can see deployed apps at:

Prod: https://pulse-dashboard-prod-dot-recidiviz-123.appspot.com/
Staging: https://pulse-dashboard-staging-dot-recidiviz-123.appspot.com/

matthewroot:
One additional thought for you, @jessex. I'm thinking we should configure sub-domain(s) for `pulse-dashboard`, probably pulse-dashboard.recidiviz.com and pulse-dashboard-staging.recidiviz.com or just pulse.recidiviz.com. This will let us add version numbers when deploying while keeping things tidy and neat for both internal and external users who don't care about that. I'll need you (or Clem or Andrew) to do this step as I don't have access to wherever we registered the domain name and setting this up will require external verification with that service.

We don't necessarily need to do this now and might be advisable to wait until we get closer to launching `pulse-dashboard`. I can file a follow-on issue if that's your preference, @jessex.

pulse-dashboard-prod.yaml
```diff
@@ -0,0 +1,39 @@
+runtime: nodejs8
+
+service: pulse-dashboard-prod
+
+handlers:
```
jessex: Did you consider placing `handlers` in something like a `base.yaml` file and then including it in each tier's app.yaml file like so?
```
includes:
- base.yaml
```

The handlers sections are identical now and will remain identical for the foreseeable future, yes?

matthewroot: Hadn't thought of this. Good suggestion, will do!

matthewroot:
@jessex: made suggested update and also configured Travis to deploy to staging on merge to master. One thing that is really opaque, and I've poured over Travis docs for about an hour trying to find this, is whether or not it's possible to deploy to a specific service rather than the default service for the project. I'm hoping that using the `config` option in the `deploy` section of the Travis config will make this happen. I'm not sure how to test this other than merging this PR (or doing a straight push to master).

matthewroot:
You can see in the log that it picked up the GAE encrypted key and also that it skipped deployment because it's a PR build: https://travis-ci.org/Recidiviz/pulse-dashboard/jobs/401557132

.gcloudignore
```diff
@@ -23,3 +23,4 @@ archive/
 coverage/
 /yarn.lock
 /*.md
+recidiviz-gae-api-key.json.enc
```
jessex: Is it okay to check this file below into version control?

jessex:
There are certainly more convoluted ways of verifying that the deploy works, but I doubt they are worthwhile. Let's just test when we merge into master. We should coordinate with @cawarren, though, who has recidiviz-data running in some sort of staging environment, I believe.

.gcloudignore
matthewroot: Yes, this is the encrypted API key and is necessary for Travis to be able to deploy to GAE. The keys for decrypting this file are securely stored in Travis.

If you end up doing this for `recidiviz-data` or other projects, just be sure to only commit the encrypted version (.enc) to source control and not the plaintext JSON file.

matthewroot:
Talked to Andrew and we should be good to merge and test the auto deploy. Merging based on last comment from @jessex despite review still being open.

matthewroot:
Deploy to staging service confirmed successfully, w00t!

https://travis-ci.org/Recidiviz/pulse-dashboard/builds/402409676

jessex:
Great job, @matthewroot.

PR #14, opened 2018-07-17
Add enzyme

### LINKED ISSUE #4: Evaluate Enzyme
Ported from @matthewroot's issue https://github.com/Recidiviz/recidiviz.com/issues/26 into this new repo.

[Enzyme](https://airbnb.io/projects/enzyme/) has become something of a standard for improving the unit testing story of applications built with React. We should evaluate Enzyme and, if we like what it provides, add it to recidiviz.com. Additionally, at least one example of a unit test that utilizes Enzyme should also be provided.

**Acceptance Criteria**
1. Evaluate Enzyme and determine whether or not we want to use it
1. Add one test, or update an existing test, leveraging some capability of Enzyme

matthewroot:
Resolves #4.

After spending some time getting familiar with Enzyme, I have to say that I dig it and get why it's become so popular. It should be a good fit for us to enable easy unit testing of our frontend that we'll build in React. An alternative would be to use Facebook's [library](https://reactjs.org/docs/test-renderer.html) but the API isn't as friendly as Enzyme's.

This PR adds Enzyme, the necessary React 16 adapter, a set of simple tests utilizing Enzyme, a new Jest snapshot test for the core App component, and updates the README with some information about testing with Enzyme + Jest.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -1,7 +1,8 @@
 {
   "extends": "airbnb",
   "rules": {
-    "react/jsx-filename-extension": ['error', { extensions: ['.js'] }]
+    "react/jsx-filename-extension": ['error', { extensions: ['.js'] }],
+    "import/no-extraneous-dependencies": ["error", {"packageDir": './'}]
   },
   "env": {
     "browser": true,
--- a/src/components/App/__snapshots__/App.test.js.snap
+++ b/src/components/App/__snapshots__/App.test.js.snap
@@ -1,0 +1,53 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`<App /> matches saved snapshot 1`] = `
+<div
+  className="app-container"
+>
+  <Header
+    contactEmail="team@recidiviz.com"
+    title="Recidiviz"
+  />
+  <Hero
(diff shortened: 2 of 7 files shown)


src/components/App/__snapshots__/App.test.js.snap
```diff
@@ -0,0 +1,53 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
```
matthewroot: Yes, these snapshot files should be committed.

.eslintrc
```diff
@@ -1,7 +1,8 @@
 {
   "extends": "airbnb",
   "rules": {
-    "react/jsx-filename-extension": ['error', { extensions: ['.js'] }]
+    "react/jsx-filename-extension": ['error', { extensions: ['.js'] }],
+    "import/no-extraneous-dependencies": ["error", {"packageDir": './'}]
```
matthewroot: Necessary to make eslint not upset about the added Enzyme config in `jestSetup.js`.

src/components/App/__snapshots__/App.test.js.snap
jessex: Oof, fine.

PR #23, opened 2019-07-23
Ports the app to React

jessex:
Ends our short-lived tenure as a Node/Express server-rendered web app in favor of a hopefully long-lived reign as a React SPA with a small Node/Express API backend.

Incorporates Julia's content upgrades based on feedback from our initial users.


jessex:
Merging as discussed offline with @jjdressel11! Several follow-up issues to be filed shortly, but I will definitely welcome feedback on this PR after merging.

PR #42, opened 2019-07-26
Fixes exports of chart images and data in the MS browsers.

### LINKED ISSUE #39: Fix broken exports in MS browsers
Exports appear to be broken in Microsoft Edge and IE11. In both, _Export Image_ appears to not even register. In the former, _Export Data_ brings up a small dialog at the bottom of the window saying the file could not be downloaded. In the latter, _Export Data_ redirects to a new tab with the json in the browser bar, escaped, producing an error.

jessex:
This closes #39. Replaces my artisan download code with a dedicated [library](https://github.com/rndme/download), which is how it should be.

Tested image and json exports successfully in multiple browsers in stage (and production, for the Friday morning release), including Edge and IE11.

### DIFF
--- a/src/components/TopBar.js
+++ b/src/components/TopBar.js
@@ -45,7 +45,7 @@ const TopBar = (props) => {
             <li className="dropdown">
               <a href="#" onClick={() => loginWithRedirect({appState: { targetUrl: "/snapshots" }})} className="dropdown-toggle no-after peers fxw-nw ai-c lh-1" data-toggle="dropdown">
                 <div className="peer mR-10">
-                  <i class="ti-power-off"></i>
+                  <i className="ti-power-off"></i>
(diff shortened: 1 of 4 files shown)


src/components/TopBar.js
```diff
@@ -45,7 +45,7 @@ const TopBar = (props) => {
             <li className="dropdown">
               <a href="#" onClick={() => loginWithRedirect({appState: { targetUrl: "/snapshots" }})} className="dropdown-toggle no-after peers fxw-nw ai-c lh-1" data-toggle="dropdown">
                 <div className="peer mR-10">
-                  <i class="ti-power-off"></i>
+                  <i className="ti-power-off"></i>
```
jessex: This was just something I noticed.

Issue #28, opened 2019-07-24
Make the backend API live


jessex:
The frontend retrieves the calculations from the backend server, but the server is hard-coding static values for now. Our design, in the interests of performance and simplicity, since the calculations are updated daily at best, is to have the calculation pipelines write data in an expected format to json filed stored in Google Cloud Storage. The backend server will periodically retrieve the latest version of those files and cache them on the server, including at app startup. It will then serve the cached values directly to the frontend. If more performance improvement is necessary, we can tweak the `cache-control` headers on the responses from the backend.

We need to:
* Design the actual API
* Write json files in a private bucket Cloud Storage after calculation pipelines, for some explicit subset of generated metrics
* Store the secret credentials to access that bucket in environment variables on the backend server
* Have the server pull those down on startup and periodically (hourly with a scheduled task?), and cache them in a format suitable for the frontend
* Update the frontend components that talk to the API to match updated structure, and to extract generalizable code copied in the components

jessex:
We have most of the data plumbing to get files to Cloud Storage complete at this point. The app backend has the fetching, caching, and serving of metrics to the frontend. Now we need to wire up the visualizations to use the updated API spec and deprecate the previous, temporary, hard-coded API stub. We will also want to ensure that the API can fetch data for specific states in the process and not send over data for other states in any API requests.

PR #44, opened 2019-07-29
Adds fetch, caching, and background refresh of metrics files from GCS

jessex:
This goes towards, but does not complete, #28. This includes:
* API endpoints on the backend for fetch admission, reincarceration, and revocation metrics
* Fetching of metrics from pre-processed json files in GCS, organized by general metric variety (the 3 aforementioned types)
* Caching of those metrics in memory so that each call does not round-trip to GCS
* Background refresh of that cache so that only rare calls actually invoke a round-trip to GCS
* Authentication against GCS via service accounts, whose credentials are stored in secret files referenced by environment variables

This does not yet hook up these metrics to any visualizations.

Tested with lots of local manual testing. Requests from the browser to the API when the metrics are in the cache lead to a cached 304 response in the sub-millisecond range. The logs show the background refresh executing on schedule, hitting GCS to refresh the cache only after the cache TTL has expired.


.gitignore
```diff
@@ -27,6 +27,8 @@ auth_config.json
 auth_config*.json
 gae*.yaml
 src/assets/scripts/metrics/*.json
+recidiviz-production-*.json
+recidiviz-staging-*.json
```
jjdressel11: Suggestion to change these files to start with `recidiviz-dev-` instead so we don't have the name of our GCP project in these files?

package.json
```diff
@@ -13,7 +13,7 @@
     "test": "react-scripts test",
     "spa": "react-scripts start",
     "server": "node server.js",
-    "server:dev": "AUTH_ENV=development nodemon server.js",
+    "server:dev": "AUTH_ENV=development GOOGLE_APPLICATION_CREDENTIALS=./recidiviz-staging-service-account.json nodemon server.js",
```
jjdressel11: Per my previous comment, this would be `recidiviz-dev-service-account.json`

server/core/metricsApi.js
```diff
@@ -0,0 +1,94 @@
...
+ * In the current implementation, metrics are stored in pre-processed json files in Google Cloud
+ * Storage. Those files are pulled down and cached in memory with a TTL. That TTL is unaffected by
+ * access to the cache, so files are re-fetched at a predictable cadence, allowing for updates to
+ * those files to be quickly reflected in the app without frequent requests to GCS.
+ */
+
+var objectStorage = require('./objectStorage');
+var cacheManager = require('cache-manager');
+
+const BUCKET_NAME = 'recidiviz-staging-dashboard-data';
```
jjdressel11: Is this a temporary solution until we are accessing the bucket in prod? Or will we have the prod bucket in plain text in prod?

.gitignore
jessex: Good callout, I'll change that.

package.json
jessex: Will change!

server/core/metricsApi.js
jessex: Ah you're right, I meant to extract that to an environment variable before submitting the PR. I'll hide this behind an environment variable, and we can rename the bucket if we'd like for security's sake.

PR #45, opened 2019-07-31
Chart API sharing

jessex:
This goes towards #28 but does not complete it.

This change makes it so that API calls to the backend originate from the view logic, e.g. `views/reincarcerations`, instead of from each individual chart component. This is not only a nice reduction in duplication and thinning of the chart contract, but it is important in that it will align much neater with the live API calls when we are ready to cut over.

The pattern here is that all of the `fetchChartData` calls have been extracted out of the many `src/components/charts` files into the 4 `src/views` files. Those files then break apart the larger API response to pass down portions to individual charts, where chart-specific processing continues.

### DIFF
--- a/src/components/charts/reincarcerations/ReincarcerationCountOverTime.js
+++ b/src/components/charts/reincarcerations/ReincarcerationCountOverTime.js
@@ -3,47 +3,26 @@ import React, { useState, useEffect } from "react";
 import { Line } from 'react-chartjs-2';
 import { configureDownloadButtons } from "../../../assets/scripts/charts/chartJS/downloads";
 import { COLORS } from "../../../assets/scripts/constants/colors";
-import { useAuth0 } from "../../../react-auth0-spa";
 
-const ReincarcerationCountOverTime = () => {
+const ReincarcerationCountOverTime = props => {
   const [chartLabels, setChartLabels] = useState([]);
   const [chartDataPoints, setChartDataPoints] = useState([]);
-  const { getTokenSilently } = useAuth0();
 
-  const processResponse = (responseData) => {
-    const countsByMonth = responseData.reincarcerationCountsByMonth;
+  const processResponse = () => {
+    const countsByMonth = props.reincarcerationCountsByMonth;
 
-    var sortable = [];
+    var sorted = [];
     for (var month in countsByMonth) {
-        sortable.push([month, countsByMonth[month]]);
+        sorted.push([month, countsByMonth[month]]);
     }
 
-    return sortable;
+    setChartLabels(sorted.map(element => element[0]));
+    setChartDataPoints(sorted.map(element => element[1]));
   }
 
-  const fetchChartData = async () => {
-    try {
-      const token = await getTokenSilently();
-      // Likely needs to point to app engine URL
-      const response = await fetch(`${process.env.REACT_APP_API_URL}/api/external`, {
-        headers: {
-          Authorization: `Bearer ${token}`
-        }
-      });
-
-      const responseData = await response.json();
-      const sorted = processResponse(responseData);
-
-      setChartLabels(sorted.map(element => element[0]));
-      setChartDataPoints(sorted.map(element => element[1]));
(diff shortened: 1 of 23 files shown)


src/components/charts/reincarcerations/ReincarcerationCountOverTime.js
```diff
@@ -3,47 +3,26 @@ import React, { useState, useEffect } from "react";
 import { Line } from 'react-chartjs-2';
 import { configureDownloadButtons } from "../../../assets/scripts/charts/chartJS/downloads";
 import { COLORS } from "../../../assets/scripts/constants/colors";
-import { useAuth0 } from "../../../react-auth0-spa";
 
-const ReincarcerationCountOverTime = () => {
+const ReincarcerationCountOverTime = props => {
```
jjdressel11: Change to be `ReincarcerationCountsByMonth` (file name too).

src/components/charts/reincarcerations/ReincarcerationRateByTransitionalFacility.js
```diff
@@ -2,50 +2,29 @@ import React, { useState, useEffect } from "react";
...
+    const ratesByFacility = props.ratesByTransitionalFacility;
 
-    var sortable = [];
+    var sorted = [];
     for (var facility in ratesByFacility) {
-        sortable.push([facility, ratesByFacility[facility]]);
+        sorted.push([facility, ratesByFacility[facility]]);
     }
-    sortable.sort(function(a, b) {
+    sorted.sort(function(a, b) {
```
jjdressel11: Can you add a comment that this is sorting the facilities by the rates?

src/components/charts/reincarcerations/ReleasesVsAdmissions.js
```diff
@@ -2,47 +2,30 @@ import React, { useState, useEffect } from "react";
 
 import { Bar } from 'react-chartjs-2';
 import { COLORS_GOOD_BAD } from "../../../assets/scripts/constants/colors";
-import { useAuth0 } from "../../../react-auth0-spa";
 
-const ReleasesVsAdmissions = () => {
+const ReleasesVsAdmissions = (props) => {
```
jjdressel11: FYI the file for this will be "admissions_versus_releases_by_month.json" and there will be a column called "population_change" instead of a column for admissions and for releases. Not necessarily logic that needs to change now - we can change everything once we're using real files with real data.

src/components/charts/reincarcerations/ReincarcerationCountOverTime.js
jjdressel11: Actually let's just wait until we do the full integration with the real files and data before renaming everything.

src/views/ProgramEvaluation.js
```diff
@@ -144,7 +147,25 @@ const ProgramEvaluation = () => {
     setReportCardRevocation3(card.thirdYearRevocation);
   }
 
-  const { loading, user } = useAuth0();
+  const fetchChartData = async () => {
```
jjdressel11: I assume react makes it possible to reuse this `fetchChartData` function elsewhere? If yes, let's file an issue for making the tabs reuse functions where possible.

src/components/charts/reincarcerations/ReincarcerationRateByTransitionalFacility.js
jessex: Will add.

src/components/charts/reincarcerations/ReleasesVsAdmissions.js
jessex: I'll go ahead and change now to be proactive.

src/views/ProgramEvaluation.js
jessex: Yeah that will be easy. Will file an issue now.

jessex: Filed #46!

PR #47, opened 2019-08-15
New snapshots view

jjdressel11:
Implements the new Snapshots view that showcase the four main metrics ND is tracking.

In this PR:
- Updates to the gae.yaml.example file
- Updates to the dependencies to include a chartjs trendline plugin and eslint support
- A new chart component for each snapshot metric with dynamically loaded headers
- A lot of small eslint fixes
- A utils file to convert month numbers to string names

Not in this PR:
- Using api calls for the data in the snapshot charts
- Implementation of export buttons for snapshot charts

In the next PR, I'm going to fully hook up all charts to the API calls, and make sure everything that needs to be dynamic is dynamic.


jessex:
One high-level comment: the massive size of the PR is because `package-lock.json` is being added. I imagine that this got created for you by virtue of doing `npm install` instead of `yarn install`. As described [here](https://stackoverflow.com/questions/44552348/should-i-commit-yarn-lock-and-package-lock-json-files), we always want to commit at least one of `yarn.lock` or `package-lock.json` to master, but there's a choice as to whether to commit both. If you commit both, you probably want to have a process for keeping them in sync, since people who use `npm install` will update one but not the other, and vice versa, so when they make commits they'll push the two files out of sync.

I think we should standardize on using `yarn` over `npm`, as is implied in the readme, and maybe even add `package-lock.json` to the `.gitignore` to discourage its use a bit. What do you think?

jjdressel11:
> One high-level comment: the massive size of the PR is because `package-lock.json` is being added. I imagine that this got created for you by virtue of doing `npm install` instead of `yarn install`. As described [here](https://stackoverflow.com/questions/44552348/should-i-commit-yarn-lock-and-package-lock-json-files), we always want to commit at least one of `yarn.lock` or `package-lock.json` to master, but there's a choice as to whether to commit both. If you commit both, you probably want to have a process for keeping them in sync, since people who use `npm install` will update one but not the other, and vice versa, so when they make commits they'll push the two files out of sync.
> 
> I think we should standardize on using `yarn` over `npm`, as is implied in the readme, and maybe even add `package-lock.json` to the `.gitignore` to discourage its use a bit. What do you think?

Yes this definitely was a result of me using `npm` instead of `yarn` accidentally. I think we should enforce only using `yarn` and add `package-lock.json` to the `.gitignore` so that the build will break on Travis if the changes in the PR don't have an updated `yarn.lock` file with a necessary dependency if someone uses `npm`.

jessex:
I checked this out locally to do some manual testing and see if I could find any bugs or things to improve. Here's my list!
* The export buttons don't work for any of the new snapshots
* It looks like there's too much padding around the "Prison Admissions Due to Revocations" chart. I think it's because there is a div with `class="layer w-100 p-40"` where for the other charts it is `class="layer w-100 p-20"`.
* I'm not sure about the narrative text for "Successful completion of supervision." "Successfully completed their supervision by that time" reads a little vague, like it might be focused on _parole having not been extended_, as opposed to people not having been revoked. Were there more conversations about this language?
* Without any sense of year in the x-axis for these charts, it's not immediately obvious what time period you're looking at. I don't think the right move is to append the year to each x-axis label, but I wonder if we could do something where the year is applied to, for example, the left-most x-axis label (Mar '18) and to January (Jan '19). How does that sound?
* The y-axis description for the "Average Change in LSIR Scores" chart should start with a capital letter.
* I see we no longer have anything below the charts, including drivers, or "Dive into" links, or methodology. That's all expected?

src/assets/scripts/constants/colors.js
```diff
@@ -289,43 +290,43 @@ const GREYS = {
...
+  '#ffd75e',
 ];
 
 const COLORS_FIVE_VALUES_LIGHT = [
   '#495259',
-  '#943CA6',
-  '#F86A89',
-  '#FDB0A8',
-  '#FDE394',
+  '#943ca6',
```
jessex: Just curious: did linting or something pick this up, or did something new break because of the casing?

src/components/charts/snapshots/DaysAtLibertySnapshot.js
```diff
@@ -0,0 +1,175 @@
...
+
+            drawTime: 'afterDatasetsDraw',
+
+            borderColor: COLORS['red-400'],
+            borderWidth: 2,
+            borderDash: [2, 2],
+            borderDashOffset: 5,
+            label: {
+              enabled: true,
+              content: 'goal: 1095 days (3 years)',
```
jessex: Eventually we'll want to add a separate `/api/goals` endpoint that lets us save and retrieve goals for specific metrics. But hard-coded is right for now!

src/components/charts/snapshots/DaysAtLibertySnapshot.js
```diff
@@ -0,0 +1,175 @@
...
+  );
+
+  // TODO: Change this to have export for days at liberty data
+  const exportedStructureCallback = function exportedStructureCallback() {
+    return {};
+  };
+  configureDownloadButtons('days-at-liberty', 'Snapshot', chart.props,
+    document.getElementById('days-at-liberty-snapshot-chart'), exportedStructureCallback);
+
+  const header = document.getElementById('daysAtLibertySnapshot-header');
```
jessex: Pursuant to our convo yesterday, I think a good compromise in terms of coupling would be to pass the id of the dynamic header element to be updated as a parameter to the chart component. Then it becomes part of the conceptual contract of this component that it will be making dynamic modifications to the header, and this component doesn't have to magically know the id of the header it wants to change.

src/components/charts/snapshots/LSIRScoreChangeSnapshot.js
```diff
@@ -0,0 +1,175 @@
...
+        datasets: [{
+          backgroundColor: COLORS['blue-standard'],
+          borderColor: COLORS['blue-standard'],
+          pointBackgroundColor: COLORS['blue-standard'],
+          pointRadius: 4,
+          hitRadius: 5,
+          fill: false,
+          borderWidth: 1,
+          lineTension: 0,
+          data: [0.26, 0.40, 0.90, 0.28, 0.42, 0.41, 0.72, -0.09, -0.10, 0.29,
```
jessex: I know you're working on hooking this up to the API now, but let's start to get out of the habit of checking in hard-coded numbers, even fake ones, even on a branch or PR.

src/components/charts/snapshots/RevocationAdmissionsSnapshot.js
```diff
@@ -0,0 +1,184 @@
...
+
+    setChartLabels(sorted.map((element) => element[0]));
+    setChartDataPoints(sorted.map((element) => element[1]));
+  };
+
+  useEffect(() => {
+    processResponse();
+  }, [props.admissionsByTypeByMonth]);
+
+  const months = [3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 1, 2, 3];
```
jessex: I am guessing these months are also becoming dynamic as part of the work you're doing right now?

src/components/charts/snapshots/RevocationAdmissionsSnapshot.js
```diff
@@ -0,0 +1,184 @@
...
+  const chartData = chart.props.data.datasets[0].data;
+  const mostRecentValue = chartData[chartData.length - 1];
+
+  const chartDataLabels = chart.props.data.labels;
+  const mostRecentMonth = monthNamesFromShortName(chartDataLabels[chartDataLabels.length - 1]);
+
+  const header = document.getElementById('revocationAdmissionsSnapshot-header');
+
+  if (header) {
+    const str1 = "<b style='color:#809AE5'>";
```
jessex: I'm not sure there's a lot of value in breaking this up into several smaller strings that then get concatenated, as opposed to something like 
```
const title = `"<span class='title-highlight'>${mostRecentValue}% of prison admissions</span> in ${mostRecentMonth} were due to parole or probation revocations.`;
```

src/utils/monthConversion.js
```diff
@@ -0,0 +1,38 @@
...
+  return MONTH_NAMES[number - 1];
+};
+
+const monthNameShortFromNumber = function monthNameShortFromNumber(number) {
+  return MONTH_NAMES_SHORT[number - 1];
+};
+
+const monthNamesShortFromNumberList = function monthNamesShortFromNumberList(
+  numberList,
+) {
```
jessex: Is this the linting-approved format when the arg to a function needs to be dropped to a newline? I would expect the closing paren and opening curly brace to still come on the same line as the arg.

src/views/Snapshots.js
```diff
@@ -50,97 +53,81 @@ const Snapshots = () => {
...
+                        <a className="btn btn-secondary btn-sm dropdown-toggle" href="#" role="button" id="exportDropdownMenuButton-successfulSupervisionSnapshot" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
                           Export
                         </a>
-                        <div className="dropdown-menu" aria-labelledby="exportDropdownMenuButton-revocationSnapshot">
-                          <a className="dropdown-item" id="downloadChartAsImage-revocationSnapshot" href="javascript:void(0);">Export image</a>
-                          <a className="dropdown-item" id="downloadChartData-revocationSnapshot" href="javascript:void(0);">Export data</a>
+                        <div className="dropdown-menu" aria-labelledby="exportDropdownMenuButton-successfulSupervisionSnapshot">
+                          <a className="dropdown-item" id="downloadChartAsImage-successfulSupervisionSnapshott" href="javascript:void(0);">Export image</a>
```
jessex: Extra t at the end

src/views/Snapshots.js
```diff
@@ -152,71 +139,51 @@ const Snapshots = () => {
...
+              </div>
+            </div>
+          </div>
+
+          {/* #Change in LSIR scores ==================== */}
+          <div className="masonry-item col-md-6">
+            <div className="bd bgc-white p-20">
+              <div className="layers">
+                <div className="layer w-100 pX-20 pT-20">
+                  <h6 className="lh-1">
```
jessex: Is this right, that we want the title of the chart to be the smaller `h6` and the narrative text to be the larger `h4`?

src/views/Snapshots.js
```diff
@@ -152,71 +139,51 @@ const Snapshots = () => {
...
-                      <ul>
-                        <li>Total admissions include unique people admitted to any DOCR prison during a particular time frame.</li>
-                        <li>Reincarceration returns include only those admissions which resulted from an incarceration, due to a new offense, of a person who was previously incarcerated in a DOCR prison.</li>
-                        <li>In this case, an admission counts as a reincarceration return if the reincarceration happened within 3 years of the person's most recent release from prison, i.e. a 3-year follow-up period.</li>
-                        <li><a href="methodology.html" target="_blank">Read more...</a></li>
-                      </ul>
-                    </div>
-                  </div>
+                <div className="layer w-100 pX-20 pT-20">
+                  <h4 className="lh-1" id="LSIRScoreChangeSnapshot-header">
```
jessex: I am personally not a fan of keeping acronyms upper-case in variable names. I think it dampens readability versus just sticking with CamelCase consistency everywhere. So I'd rename all of these variables to `LsirScoreChangeSnapshot`, etc.

jjdressel11:
> * The export buttons don't work for any of the new snapshots
I will have these working once the data isn't hard coded.

> * It looks like there's too much padding around the "Prison Admissions Due to Revocations" chart. I think it's because there is a div with `class="layer w-100 p-40"` where for the other charts it is `class="layer w-100 p-20"`.

Yeah - I did this to make the boxes the same height when the headers have different amounts of lines in them. Let me know if you know other ways to enforce this.

> * I'm not sure about the narrative text for "Successful completion of supervision." "Successfully completed their supervision by that time" reads a little vague, like it might be focused on _parole having not been extended_, as opposed to people not having been revoked. Were there more conversations about this language?

There wasn't a conversation about any of this language, but there should be. I agree that the language on all of these snapshots could be clearer. Will schedule something with Sonia + Clem to go over this.

> * Without any sense of year in the x-axis for these charts, it's not immediately obvious what time period you're looking at. I don't think the right move is to append the year to each x-axis label, but I wonder if we could do something where the year is applied to, for example, the left-most x-axis label (Mar '18) and to January (Jan '19). How does that sound?

Sounds good.

> * The y-axis description for the "Average Change in LSIR Scores" chart should start with a capital letter.
Agreed

> * I see we no longer have anything below the charts, including drivers, or "Dive into" links, or methodology. That's all expected?

I was going off of Sonia's mocks but will confirm this with Sonia and Clem.

src/assets/scripts/constants/colors.js
jjdressel11: All of the other colors were in lowercase, so I changed all of the ones I had added to be lowercase also. Nothing broke and this wasn't caught by linting.

src/components/charts/snapshots/RevocationAdmissionsSnapshot.js
jjdressel11: Yep - all hard coded numbers in here will be gone.

src/utils/monthConversion.js
jjdressel11: This is the linting-approved format! There's an expected newline before the closing parens.

src/views/Snapshots.js
jjdressel11: Yes this is following the design (you can see all of the mocks [here](https://docs.google.com/presentation/d/1weRaf9TQhB3PN_IG15i3_D59DYsDiTOmPGBmoeFg1l0/edit?pli=1#slide=id.g5f5d7bab86_0_11))

jessex: Right you are!

jessex:
Thanks for the point-by-point responses, @jjdressel11.

>Yeah - I did this to make the boxes the same height when the headers have different amounts of lines in them. Let me know if you know other ways to enforce this.

I see why you did that after tinkering with it locally. On one hand, I'd say that the masonry layout is meant to be flexible to content of different heights and that we shouldn't be so dogmatic about ensuring rigid symmetry. On the other hand, it looks and feels better when they are symmetrical. More importantly from a maintenance perspective, artificially padding internal content to deal with different text length is a non-scalable approach (for example, if a change in the dynamic portion causes the successful completion sentence to shrink slightly, then your padding change is now doing the opposite of what you want). 

One approach would be to try to use `min-height` to ensure that snapshots were never shorter than a specific value and set it to the height of the tallest snapshot card. But that wouldn't work well responsively without further changes, and IE11 has issues with flex containers (which Masonry uses) that have min-height.

I am confident a more gifted frontend engineer knows how to solve this in a universal fashion but in the meanwhile, it's probably not worth our time. Your approach appears to be fairly responsive and good enough for now. But can we file an issue to figure out how to ensure consistent visualization card height when content is of different lengths, and put a TODO on this p-40 referencing the issue?

jjdressel11:
Filed this issue: https://github.com/Recidiviz/pulse-dashboard/issues/50

src/components/charts/snapshots/DaysAtLibertySnapshot.js
```diff
@@ -1,41 +1,37 @@
...
+  }, [props.daysAtLibertyByMonth]);
 
   const chart = (
     <Line
       id="days-at-liberty-snapshot-chart"
       data={{
-        labels: monthNamesShort,
+        labels: chartLabels,
         datasets: [{
+          // TODO: Add custom trendline plugin
```
jessex: What's this about?

src/utils/monthConversion.js
```diff
@@ -26,6 +26,27 @@ const monthNamesShortFromNumberList = function monthNamesShortFromNumberList(
   return monthList;
 };
 
+const monthNamesShortWithYearsFromNumberList = function monthNamesShortWithYearsFromNumberList(
+  numberList,
+) {
+  const monthNames = monthNamesShortFromNumberList(numberList);
+
+  if (numberList.length > 12 || numberList[numberList.length - 1] < numberList[0]) {
```
jessex: So the result of this is that we'll only display years if we cross over a year boundary, regardless of whether we have 12 months of data points or not. I wonder if we should still show the year on the left-most data point even when there's no cross over a year boundary? What do you think?

src/views/Snapshots.js
```diff
@@ -38,7 +38,7 @@ const Snapshots = () => {
     fetchChartData();
   }, []);
 
-  if (loading || !user) {
+  if (loading || !user || awaitingApi) {
```
jessex: I think this is why you're seeing the weird card placement issue, if you're still seeing it. I remember that, for reasons I could not figure out and forgot to document anyway, adding the `awaitingApi` check on this visualization caused card placement issues.

jjdressel11: Yep, you were right!

src/components/charts/snapshots/DaysAtLibertySnapshot.js
jjdressel11: I could not get the trendline plugin working yesterday because it does not redraw the trendline once the data is set. We will eventually need to build a custom trendline plugin so that we can assess what direction trendlines are headed towards. Filed Issue #51.

src/utils/monthConversion.js
jjdressel11: I think you're right. Especially since these users will not be used to seeing such real-time data, giving context to when these metrics are from is important.

jjdressel11:
Ready for another pass!

jjdressel11:
Ready for another pass!

src/utils/monthConversion.js
```diff
@@ -30,28 +30,52 @@ const monthNamesShortWithYearsFromNumberList = function monthNamesShortWithYears
   numberList,
 ) {
   const monthNames = monthNamesShortFromNumberList(numberList);
+  const multipleYears = (numberList.length > 12
```
jessex: `numberList` refers to a list of month numbers, right? If so, let's call it `monthNumbers` or `monthIndices` and skip the Hungarian notation.

src/views/Snapshots.js
```diff
@@ -8,7 +8,7 @@ import { useAuth0 } from '../react-auth0-spa';
 
 import SupervisionSuccessSnapshot from '../components/charts/snapshots/SupervisionSuccessSnapshot';
 import RevocationAdmissionsSnapshot from '../components/charts/snapshots/RevocationAdmissionsSnapshot';
-import LsirScoreChangeSnapshot from '../components/charts/snapshots/LsirScoreChangeSnapshot';
+import LsirScoreChangeSnapshot from '../components/charts/snapshots/LSIRScoreChangeSnapshot';
```
jessex: Why is this back to upper-case?

src/views/Snapshots.js
```diff
@@ -57,19 +57,19 @@ const Snapshots = () => {
...
+                          <a className="dropdown-item" id="downloadChartAsImage-supervisionSuccessSnapshot" href="javascript:void(0);">Export image</a>
+                          <a className="dropdown-item" id="downloadChartData-supervisionSuccessSnapshot" href="javascript:void(0);">Export data</a>
                         </div>
                       </div>
                     </span>
                   </h6>
                 </div>
                 <div className="layer w-100 pX-20 pT-20">
-                  <h4 className="lh-1" id="supervisionSuccessSnapshot-header">
+                  <h4 style={{ height: '60px' }} className="lh-1" id="supervisionSuccessSnapshot-header">
```
jessex: I know I did this previously which was sloppy of me, but we should almost never use manual `style` attributes in our HTML. We should create a new CSS class in `index.scss` named something like `snapshot-header` with `height: 60px` and then add that to the class name for each of these headers.

jessex:
Also, I notice there's still a stacktrace that shows up on backend server start, perhaps because there's a reference to a file that doesn't exist in Cloud Storage. Are you seeing that yourself, @jjdressel11?

src/views/Snapshots.js
jjdressel11: UGH I keep renaming the file in Atom and it doesn't check it in. Going to delete the file and recreate with a new name so it stays lowercase here.

src/views/Snapshots.js
jjdressel11: Okay. The reason for this is the text of these headers are dynamically loaded, so the view doesn't know how big to make it without any text in it. Manually setting it here fixes that and also fixes the headers of each chart not being the same height if they aren't the same number of lines. Will create a custom CSS class for `snapshot-header` !

src/components/charts/snapshots/LSIRScoreChangeSnapshot.js
```diff
@@ -0,0 +1,167 @@
+import React, { useState, useEffect } from 'react';
```
jessex: Looks like this file is still named `LSIRScoreChangeSnapshot.js`. I've had issues with casing and file names in Git before. Do whatever you need to do to get this merged and we can fix it in a follow-up if necessary.

Issue #34, opened 2019-07-25
Make landing page consistent with recidiviz.org


jessex:
We need to generally improve the landing page ("/") for the app, both authenticated and unauthenticated. This should likely copy the styling/layout of recidiviz.org once it has been overhauled, which should happen within the next couple of weeks (as of writing, July 24, 2019).

jessex:
recidiviz.org has indeed been overhauled!

For this particular issue, we want to make the entire dashboard app visually consistent with the redesigned recidiviz.org. The favicon, the landing page, the color palette, the type faces, the logos, etc. The whole thing.

jessex:
@jjdressel11 when I tried out both of the fonts (Libre Baskerville and Poppins), they didn't go really well with any of the current texts in the dashboard, paragraph or title or sub-title. Further, I didn't find a great place to reuse the dark green background anywhere on the dashboard, unless we wanted to really go all in, e.g. dark green for header, sidebar, and footer. I think we should use the dark green text for the top left logo text and I think we should adjust blue links and red goal lines to match the blue and red from the site.

However, that's all to do with the app experience of the actual views like Snapshots and Revocations and so forth. The landing page of the dashboard is a different thing entirely that deserves its own attention.

jjdressel11:
I think it's okay that the fonts are different and that we don't have the dark green background anywhere on the dashboard. The use cases are significantly different (communicating information about who we are as an organization for a first-time visitor vs. communicating criminal justice analytics on an ongoing basis). I think that the landing page should definitely have more consistency with the website, but within the dashboard app we should keep the consistency to smaller elements (logos, using the dark green text color where it fits, using the same red as our logo for goal lines, etc.)

jessex:
I think we should close this now that we've essentially replaced the landing page with the Auth0 login hosted on recidiviz.org. Sound good, @jjdressel11?

jjdressel11:
Yep - closing!

PR #49, opened 2019-08-17
Copies some visual elements from revamped recidiviz.org to the dashboard

jessex:
This goes toward but does not complete #34. Changes:
* Uses the dark green as the color for the logo text in the top left, and the breadcrumb text in the top bar.
* Replaces the logo and the favicons (need to do the same with the Auth0 universal login screen).
* Changes the export buttons to match our call-to-action light green.
* Fixes the collapse and resizing behavior with the logo and text in the top left.


src/components/SideBar.js
```diff
@@ -26,21 +26,19 @@ const SideBar = () => {
...
-                        <img className="logo-icon-holder" src={logo} alt="Logo" />
-                      </div>
-                    </div>
-                    <div className="col-md-9 my-auto peer peer-greed">
-                      <h5 className="lh-1 mB-0 logo-text c-blue-500">North Dakota</h5>
+              <a className="sidebar-link td-n" href="https://www.recidiviz.org/">
+                <div className="peers ai-c fxw-nw pT-15">
+                  <div className="peer">
+                    <div className="col-md-3 my-auto peer">
+                      <img className="logo-icon-holder" src={logo} alt="Logo" />
```
jjdressel11: The `logo-icon-holder` was cutting off the logo a bit on the bottom left corner. If you update the `border-radius` of the `logo-icon-holder` (line 56 of index.scss) to be 3px instead of 6px, it fixes this issue.

src/views/Reincarcerations.js
```diff
@@ -55,7 +55,7 @@ const Reincarcerations = () => {
                     The <span className="font-weight-bold">reincarceration</span> count this month was <span className="font-weight-bold">14</span> over target
                     <span className="fa-pull-right">
                       <div className="dropdown show">
-                        <a className="btn btn-secondary btn-sm dropdown-toggle" href="#" role="button" id="exportDropdownMenuButton-reincarcerationDrivers" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
+                        <a className="btn btn-secondary btn-sm dropdown-toggle recidiviz-light-green-bg" href="#" role="button" id="exportDropdownMenuButton-reincarcerationDrivers" data-toggle="dropdown" aria-haspopup="true" aria-expanded="false">
```
jjdressel11: I understand the motivation behind making these buttons consistent with our call-to-action buttons on the website, but I think that the light green is too distracting in this context. We want the exports to be available as an option, but the "Export" functionality is not where we want people'e eyes to be focused on in the dashboard (compared to the website, where we do want people to click the call-to-action buttons).

src/components/SideBar.js
jessex: Nice! 🦅 👀

src/views/Reincarcerations.js
jessex: Okay fair enough. After doing a side-by-side I think it's fair (but debatable!) to say that the light green is too distracting since the exports are not the real call to action here. Reverting.

jessex:
Ready for a final pass!

PR #53, opened 2019-08-22
Dynamic data and headers

jjdressel11:
In this PR:
- Updates all charts to use the data coming from the API endpoints
- All headers built dynamically (except trendline ones) 
- Removes all hard-coded numbers
- New `snapshots` and `programEval` endpoints
- Fixes a lot of eslint errors in each of these files

Not in this PR:
- Trendlines

Closes: #28


server/core/metricsApi.js
```diff
@@ -98,15 +102,15 @@ function convertDownloadToJson(contents) {
   if (!stringContents || stringContents.length === 0) {
     return null;
   }
-  return JSON.parse(stringContents);
-}
+  // Converts the newline delimited JSON objects into one JSON array
+  let jsonObjectString = '['.concat(stringContents.split('\n').join(','));
```
jessex: An alternate approach that might be a bit more resilient to fluctuation in exact file changes would be to read in the file line by line with a file reader (instead of reading all of the file into memory straight away), passing each line into `JSON.parse` to create an object, and then appending the resulting object to an array of objects.

This would require that the structure of the file is guaranteed to be one json object per line, which also appears to  be an assumption in your current approach. It has the benefits of handling the json at the atomic unit that BQ is outputting (at the object level, instead of a manually composed array level) and of not having to do substring and concat operations that would break down if the file format changes without warning.

src/components/charts/reincarcerations/AdmissionsVsReleases.js
```diff
@@ -0,0 +1,97 @@
...
+  const [chartDataPoints, setChartDataPoints] = useState([]);
+
+  const processResponse = () => {
+    const { admissionsVsReleases } = props;
+
+    let sorted = [];
+    admissionsVsReleases.forEach((data) => {
+      const { year } = data;
+      const { month } = data;
+      const { population_change: delta } = data;
```
jessex: I may be mistaken, but I think you can collapse these lines:
```
const { year, month, population_change: delta } = data;
```

src/components/charts/reincarcerations/ReincarcerationCountOverTime.js
```diff
@@ -1,129 +1,151 @@
...
+      const { month } = data;
+      const { returns: count } = data;
+      sorted.push([year, month, count]);
+    });
 
-    setChartLabels(sorted.map(element => element[0]));
-    setChartDataPoints(sorted.map(element => element[1]));
-  }
+    // Sort by month and year
+    sorted.sort((a, b) => ((a[0] === b[0]) ? (a[1] - b[1]) : (a[0] - b[0])));
```
jessex: We should move these sort and slice functions out into utils that all chart visualizations can reuse, e.g. `sortByYearAndMonth(dataPoints)`, `filterMostRecentMonths(dataPoints, monthCount)`

src/components/charts/reincarcerations/ReincarcerationRateByReleaseFacility.js
```diff
@@ -1,87 +1,93 @@
-import React, { useState, useEffect } from "react";
+import React, { useState, useEffect } from 'react';
 
 import { Bar } from 'react-chartjs-2';
-import { COLORS } from "../../../assets/scripts/constants/colors";
+import { COLORS } from '../../../assets/scripts/constants/colors';
+
+const TRANSITIONAL_FACILITIES = ['FTPFAR', 'GFC', 'BTC', 'FTPMND', 'MTPFAR', 'LRRP', 'MTPMND'];
```
jessex: Let's move this into a facility filter utility. It would take in the array of data points and the state code, then look for filter arrays mapped to that particular state code, and process the filter. Then it will be ready for future states to reuse, and we won't have to hard-code state specific information directly into a chart component. And moving this array out behind a function will make it easier to extract it out to an API in the future.

src/components/charts/revocations/RevocationCountByViolationType.js
```diff
@@ -1,79 +1,123 @@
...
-    for (var month in countsByMonth) {
-        sorted.push([month, countsByMonth[month]]);
+    let sorted = [];
+    countsByMonth.forEach((data) => {
+      const { year } = data;
+      const { month } = data;
+      const { absconsion_count: absconsionCount } = data;
+      const { felony_count: felonyCount } = data;
+      const { technical_count: technicalCount } = data;
+      const { unknown_count: unknownCount } = data;
```
jessex: I won't comment on it to death, but I think we can collapse these destructures into a single call everywhere.

src/components/charts/revocations/RevocationProportionByRace.js
```diff
@@ -1,102 +1,128 @@
...
+    const total = countsByRaceData.map((element) => element[1]).reduce(
+      (previousValue, currentValue) => (previousValue + currentValue),
+    );
 
-    setChartLabels(sorted.map(element => element[0]));
-    setChartProportions(sorted.map(element => element[1]));
-    setStateProportions(sorted.map(element => ND_RACE_PROPORTIONS[element[0]]));
-  }
+    // Sort by race alphabetically
+    countsByRaceData.sort((a, b) => a[0].localeCompare(b[0]));
```
jessex: This alphabetical sort by label has been used in multiple places, too, so this can also be extracted out to a utility.

src/components/charts/revocations/RevocationProportionByRace.js
```diff
@@ -1,102 +1,128 @@
...
+      data={{
+        labels: ['Revocations', 'ND Population'],
+        datasets: [{
           label: chartLabels[0],
           backgroundColor: COLORS_FIVE_VALUES[0],
           data: [chartProportions[0], statePopulationProportions[0]],
         }, {
           label: chartLabels[1],
-          backgroundColor: COLORS_FIVE_VALUES[1],
+          backgroundColor: COLORS_FIVE_VALUES[3],
```
jessex: It's non-obvious to future readers why these color values are "out of order." Why is that?

src/components/charts/snapshots/DaysAtLibertySnapshot.js
```diff
@@ -3,22 +3,34 @@ import React, { useState, useEffect } from 'react';
...
-    for (var month in countsByMonth) {
-      sorted.push([month, countsByMonth[month]]);
-    }
+    if (daysAtLibertyByMonth) {
+      let sorted = [];
+
+      daysAtLibertyByMonth.forEach((data) => {
+        const { year } = data;
+        const { month } = data;
+        const count = parseFloat(data.avg_liberty).toFixed(2);
```
jessex: Nit: it's not a count, but an average.

server/core/metricsApi.js
jjdressel11: Changed to this: still reading the whole file into memory, but then splitting by new lines and parsing each line as an object and adding to an array of objects. This still assumes one JSON object per line, but avoids the concat operations.

src/components/charts/snapshots/SupervisionSuccessSnapshot.js
```diff
@@ -3,22 +3,43 @@ import React, { useState, useEffect } from 'react';
...
+      let sorted = [];
+      countsByMonth.forEach((data) => {
+        const { projected_year: year } = data;
+        const { projected_month: month } = data;
+        const successful = parseInt(data.successful_termination, 10);
+        const revocation = parseInt(data.revocation_termination, 10);
+        const successRate = (100 * (successful / (successful + revocation))).toFixed(2);
+
+        // Don't add completion rates for months in the future
+        if (year < yearNow || month <= monthNow) {
```
jessex: Should this be an `&&`? If it's currently August 2019 and the data point is February 2024, it will still get added as currently implemented, right?

src/components/charts/reincarcerations/AdmissionsVsReleases.js
jjdressel11: You can! Will change this everywhere it applies.

src/views/Revocations.js
```diff
@@ -166,18 +166,11 @@ const Revocations = () => {
...
-                      <span className="fsz-def fw-600 mR-10 c-grey-800">Last 60 days</span>
-                    </div>
-                  </div>
-                </div>
               </div>
             </div>
           </div>
 
-          {/* #Revocations by officer ==================== */}
+          {/* TODO: Bring this chart back once the calculation is implemented in PROD: #Revocations by officer ====================
```
jessex: Let's file an issue and reference it here.

src/components/charts/revocations/RevocationProportionByRace.js
jjdressel11: Actually now that I'm sorting the races alphabetically these should be in order. Changing.

src/components/charts/snapshots/SupervisionSuccessSnapshot.js
jjdressel11: Changed to `if (year < yearNow || (year === yearNow && month <= monthNow))`

src/utils/dataOrganizing.js
```diff
@@ -0,0 +1,55 @@
...
+/**
+ * Sorts the data points by year and month, ascending.
+ * Assumes that the the data is in the format:
+ * [[year, month, data], [year, month, data], ...]
+ */
+function sortByYearAndMonth(dataPoints) {
+  return dataPoints.sort((a, b) => ((a[0] === b[0]) ? (a[1] - b[1]) : (a[0] - b[0])));
+}
+
+function sortByLabel(dataPoints, labelIndex) {
```
jessex: Let's add a little docstring for this one explaining what the format is expected to be, like the function above.

jessex:
Notes from local manual testing!
* We should apply the conditional year label logic to charts in the other views, not just snapshots (e.g. Mar '19)
* We should remove the "Read more..." links from the methodology accordions--we should also double check that we are happy with all of at methodology text
* We should file an issue to add export buttons to all charts

jjdressel11:
I removed the "Read more..." from the methodology accordions and updated a lot of the Methodology strings. Would love for you to look them over!

src/views/Revocations.js
```diff
@@ -275,10 +271,8 @@ const Revocations = () => {
                   <div className="collapse" id="collapseMethodologyRevocationsByRace" aria-labelledby="methodologyHeadingRevocationsByRace" data-parent="#methodologyRevocationsByRace">
                     <div>
                       <ul>
-                        <li>Violations include all behavioral violations officially recorded by a supervision officer, including new offenses, technical violations, and absconsion.</li>
-                        <li>Revocations include all instances of a person having any form of supervision revoked for any behavioral violation, and being reincarcerated as a result.</li>
-                        <li>Technical revocations include only those revocations which result only from a technical violation and/or absconsion. If there is a technical violation in addition to a new offense, it is considered a non-technical revocation.</li>
-                        <li><a href="methodology.html" target="_blank">Read more...</a></li>
+                      <li>Revocations include all instances of a person being incarcerated because their supervision was revoked for a behavioral violation.</li>
```
jessex: Nit: reindent these back to their previous level.

src/views/Revocations.js
```diff
@@ -158,10 +154,10 @@ const Revocations = () => {
                   <div className="collapse" id="collapseMethodologyRevocationsByViolationType" aria-labelledby="methodologyHeadingRevocationsByViolationType" data-parent="#methodologyRevocationsByViolationType">
                     <div>
                       <ul>
-                        <li>Violations include all behavioral violations officially recorded by a supervision officer, including new offenses, technical violations, and absconsion.</li>
-                        <li>Revocations include all instances of a person having any form of supervision revoked for any behavioral violation, and being reincarcerated as a result.</li>
-                        <li>Technical revocations include only those revocations which result only from a technical violation and/or absconsion. If there is a technical violation in addition to a new offense, it is considered a non-technical revocation.</li>
-                        <li><a href="methodology.html" target="_blank">Read more...</a></li>
+                      <li>Revocations include all instances of a person being incarcerated because their supervision was revoked for a behavioral violation.</li>
```
jessex: Nit: reindent these back to their previous level.

src/views/Revocations.js
```diff
@@ -125,10 +123,8 @@ const Revocations = () => {
                   <div className="collapse" id="collapseMethodologyRevocationBySupervisionType" aria-labelledby="methodologyHeadingRevocationBySupervisiontype" data-parent="#methodologyRevocationBySupervisionType">
                     <div>
                       <ul>
-                        <li>Violations include all behavioral violations officially recorded by a supervision officer, including new offenses, technical violations, and absconsion.</li>
-                        <li>Revocations include all instances of a person having any form of supervision revoked for any behavioral violation, and being reincarcerated as a result.</li>
-                        <li>Technical revocations include only those revocations which result only from a technical violation and/or absconsion. If there is a technical violation in addition to a new offense, it is considered a non-technical revocation.</li>
-                        <li><a href="methodology.html" target="_blank">Read more...</a></li>
+                      <li>Revocations include all instances of a person being incarcerated because their supervision was revoked for a behavioral violation.</li>
```
jessex: Nit: reindent these back to their previous level.

PR #54, opened 2019-08-22
Removing program evaluation tab

jjdressel11:
In the next release of the dashboard, we are going to have all **real** and **live** data, so we want to remove any fake or non-updated data. Temporarily disabling the program evaluation tab until we have real program eval metrics to report.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -12,7 +12,6 @@ import Footer from "./components/Footer";
 import Home from "./views/Home";
 import NotFound from "./views/NotFound";
 import Profile from "./views/Profile";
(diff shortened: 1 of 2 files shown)


jessex:
Quick question since I'm too lazy to check myself: because you removed the `PrivateRoute` on the React side, if you try to navigate directly to the URL, it will 404, right?

jjdressel11:
> Quick question since I'm too lazy to check myself: because you removed the `PrivateRoute` on the React side, if you try to navigate directly to the URL, it will 404, right?

Yes, it will!

Issue #52, opened 2019-08-22
Make text about trend directions dynamic


jjdressel11:
These are currently hard-coded. These should be dynamic strings based on the direction of the trend and the location of the goal.

PR #58, opened 2019-08-23
Adding trendlines to snapshot charts

### LINKED ISSUE #51: Build trendline Chart.js plugin
See [documentation](https://www.chartjs.org/docs/latest/developers/plugins.html) on building custom plugins

Requirements of the plugin:
- Recalculates whenever datasets are changed (`afterDatasetUpdate`)
- Can access slope values to asses in what direction trends are headed

jjdressel11:
In this PR:
- Utils for calculating the values of a linear trendline for a set of data points
- Dynamically building header text about trends based on the type of goal (either a goal to increase a metric or to decrease a metric) and the slope of the trendline

Closes #51 and #52


src/components/charts/snapshots/LsirScoreChangeSnapshot.js
```diff
@@ -40,16 +47,26 @@ const LsirScoreChangeSnapshot = (props) => {
...
           borderColor: COLORS['blue-standard'],
           pointBackgroundColor: COLORS['blue-standard'],
           pointRadius: 4,
           hitRadius: 5,
           fill: false,
-          borderWidth: 1,
+          borderWidth: 2,
           lineTension: 0,
           data: chartDataPoints,
+        }, {
```
jessex: If the trendline definition is the same for all charts, we should extract its definition out into a utility, e.g. `generateTrendline(chartDataPoints)`.

src/components/charts/snapshots/DaysAtLibertySnapshot.js
```diff
@@ -5,11 +5,18 @@ import { configureDownloadButtons } from '../../../assets/scripts/charts/chartJS
...
 import { monthNamesWithYearsFromNumbers } from '../../../utils/monthConversion';
 import { sortAndFilterMostRecentMonths } from '../../../utils/dataOrganizing';
+import { slopeOfTrendline, trendlineData } from '../../../utils/trendline';
 
 const DaysAtLibertySnapshot = (props) => {
   const [chartLabels, setChartLabels] = useState([]);
   const [chartDataPoints, setChartDataPoints] = useState([]);
 
+  const GOAL = {
+    aimingHigher: true,
```
jessex: A more general key name here might be `direction` with possible values of `increase` or `decrease`. I leave it to your better judgement.

src/components/charts/snapshots/DaysAtLibertySnapshot.js
```diff
@@ -5,11 +5,18 @@ import { configureDownloadButtons } from '../../../assets/scripts/charts/chartJS
 import { COLORS } from '../../../assets/scripts/constants/colors';
 import { monthNamesWithYearsFromNumbers } from '../../../utils/monthConversion';
 import { sortAndFilterMostRecentMonths } from '../../../utils/dataOrganizing';
+import { slopeOfTrendline, trendlineData } from '../../../utils/trendline';
 
 const DaysAtLibertySnapshot = (props) => {
   const [chartLabels, setChartLabels] = useState([]);
   const [chartDataPoints, setChartDataPoints] = useState([]);
 
+  const GOAL = {
```
jessex: We could extract this out into a dedicated class for the type safety and ability to create a stable contract for goal-oriented charts.

src/components/charts/snapshots/DaysAtLibertySnapshot.js
```diff
@@ -68,7 +80,13 @@ const DaysAtLibertySnapshot = (props) => {
         },
         tooltips: {
           enabled: true,
-          mode: 'x',
+          mode: 'point',
+          callbacks: {
+            label: (tooltipItem) => {
+              if (tooltipItem.datasetIndex > 0) return '';
```
jessex: It this what makes it so there's no tooltip addition for the trendline? If so, it would be more stable to look at somethign about the dataset itself instead of just its placement within the array of datasets, i.e. look for `label='trendline'`.

jessex: Also, since it looks like this exists for all charts with trendlines (with a small variation with the % sign), it would make sense to extract this function out to a utility, as well, e.g. `getTooltipWithoutTrendline(tooltipItem)`.

src/components/charts/snapshots/RevocationAdmissionsSnapshot.js
```diff
@@ -5,11 +5,18 @@ import { configureDownloadButtons } from '../../../assets/scripts/charts/chartJS
 import { COLORS } from '../../../assets/scripts/constants/colors';
 import { monthNamesWithYearsFromNumbers, monthNamesFromShortName } from '../../../utils/monthConversion';
 import { sortAndFilterMostRecentMonths } from '../../../utils/dataOrganizing';
+import { trendlineData } from '../../../utils/trendline';
 
 const RevocationAdmissionsSnapshot = (props) => {
   const [chartLabels, setChartLabels] = useState([]);
   const [chartDataPoints, setChartDataPoints] = useState([]);
 
+  const GOAL = {
```
jessex: I think we should go ahead and create a function for getting goals for specific charts, for specific states, e.g. `getGoalForChart(state_code, chart_name)` or something like that. That will make it easier to shift from hard-coded goals to an API in the near future.

src/components/charts/snapshots/LsirScoreChangeSnapshot.js
```diff
@@ -163,10 +186,18 @@ const LsirScoreChangeSnapshot = (props) => {
     document.getElementById('lsir-score-change-snapshot-chart'), exportedStructureCallback);
 
   const header = document.getElementById(props.header);
+  const trendlineValues = chart.props.data.datasets[1].data;
+  const trendlineSlope = slopeOfTrendline(trendlineValues);
+  let trendText = '';
+
+  if (GOAL.aimingHigher) {
```
jessex: I see this logic is only in place for 2 charts instead of 4. How does the dynamic header work for the other 2?

src/components/charts/snapshots/DaysAtLibertySnapshot.js
```diff
@@ -167,10 +185,18 @@ const DaysAtLibertySnapshot = (props) => {
     document.getElementById('days-at-liberty-snapshot-chart'), exportedStructureCallback);
 
   const header = document.getElementById(props.header);
+  const trendlineValues = chart.props.data.datasets[1].data;
+  const trendlineSlope = slopeOfTrendline(trendlineValues);
+  let trendText = '';
+
+  if (GOAL.aimingHigher) {
```
jessex: If there's consistency in the text we choose based on goal direction across charts, we should extract this out into a utility as well.

src/components/charts/snapshots/LsirScoreChangeSnapshot.js
```diff
@@ -163,10 +186,18 @@ const LsirScoreChangeSnapshot = (props) => {
     document.getElementById('lsir-score-change-snapshot-chart'), exportedStructureCallback);
 
   const header = document.getElementById(props.header);
+  const trendlineValues = chart.props.data.datasets[1].data;
+  const trendlineSlope = slopeOfTrendline(trendlineValues);
```
jessex: Nit: I'd call this `trendlineSlope` so it's consistent with `trendlineData`.

PR #67, opened 2019-09-17
Adds mistakenly ignored package.json file.

jessex:
This file should not be git-ignored, just `package-lock.json`. If you have any local changes you've made, we should issue a PR to add those now.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,5 +38,4 @@ yarn-debug.log*
 yarn-error.log*
 
 # Package files
-package.json
 package-lock.json


jjdressel11:
What do you mean by "local changes" that I should get in now?

jessex:
If you made changes to the package.json file on your local machine after we merged the `.gitignore` change here, and they are required, then we should issue a PR to make those changes on master.

jjdressel11:
Nope - haven't made any changes that need to be on master yet.

Issue #55, opened 2019-08-22
Re-implement RevocationCountByOfficer chart


jjdressel11:
Once we start ingesting the `TERMINATING_OFFICER` field on the `OffenderCasesTable`, we will be able to provide revocation counts by officer for ND. Once we have these calculations, we need to re-implement the `RevocationCountByOfficer` chart on the revocations page.

PR #68, opened 2019-09-18
Revocations by county and officer ID

jjdressel11:
In this PR:

- Revocation counts by county for the last 60 days
- Revocation counts by officer, broken down by violation type, for the last 60 days
- Updates colors of `AdmissionTypeProportions` chart to match the colors of other charts
- Package updates to support these

Closes: https://github.com/Recidiviz/pulse-dashboard/issues/55


src/components/charts/revocations/RevocationCountByOfficer.js
```diff
@@ -1,80 +1,126 @@
...
+        TECHNICAL: parseInt(technicalCount, 10),
+        UNKNOWN_VIOLATION_TYPE: parseInt(unknownCount, 10),
+      };
+
+      let overallRevocationCount = 0;
+      Object.keys(officerDict).forEach((violationType) => {
+        overallRevocationCount += officerDict[violationType];
+      });
+
+      if (officerID !== 'OFFICER_UNKNOWN') dataPoints.push([officerID, officerDict, overallRevocationCount]);
```
jessex: Let's break this out into a full if statement instead of a one-liner. I find one-liners more difficult to read at a glance and more error-prone.

jessex: Instead of having `dataPoints` be an array of arrays that we then have to index into magically, e.g. officerId = index 0, we should have it be an array of objects with proper keys, e.g. officerId = object.officerId.

src/components/charts/revocations/RevocationCountByOfficer.js
```diff
@@ -1,80 +1,126 @@
...
+      ABSCONDED: [],
+      FELONY: [],
+      TECHNICAL: [],
+      UNKNOWN_VIOLATION_TYPE: [],
+    };
+
+    const sortedDataPoints = dataPoints.sort((a, b) => (b[2] - a[2]));
+
+    for (let i = 0; i < 10; i += 1) {
+      officerLabels.push(sortedDataPoints[i][0]);
```
jessex: See my comment above: if we switch to objects instead of arrays for each data point, then we can get rid of all of this magical indexing.

src/components/charts/revocations/RevocationsByCounty.js
```diff
@@ -0,0 +1,127 @@
...
+      const {
+        state_code: stateCode,
+        county_code: countyCode,
+        revocation_count: revocationCount,
+      } = data;
+
+      const revocationCountNum = parseInt(revocationCount, 10);
+
+      if (countyCode !== 'UNKNOWN_COUNTY') {
+        if (revocationCountNum > max) max = revocationCountNum;
```
jessex: Same line here, let's avoid one-liner if statements.

src/components/charts/revocations/RevocationsByCounty.js
```diff
@@ -0,0 +1,127 @@
...
+    Markers,
+    Marker,
+  } from 'react-simple-maps';
+import ReactTooltip from 'react-tooltip';
+import { geoAlbersUsa } from 'd3-geo';
+import { scaleLinear } from 'd3-scale';
+import geographyObject from '../../../assets/static/maps/us_nd.json';
+import { COLORS } from '../../../assets/scripts/constants/colors';
+
+let max = -1e100;
```
jessex: So this being a static value could produce bugs in edge cases, e.g. if the values in the metrics bucket change while the app is still up with a single instance and this value becomes invalidated but has not been appropriately updated yet. I think we should turn this into state of the component. The value should be set again whenever the component is reinitialized.

src/components/charts/revocations/RevocationsByCounty.js
```diff
@@ -0,0 +1,127 @@
...
+  if (revocationsForCounty) {
+    return revocationsForCounty;
+  }
+
+  return 0;
+}
+
+function colorForCounty(chartDataPoints, countsByCounty, countyName) {
+  const countyScale = scaleLinear()
+    .domain([0, max / 2, max])
```
jessex: When we convert `max` into state, we'll pass it into this function.

src/components/charts/revocations/RevocationsByCounty.js
```diff
@@ -0,0 +1,127 @@
...
+    .range(['#F5F6F7', '#9FB1E3', COLORS['blue-standard-2']]);
+
+  const revocationsForCounty = revocationCountForCounty(chartDataPoints, countyName);
+  return countyScale(revocationsForCounty);
+}
+
+class RevocationsByCounty extends Component {
```
jessex: Why did we choose to make this a stateful component instead of stateless like the others?

src/components/charts/revocations/RevocationsByCounty.js
```diff
@@ -0,0 +1,127 @@
...
+          projection={geoAlbersUsa}
+          projectionConfig={{ scale: 1000 }}
+          width={980}
+          height={500}
+          style={{
+            width: '100%',
+            height: 'auto',
+          }}
+        >
+          <ZoomableGroup center={[-100.5, 47.3]} zoom={7} disablePanning>
```
jessex: Can we extract these numbers out to well-named constants?

src/views/Revocations.js
```diff
@@ -166,15 +167,58 @@ const Revocations = () => {
...
+                      <button className="btn btn-link collapsed pL-0" type="button" data-toggle="collapse" data-target="#collapseMethodologyRevocationsByCounty" aria-expanded="true" aria-controls="collapseMethodologyRevocationsByCounty">
+                        <h6 className="lh-1 c-blue-500 mb-0">Methodology</h6>
+                      </button>
+                    </div>
+                  </div>
+                  <div className="collapse" id="collapseMethodologyRevocationsByCounty" aria-labelledby="methodologyHeadingRevocationsByCounty" data-parent="#methodologyRevocationsByCounty">
+                    <div>
+                      <ul>
+                        <li>Revocations include all instances of a person being incarcerated because their supervision was revoked for a behavioral violation.</li>
+                        <li>The county is the recorded county where the supervision was terminated.</li>
```
jessex: I might reword this slightly to: "Revocations are attributed to the county where the person's supervision was terminated."

Can you remind me exactly which county code field we attributing this to, so I understand?

src/views/Revocations.js
```diff
@@ -187,25 +231,24 @@ const Revocations = () => {
                   <div id="collapseMethodologyRevocationByOfficer" className="collapse" aria-labelledby="methodologyHeadingRevocationByOfficer" data-parent="#methodologyRevocationByOfficer">
                     <div>
                       <ul>
+                        <li>This chart displays the top 10 revocation counts of officers in the state.</li>
```
jessex: Might reword this to: "This chart lists the 10 officers with the highest revocation counts in the state over the period."

jessex:
Okay! I tested locally and have a few suggestions:
* On the county chart, the counties with the fewest revocations are perhaps a bit too difficult to distinguish from those without any revocations. Have we done an accessibility check on our color scheme for that? 
* As a product point, it occurs to me in looking at the county chart that if we just show absolute revocation volumes, it may be that the same few counties are always the darkest shade. If we had a toggle that let us switch to graphing the rate of revocation success for that county, it might give us more interesting insights. That could be a second pass for this work.
* We should configure the Y-axis for the officer chart to only show whole number increments, without decimal points.
* I notice a very slight difference in the height of the two new graphs, which leads to odd behavior when the sidebar is collapsed (the order of the next two charts after that row gets swapped, and the first instance of sidebar collapse sends us back to the top of the page, unfortunately). I'm not sure we have a solution for that that's not hacky or brittle. It may be in our long-term interest to abandon the masonry layout but certainly not in our short-term interest.

src/components/charts/revocations/RevocationCountByOfficer.js
jjdressel11: Both great ideas - thanks!

src/components/charts/revocations/RevocationsByCounty.js
jjdressel11: Done!

src/components/charts/revocations/RevocationsByCounty.js
jjdressel11: I don't really know how react components work, and this was the way I found to make the `'react-simple-maps'` work properly (they weren't working if I tried returning them like we do the chart.js charts).

src/views/Revocations.js
jjdressel11: We're taking the county code from the `StateSupervisionPeriod` that's connected to the `SourceSupervisionViolationResponse` through the violation.

jjdressel11:
> * On the county chart, the counties with the fewest revocations are perhaps a bit too difficult to distinguish from those without any revocations. Have we done an accessibility check on our color scheme for that?

Yeah, I noticed that as well. I upped the contrast between a state with 0 revocations and a state with 1. You can see the difference between that those two colors for all forms of color blindness [here](https://davidmathlogic.com/colorblind/#%23D3D9ED-%23F2F4F5) 

> * As a product point, it occurs to me in looking at the county chart that if we just show absolute revocation volumes, it may be that the same few counties are always the darkest shade. If we had a toggle that let us switch to graphing the rate of revocation success for that county, it might give us more interesting insights. That could be a second pass for this work.

This is actually what we're planning to do with MO. Filed #71 

> * We should configure the Y-axis for the officer chart to only show whole number increments, without decimal points.

Done. 

> * I notice a very slight difference in the height of the two new graphs, which leads to odd behavior when the sidebar is collapsed (the order of the next two charts after that row gets swapped, and the first instance of sidebar collapse sends us back to the top of the page, unfortunately). I'm not sure we have a solution for that that's not hacky or brittle. It may be in our long-term interest to abandon the masonry layout but certainly not in our short-term interest.

Yeah... the problem with masonry is that it's built for things like pinterest where you don't have rigid grid-like layouts where the rows are the same height. We could probably work around this by making the height static for individual divs within each chart div? But seems not super high priority right now.

jessex:
Thanks for the replies to all of my points!

>We could probably work around this by making the height static for individual divs within each chart div? But seems not super high priority right now.

Agreed, not high priority for now.

PR #69, opened 2019-09-18
Adds a rudimentary 'demo mode' to the app

jessex:
For debugging and demo purposes. Pulls data from static files structured in the same way as live metric files. Data has been more or less randomized with some trends baked in for visual interest and testability. We'll need to keep this directory up to date as we add new files but we'd need to do this eventually anyway when we introduce unit testing to the app.


server/core/metricsApi.js
```diff
@@ -114,20 +147,20 @@ function convertDownloadToJson(contents) {
   return jsonObject;
 }
 
-function fetchSnapshotMetrics(callback) {
-  return fetchMetrics('US_ND', 'snapshot', callback);
+function fetchSnapshotMetrics(isDemo, callback) {
+  return fetchMetrics('US_ND', 'snapshot', IS_DEMO, callback);
```
jjdressel11: Why are you having all of these functions get `isDemo` if you are using the environment `IS_DEMO` variable in the return calls?

server/core/metricsApi.js
```diff
@@ -74,14 +94,27 @@ function fetchMetricsFromGCS(stateCode, metricType) {
  * First checks the cache to see if the metrics with the given type are already in memory and not
  * expired beyond the configured TTL. If not, then fetches the metrics for that type from the
  * appropriate files and invokes the callback only once all files have been retrieved.
+ *
+ * If we are in demo mode, then fetches the files from a static directory, /server/core/demoData/.
+ * Otherwise, fetches from Google Cloud Storage.
  */
-function fetchMetrics(stateCode, metricType, callback) {
+function fetchMetrics(stateCode, metricType, isDemo, callback) {
   return memoryCache.wrap(metricType, function (cacheCb) {
```
jjdressel11: Change to: 
```
  return memoryCache.wrap(metricType, (cacheCb) => {
    let fetcher = null;
    let source = null;
    if (isDemo) {
      source = 'local';
      fetcher = fetchMetricsFromLocal;
    } else {
      source = 'GCS';
      fetcher = fetchMetricsFromGCS;
    }
```

To fix a bunch of eslint errors.

src/components/charts/reincarcerations/ReincarcerationRateByReleaseFacility.js
```diff
@@ -13,11 +13,18 @@ const ReincarcerationRateByReleaseFacility = (props) => {
...
+        recidivism_rate: recidivismRate,
+        state_code: stateCode,
+      } = data;
+      dataPoints.push([facility, recidivismRate, stateCode]);
     });
 
-    dataPoints = filterFacilities(dataPoints, 'RELEASE', 'US_ND');
+    if (dataPoints.length > 0) {
+      const stateCode = dataPoints[0][2];
+      dataPoints = filterFacilities(dataPoints, 'RELEASE', stateCode);
```
jjdressel11: This is great - thank you!

server/core/metricsApi.js
jjdressel11: There are a bunch of existing eslint errors here that you could clean up if you wanted to 😎

jessex: I'm planning to submit a follow-up PR with all of the lint fixes!

server/core/metricsApi.js
jessex: Ah that was a mistake. I changed approaches at one point and forgot to clean this up.

jessex:
Ready for a final pass, @jjdressel11.

Issue #26, opened 2019-07-24
Reenable linting


jessex:
We have a medley of linting errors, via `eslint`, after the quick port to React. We need to clean those up and reenable automatic linting in local and Travis builds.

PR #74, opened 2019-09-21
Applies many lint fixes across the server side of the app.

jessex:
This goes towards #26 but does not complete it.


server/core/metricsRefresh.js
```diff
@@ -17,12 +17,12 @@ const METRIC_REFRESH_INTERVAL_MS = 1000 * 60 * 30; // Refresh metrics every 30 m
  * Performs a refresh of the program evaluation metrics cache, logging success or failure.
  */
 function refreshProgramEvalMetrics() {
-  metricsApi.fetchProgramEvalMetrics(IS_DEMO, function (err, data) {
```
jjdressel11: The `data` variable is never used in this file. Do you need it in these functions?

jessex: It's part of the interface for promises, so we do need it in the function signatures, I believe.

PR #76, opened 2019-09-25
Fixes broken export buttons for top-line revocations and reincarcerations charts

jessex:
The export functionality is a bit brittle in how it expects things to be named. We'll fix that later.


jessex:
Merging now to proceed with deploy.

PR #78, opened 2019-09-26
Updates methodological text in several places to increase clarity.

jessex:
These are the updates I made based on talk with Tom@ND. Feel free to wordsmith!

* Increase precision of supervision success versus non-success in the snapshot methodology
* Updated the narrative headlines for the top-line revocation and reincarceration counts to note that it's specifically when a person goes back to a DOCR facility (as opposed to, say, being revoked and winding up in a county jail)
* Made explicit in the reincarceration count methodology that there's no follow up period upper bound as currently implemented
* Updated the line that appears in several revocation charts to highlight that it's person-based, not case-based (I know this is only an implicit constraint right now, but it's effectively true except in really rare cases, to be addressed in #2191 on the platform side)
* Tweaked some text about revocation dates to be more clear

### DIFF
--- a/src/views/Reincarcerations.js
+++ b/src/views/Reincarcerations.js
@@ -97,7 +97,7 @@ const Reincarcerations = () => {
                         </li>
                         <li>
                         Reincarcerations are included regardless of when the initial incarceration
-                        took place. There is no upper bound on the follow up period in this chart.
+                        took place. There is no upper bound on the follow up period in this metric.
                         </li>
                       </ul>
                     </div>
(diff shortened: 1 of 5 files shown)


src/views/Reincarcerations.js
```diff
@@ -95,6 +95,10 @@ const Reincarcerations = () => {
                         the person has been incarcerated previously in a North
                         Dakota prison.
                         </li>
+                        <li>
+                        Reincarcerations are included regardless of when the initial incarceration
+                        took place. There is no upper bound on the follow up period in this chart.
```
jjdressel11: Let's use the word "metric" instead of "chart"?

Issue #62, opened 2019-08-28
Compare revocations by race to supervision pop by race


jjdressel11:
Show breakdowns of the ND supervision population by race and compare to the revocations broken down by race instead of just the overall ND population by race.

PR #79, opened 2019-09-27
Adding supervision population by race to revocation race chart

jjdressel11:
Closes https://github.com/Recidiviz/pulse-dashboard/issues/62

Adds the breakdown of the current supervision population by race/ethnicity to the revocations by race chart.

Includes updates to the relevant demo data files.

### DIFF
--- a/server/core/demoData/revocations_by_race_and_ethnicity_60_days.json
+++ b/server/core/demoData/revocations_by_race_and_ethnicity_60_days.json
@@ -1,0 +1,5 @@
+{"state_code":"US_DEMO","race_or_ethnicity":"OTHER","revocation_count":"7"}
+{"state_code":"US_DEMO","race_or_ethnicity":"AMERICAN_INDIAN_ALASKAN_NATIVE","revocation_count":"35"}
+{"state_code":"US_DEMO","race_or_ethnicity":"BLACK","revocation_count":"10"}
+{"state_code":"US_DEMO","race_or_ethnicity":"NATIVE_HAWAIIAN_PACIFIC_ISLANDER","revocation_count":"4"}
+{"state_code":"US_DEMO","race_or_ethnicity":"WHITE","revocation_count":"53"}
--- a/src/components/charts/revocations/RevocationProportionByRace.js
+++ b/src/components/charts/revocations/RevocationProportionByRace.js
@@ -27,74 +27,114 @@ const ND_RACE_PROPORTIONS = {
 const RevocationProportionByRace = (props) => {
   const [chartLabels, setChartLabels] = useState([]);
   const [chartProportions, setChartProportions] = useState([]);
-  const [statePopulationProportions, setStateProportions] = useState([]);
+  const [statePopulationProportions, setStatePopulationProportions] = useState([]);
+  const [stateSupervisionProportions, setStateSupervisionProportions] = useState([]);
 
   const processResponse = () => {
-    const { revocationProportionByRace: proportionsByRace } = props;
+    const { revocationProportionByRace } = props;
+    const { supervisionPopulationByRace } = props;
 
-    const dataPoints = [];
-    proportionsByRace.forEach((data) => {
+    const revocationDataPoints = [];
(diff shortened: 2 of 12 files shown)


server/core/demoData/revocations_by_race_and_ethnicity_60_days.json
```diff
@@ -0,0 +1,5 @@
+{"state_code":"US_DEMO","race_or_ethnicity":"OTHER","revocation_count":"7"}
```
jessex: To be sure, these are fake values and not real numbers for any particular state, right?

src/components/charts/revocations/RevocationProportionByRace.js
```diff
@@ -27,74 +27,114 @@ const ND_RACE_PROPORTIONS = {
...
-      dataPoints.push([labelStringConversion[race], count]);
+      revocationDataPoints.push([labelStringConversion[race], count]);
     });
 
-    const racesRepresented = dataPoints.map((element) => element[0]);
+    const supervisionDataPoints = [];
+    supervisionPopulationByRace.forEach((data) => {
+      const { race_or_ethnicity: race } = data;
+      const count = parseInt(data.count, 10);
+      supervisionDataPoints.push([labelStringConversion[race], count]);
```
jessex: Same comment here as in the recent PR about not using arrays for structuring object-like information. This and the parallel block above should output arrays of objects.

src/components/charts/revocations/RevocationProportionByRace.js
```diff
@@ -27,74 +27,114 @@ const ND_RACE_PROPORTIONS = {
...
+        supervisionDataPoints.push([race, 0]);
       }
     });
 
-    const total = dataPoints.map((element) => element[1]).reduce(
+    const totalRevocations = revocationDataPoints.map((element) => element[1]).reduce(
+      (previousValue, currentValue) => (previousValue + currentValue),
+    );
+
+    const totalSupervisionPopulation = supervisionDataPoints.map((element) => element[1]).reduce(
```
jessex: These reduce objects have enough complexity that it might be worth extracting out to a lil' helper to share between the two.

src/views/Revocations.js
```diff
@@ -297,11 +297,17 @@ const Revocations = () => {
             <div className="bd bgc-white p-20">
               <div className="layers">
                 <div className="layer w-100 pX-20 pT-20">
-                  <h4 className="lh-1">Revocations by race compared to ND population</h4>
+                  <h4 className="lh-1">Revocations and supervision by race compared to ND population</h4>
```
jessex: We should make this a little bit more precise. Maybe "Revocations and admissions to supervision by race..."?

src/views/Revocations.js
```diff
@@ -316,6 +322,7 @@ const Revocations = () => {
                     <div>
                       <ul>
                         <li>Revocations include all instances of a person being incarcerated because their supervision was revoked for a behavioral violation.</li>
+                        <li>The supervision population includes anyone on probation or parole in North Dakota at any point during the time period.</li>
```
jessex: Pursuant to my previous updates to be more specific about supervision calculations using person as the unit of analysis instead of case, a slight update might be, "The supervision population counts people on ..."?

server/core/demoData/revocations_by_race_and_ethnicity_60_days.json
jjdressel11: Yep! Fake numbers.

src/views/Revocations.js
jjdressel11: These aren't admissions to supervision. It's a count of anyone with a supervision period that was active in the last 60 days.

What if we called this "Revocations by race" and then it's clear that the other two datasets are there for context?

jessex: That sounds fair!

src/views/Revocations.js
jjdressel11: Updated the text for "Revocations include all instances of a person" in multiple places in this file as well.

PR #81, opened 2019-09-30
Redirects to the Snapshots view instead of the home page

jessex:
When authenticated, it redirects to the Snapshots view. When you log in at the home page, it already successfully redirected to the Snapshots view on successful login. But you could still navigate directly back `dashboard.recidiviz.org/` and it would take you back to the home page, which is also what happens when you just open the dashboard naturally (no particular view) in your browser. Now if you attempt to do either of these, it will redirect to Snapshots successfully.

### DIFF
--- a/src/App.js
+++ b/src/App.js
@@ -1,41 +1,42 @@
-import React, { useEffect, useState } from "react";
-import { BrowserRouter as Router, Route, Switch } from "react-router-dom";
+import React, { useEffect, useState } from 'react';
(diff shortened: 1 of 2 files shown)


src/App.js
```diff
@@ -83,7 +84,9 @@ const App = () => {
             <div className={containerClass}>
               <TopBar pathname={window.location.pathname} />
               <Switch>
-                <Route path="/" exact component={Home} />
+                <Route exact path="/">
```
jessex: All of the other changes in this PR were lint fixes that Atom complained about. This is the only functional change for the redirect.

PR #90, opened 2019-10-03
Revocations by office and officers by office charts

jjdressel11:
In this PR:
- Bringing in two new files from cloud storage: revocations_by_site_id_60_days.json and site_offices.json (Note: site_offices.json is a static file with metadata about the ND P&P offices)
- New chart for revocation count by the site of the terminating officer
- Revocations by officer now shows officers of each office, with a toggle to select the office
- Interactions between these two charts where clicking on a site id of the map will change the displayed dataset of the officer revocation chart


src/components/charts/reincarcerations/AdmissionsVsReleases.js
```diff
@@ -112,7 +112,7 @@ const AdmissionsVsReleases = (props) => {
   const header = document.getElementById(props.header);
 
   if (header && mostRecentValue) {
-    const title = `The ND facilities <b style='color:#809AE5'>${direction} by ${mostRecentValue} people</b> this month.`;
+    const title = `The ND facilities <b style='color:#809AE5'>${direction} by ${Math.abs(mostRecentValue)} people</b> this month.`;
```
jessex: Woo, good fix.

src/components/charts/revocations/RevocationCountByOfficer.js
```diff
@@ -27,18 +27,83 @@ const RevocationCountByOfficer = (props) => {
   const [newOffenseDataPoints, setNewOffenseDataPoints] = useState([]);
   const [technicalDataPoints, setTechnicalDataPoints] = useState([]);
   const [unknownDataPoints, setUnknownDataPoints] = useState([]);
-
+  const [allChartData, setAllChartData] = useState({});
+  const [visibleOfficeName, setvisibleOfficeName] = useState('');
   const chartId = 'revocationsByOfficer';
 
+  function setDataForVisibleOffice(dataPoints, visibleOffice) {
```
jessex: This function needs a docstring explaining the format of the inputs and its side-effects (i.e. the various states it sets). Without that information it's hard to understand how this operates internally.

src/components/charts/revocations/RevocationCountByOfficer.js
```diff
@@ -27,18 +27,83 @@ const RevocationCountByOfficer = (props) => {
...
+    const officerLabels = [];
+    const violationArrays = {
+      ABSCONDED: [],
+      FELONY: [],
+      TECHNICAL: [],
+      UNKNOWN_VIOLATION_TYPE: [],
+    };
+
+    const visibleOfficeData = dataPoints[visibleOffice];
+    const sortedDataPoints = visibleOfficeData.sort((a, b) => (
```
jessex: The officer id is a string and not an integer, right? If so, sorting by string officer ids may yield an order like, "13, 145, 24, 269, 28, ...". If the order is important, then we should probably sort by the integer form of the officer id.

src/components/charts/revocations/RevocationCountByOfficer.js
```diff
@@ -27,18 +27,83 @@ const RevocationCountByOfficer = (props) => {
...
+      a.officerID - b.officerID));
+
+    for (let i = 0; i < sortedDataPoints.length; i += 1) {
+      officerLabels.push(sortedDataPoints[i].officerID);
+      const data = sortedDataPoints[i].violationsByType;
```
jessex: Can we call `data` something like `revocationCountsByType`?

src/components/charts/revocations/RevocationCountByOfficer.js
```diff
@@ -27,18 +27,83 @@ const RevocationCountByOfficer = (props) => {
...
   const [technicalDataPoints, setTechnicalDataPoints] = useState([]);
   const [unknownDataPoints, setUnknownDataPoints] = useState([]);
-
+  const [allChartData, setAllChartData] = useState({});
+  const [visibleOfficeName, setvisibleOfficeName] = useState('');
   const chartId = 'revocationsByOfficer';
 
+  function setDataForVisibleOffice(dataPoints, visibleOffice) {
+    const officerLabels = [];
+    const violationArrays = {
```
jessex: Another naming nit: this is an example of Hungarian notation, which we want to avoid (including the type in the variable name). How about "officerViolationCountsByType"?

src/components/charts/revocations/RevocationCountByOfficer.js
```diff
@@ -53,41 +118,38 @@ const RevocationCountByOfficer = (props) => {
...
-      if (officerID !== 'OFFICER_UNKNOWN') {
-        dataPoints.push({
-          officerID,
-          violationsByType,
-          overallRevocationCount,
-        });
+      let officeName = offices[parseInt(officeId, 10)];
+      if (officeName && officerID !== 'OFFICER_UNKNOWN') {
+        officeName = officeName.replace(' ', '-');
+        if (dataPoints[officeName] == null) {
```
jessex: I think it would be a little neater to do this:
```
if (dataPoints[officeName] == null) {
    dataPoints[officeName] = [];
}
dataPoints[officeName].push({
  officerID,
  violationsByType,
  overallRevocationCount,
});
```

src/components/charts/revocations/RevocationCountByOfficer.js
```diff
@@ -53,41 +118,38 @@ const RevocationCountByOfficer = (props) => {
...
+        $(`#${dropdownId}-${officeName}`).addClass('disabled');
+      }
+    });
 
-    setChartLabels(officerLabels);
-    setAbsconsionDataPoints(violationArrays.ABSCONDED);
-    setNewOffenseDataPoints(violationArrays.FELONY);
-    setTechnicalDataPoints(violationArrays.TECHNICAL);
-    setUnknownDataPoints(violationArrays.UNKNOWN_VIOLATION_TYPE);
+    // Show data for the first office name that has data
```
jessex: Is there a way to see the officers across all offices, i.e. statewide and not for a particular office? I think there should be a way to see this. It would be what you see when we first load, and the first option in the dropdown would let you select statewide.

src/components/charts/revocations/RevocationCountByOfficer.js
```diff
@@ -53,41 +118,38 @@ const RevocationCountByOfficer = (props) => {
...
 
-      if (officerID !== 'OFFICER_UNKNOWN') {
-        dataPoints.push({
-          officerID,
-          violationsByType,
-          overallRevocationCount,
-        });
+      let officeName = offices[parseInt(officeId, 10)];
+      if (officeName && officerID !== 'OFFICER_UNKNOWN') {
+        officeName = officeName.replace(' ', '-');
```
jessex: Why do we need to do these replace calls?

src/components/charts/revocations/RevocationsByOffice.js
```diff
@@ -0,0 +1,252 @@
...
+import geographyObject from '../../../assets/static/maps/us_nd.json';
+import { COLORS } from '../../../assets/scripts/constants/colors';
+import { configureDownloadButtons } from '../../../assets/scripts/utils/downloads';
+
+const chartId = 'revocationsByOffice';
+const centerNDLong = -100.5;
+const centerNDLat = 47.3;
+
+function xOffsetForOfficeTitle(office) {
+  const revLabelOffset = (office.revocationCount > 9) ? 35 : 25;
```
jessex: I think we should extract _all_ of these magical integers into well-named constants. It's pretty foreboding to open the file and just see a lot of integers and feel like changing any of them could do anything.

src/components/charts/revocations/RevocationsByOffice.js
```diff
@@ -0,0 +1,252 @@
...
+  }
+
+  if (office.titleSide === 'right') {
+    return office.revocationCount + (office.officeName.length * 7) + revLabelOffset;
+  }
+
+  return 0;
+}
+
+function yOffsetForOfficeTitle(office) {
```
jessex: We should document what these functions are doing, and whether their behavior (including the integers used therein) have any assumptions about the browser or device in use.

src/components/charts/revocations/RevocationsByOffice.js
```diff
@@ -0,0 +1,252 @@
...
+  }
+
+  return 10;
+}
+
+function colorForMarker(office) {
+  return (office.revocationCount > 0) ? COLORS['red-standard'] : COLORS['grey-600'];
+}
+
+const officeClicked = (evt) => {
```
jessex: evt --> event

src/components/charts/revocations/RevocationsByOffice.js
```diff
@@ -0,0 +1,252 @@
...
+    this.officerDropdownId = this.props.officerDropdownId;
+    this.offices = {};
+    this.officeIds = [];
+
+    // Load office metadata
+    this.officeData.forEach((officeData) => {
+      const {
+        site_id: officeId,
+        site_name: officeName,
+        long, lat,
```
jessex: I know it's syntactically permissible, but I don't love having a mix of manual and automatic keys in a single object. Can we do `long: long, lat: lat` without eslint complaining? If not, let's at least shift these below `title_side` to be a bit more visually consistent.

src/components/charts/revocations/RevocationsByOffice.js
```diff
@@ -0,0 +1,252 @@
...
+        title_side: titleSide,
+      } = officeData;
+
+      const office = {
+        officeName, coordinates: [long, lat], titleSide,
```
jessex: Same comment here as above, re: keys and ordering.

src/components/charts/revocations/RevocationsByOffice.js
```diff
@@ -0,0 +1,252 @@
...
+    });
+
+    // Load revocation data for each office
+    this.chartDataPoints = [];
+    this.officeIdsWithData = [];
+    this.revocationsByOffice.forEach((data) => {
+      const {
+        site_id: officeId,
+        absconsion_count: absconsionCount,
+        felony_count: felonyCount, technical_count: technicalCount,
```
jessex: Let's drop `technical_count` to a new line.

src/components/charts/revocations/RevocationsByOffice.js
```diff
@@ -0,0 +1,252 @@
...
+        unknown_count: unknownCount,
+      } = data;
+
+      const revocationCountNum = parseInt(absconsionCount, 10)
```
jessex: I think the `parseInt(whatever, 10)` calls feel a bit verbose. Can we create a helper somewhere like `toInt(whatever)` and use that in all components that call `parseInt`?

src/components/charts/revocations/RevocationCountByOfficer.js
jessex: If we're doing this to match ids in the HTML, we should extract the replace calls out to helpers, like `toHtmlFriendly` and `toHumanReadable` or some such.

src/components/charts/revocations/RevocationsByOffice.js
```diff
@@ -0,0 +1,252 @@
...
+  }
+
+  componentDidMount() {
+    const exportedStructureCallback = () => (
+      {
+        metric: 'Revocations by P&P office',
+        series: [],
+      });
+
+    const dataArray = [];
```
jessex: Variable names like `dataArray` and `data` reduce readability, since they don't have any inherent meaning except in pretty specific circumstances, like really general purpose code. In this case, we know that these data points mean specific things, and we should name them accordingly.

src/views/Revocations.js
```diff
@@ -193,6 +194,159 @@ const Revocations = () => {
...
+                    <div className="mb-0">
+                      <button className="btn btn-link collapsed pL-0" type="button" data-toggle="collapse" data-target="#collapseMethodologyRevocationsByOffice" aria-expanded="true" aria-controls="collapseMethodologyRevocationsByOffice">
+                        <h6 className="lh-1 c-blue-500 mb-0">Methodology</h6>
+                      </button>
+                    </div>
+                  </div>
+                  <div className="collapse" id="collapseMethodologyRevocationsByOffice" aria-labelledby="methodologyHeadingRevocationsByOffice" data-parent="#methodologyRevocationsByOffice">
+                    <div>
+                      <ul>
+                        <li>Revocations include all instances of a person being incarcerated because their supervision was revoked for a behavioral violation.</li>
```
jessex: Looks like these methodology bullet points are too long and need to be linebroken.

src/components/charts/revocations/RevocationCountByOfficer.js
jjdressel11: It was oddly sorting correctly... but I converted to ints before sorting just to be safe.

src/components/charts/revocations/RevocationCountByOfficer.js
jjdressel11: On load it now shows "All Officers"!

src/components/charts/revocations/RevocationCountByOfficer.js
jjdressel11: Done!

src/components/charts/revocations/RevocationsByOffice.js
jjdressel11: (Ignoring the fact that our dashboard is a disaster on mobile currently)
I tested resizing the window and looked on safari, and these values don't seem to be browser or size-of-window specific, which is great.

src/components/charts/revocations/RevocationsByOffice.js
jjdressel11: Renamed to `longValue` and `latValue` to keep the accessing of all values consistent.

src/components/charts/revocations/RevocationsByOffice.js
jjdressel11: Updated everywhere!

src/components/charts/revocations/RevocationsByOffice.js
jjdressel11: (Ignoring the fact that our dashboard is a disaster on mobile currently)
I tested resizing the window and looked on safari, and these values don't seem to be browser or size-of-window specific, which is great.

src/components/charts/revocations/RevocationCountByOfficer.js
```diff
@@ -159,6 +172,7 @@ const RevocationCountByOfficer = (props) => {
   const chart = (
     <Bar
       id={chartId}
+      redraw
```
jessex: I think given how non-intuitive this was, it would be worth adding an inline comment explaining why this is so important.

jjdressel11: Great call!

Issue #95, opened 2019-10-03
Update the hover background color of charts


jjdressel11:
In most cases, we don't want the color of the data point, dataset, or the tooltip legend item to change color on hover. Set the `pointHoverBackgroundColor` property for line charts, and the `hoverBackgroundColor` property for bar charts.

Issue #94, opened 2019-10-03
Make all chart tooltips more in-line with our color scheme


jjdressel11:
The black tooltip background colors don't fit well with the dashboard color scheme. A gray, or a white with a light gray outline or drop-shadow would be better.

PR #98, opened 2019-10-08
Updating tooltip and hover background colors

jjdressel11:
Closes: #95 
Closes: #94 

In this PR:
- Changes the tooltip background color to be a dark grey instead of black to be more in-line with our dashboard color scheme
- Explicitly sets the background color on hover for all charts so that the color does not change on hover
- Fix in the calculation for the title offset for `RevocationsByOffice` chart


src/components/charts/reincarcerations/AdmissionsVsReleases.js
```diff
@@ -40,8 +41,18 @@ const AdmissionsVsReleases = (props) => {
 
     const sorted = sortAndFilterMostRecentMonths(dataPoints, 6);
 
+    const colorsForValues = [];
+    sorted.forEach((dataPoint) => {
```
jessex: Why did we need/want to move this logic up to here?

src/components/charts/reincarcerations/ReincarcerationRateByReleaseFacility.js
```diff
@@ -65,8 +65,7 @@ const ReincarcerationRateByReleaseFacility = (props) => {
         datasets: [{
           label: 'Reincarceration rate',
           backgroundColor: COLORS['blue-standard'],
-          borderColor: COLORS['blue-standard'],
-          borderWidth: 1,
```
jessex: Is there a default border width that this is now changing to?

src/components/charts/revocations/AdmissionTypeProportions.js
```diff
@@ -74,6 +74,13 @@ const AdmissionTypeProportions = (props) => {
             COLORS_FIVE_VALUES[3],
             COLORS_FIVE_VALUES[2],
           ],
+          hoverBorderColor: [
+            COLORS_FIVE_VALUES[1],
```
jessex: Can you remind me why the order is 1,0,3,2?

src/components/charts/revocations/RevocationsByCounty.js
```diff
@@ -50,13 +50,18 @@ function revocationCountForCounty(chartDataPoints, countyName) {
   return 0;
 }
 
-function colorForCounty(chartDataPoints, countsByCounty, countyName, maxValue) {
+function colorForCounty(chartDataPoints, countsByCounty, countyName, maxValue, dark) {
```
jessex: We should make `dark` more obviously a boolean, e.g. `is_dark` or `should_be_dark`.

src/components/charts/revocations/RevocationsByOffice.js
```diff
@@ -202,10 +201,6 @@ class RevocationsByOffice extends Component {
     configureDownloadButtons(chartId, downloadableDataFormat,
       Object.keys(this.chartDataPoints),
       document.getElementById(chartId), exportedStructureCallback);
-
-    setTimeout(() => {
-      ReactTooltip.rebuild();
-    }, 100);
```
jessex: What was this doing before?

src/components/charts/reincarcerations/AdmissionsVsReleases.js
jjdressel11: For some reason the `hoverBackgroundColor` was freaking out if I tried to use the exact same indexing logic that the `backgroundColor` was using. Also, since these two use the same value, moving this logic here prevents calculating these unnecessarily.

src/components/charts/reincarcerations/ReincarcerationRateByReleaseFacility.js
jjdressel11: Default `borderWidth` is 0.

src/components/charts/revocations/AdmissionTypeProportions.js
jjdressel11: Keeps consistency with the other charts, so that technical and unknown are the same colors that they are in the other charts.

src/components/charts/revocations/RevocationsByOffice.js
jjdressel11: This was leftover from the revocation by county chart that has a tooltip. Should not have been in here.

src/components/charts/revocations/AdmissionTypeProportions.js
jessex: Maybe add a comment above all of these indicating that's why the order is at it is.

Issue #80, opened 2019-09-27
Fix issue where dashboard crashes if a chart file is empty


jjdressel11:
If the file corresponding to the data for a given chart is empty, the app crashes because the chart tries to iterate on the data being sent to it without validating that the data object isn't empty. This can happen validly in prod if a query returns no results, so we should handle the case of a chart's data being empty.

PR #100, opened 2019-10-09
Handles empty data in charts

jjdressel11:
Closes: #80 

In this PR:
- Handles any case where a chart is sent empty data from the API so that an empty chart is displayed instead of the dashboard app crashing
- For charts that display count values over months, adds counts of 0 for any month that is missing data
- Does not add values to the snapshot charts if there are missing months, because those are not count charts and on those charts a value of 0 is significant

Filed: https://github.com/Recidiviz/recidiviz-data/issues/2513

### DIFF
--- a/src/utils/dataOrganizing.js
+++ b/src/utils/dataOrganizing.js
@@ -14,8 +14,6 @@
 // You should have received a copy of the GNU General Public License
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 // =============================================================================
-import { toInt } from './variableConversion'
-
 const TRANSITIONAL_FACILITY_FILTERS = {
   US_ND: ['FTPFAR', 'GFC', 'BTC', 'FTPMND', 'MTPFAR', 'LRRP', 'MTPMND'],
   US_DEMO: ['GHI', 'PQR', 'VWX'],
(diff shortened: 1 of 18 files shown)


src/utils/dataOrganizing.js
```diff
@@ -75,9 +74,60 @@ function sortAndFilterMostRecentMonths(unsortedDataPoints, monthCount) {
   return filterMostRecentMonths(sortedData, monthCount);
 }
 
+/**
+ * Adds a dictionary for any month in the last `monthCount` number of months
+ * that is missing data, where the the value for the `valueKey` property is `emptyValue`.
```
jessex: Nit: "the the value"

src/utils/dataOrganizing.js
```diff
@@ -75,9 +74,60 @@ function sortAndFilterMostRecentMonths(unsortedDataPoints, monthCount) {
   return filterMostRecentMonths(sortedData, monthCount);
 }
 
+/**
+ * Adds a dictionary for any month in the last `monthCount` number of months
```
jessex: I'd be a bit more explicit about return type here. "Returns a new list of data points consisting of the given data points and new data points appended for the last ...

jjdressel11: Updated.

PR #105, opened 2019-10-15
Dashboard tab for descriptive FTR referral stats

jjdressel11:
Note: can only be run in demo mode while @terinpw is working on the calculations.

In this PR:
- New Free Through Recovery tab
- Charts for the following
     - Referral counts by month
     - Referrals and supervision pop by race
     - Referrals and supervision pop by gender
     - Referrals and supervision pop by age
     - Referrals and supervision pop by LSI-R

Terin is working on the calculations and the methodology text for each of these charts right now. This won't be merged until Terin is done with the calculations for each of these charts and the methodology text has been approved. (Methodology sections say `METHODOLOGY` right now)


server.js
```diff
@@ -75,7 +75,7 @@ if (isDemoMode) {
   checkJwt = function (req, res, next) { next(); };
 }
 
-app.get('/api/programEval', checkJwt, api.programEval);
+app.get('/api/free-through-recovery', checkJwt, api.freeThroughRecovery);
```
jessex: I think a better API structure would be to group specific instances of a general concept under a more generalized API path. So Free Through Recovery is a specific program under evaluation. We could have API results that are in broadly one of two camps: they are in a form that is general to understanding any program but happen to be about FTR in this instance, or they are in a form that is specific to understanding FTR. 

If the former, I think they should be provided through a general purpose `/api/programEvaluation` endpoint but with a query parameter of something like `?programName=free-through-recovery`. Then that API returns results that are in the same format no matter what program you request, but the content is specific to FTR.

If the latter, I think they would be provided through a dedicated endpoint grouped under the program evaluation API, like `/api/programEvaluation/freeThroughRecovery`. That endpoint would return results that are specific to FTR in both content and structure, i.e. it returns information that is only relevant to FTR and would not be relevant to our understanding of other programs.

So the main question to ask here is: what information are we visualizing about FTR? Is it general, or FTR-specific?

src/App.js
```diff
@@ -108,6 +109,7 @@ const App = () => {
                 <PrivateRoute path="/snapshots" component={Snapshots} />
                 <PrivateRoute path="/revocations" component={Revocations} />
                 <PrivateRoute path="/reincarcerations" component={Reincarcerations} />
+                <PrivateRoute path="/free-through-recovery" component={FreeThroughRecovery} />
```
jessex: Similar to my previous comment, I wonder if the app path should be something like `dashboard.recidiviz.org/programEvaluation/freeThroughRecovery`. This would keep route structure predictable as more programs are brought online across many states, and make it easier to build automated tooling in the future that relied on looking at these routes to, say, provide a catalog of programs.

src/components/charts/freeThroughRecovery/FtrReferralCountByMonth.js
```diff
@@ -0,0 +1,124 @@
+// Recidiviz - a data platform for criminal justice reform
```
jessex: Since FTR is specific to the state of North Dakota, I think it should live inside of a package dedicated to that state. Something like `src/components/charts/us_nd/free_through_recovery`. Snake case is the closest to a convention on this apparently, and my use of camelCase with `server/core/demoData` was a mistake.

src/components/charts/freeThroughRecovery/FtrReferralsByRace.js
```diff
@@ -0,0 +1,260 @@
...
+
+import React, { useState, useEffect } from 'react';
+
+import { HorizontalBar } from 'react-chartjs-2';
+import { COLORS_FIVE_VALUES, COLORS } from '../../../assets/scripts/constants/colors';
+import { sortByLabel } from '../../../utils/dataOrganizing';
+import { configureDownloadButtons } from '../../../assets/scripts/utils/downloads';
+import { raceValueToHumanReadable, toInt } from '../../../utils/variableConversion';
+
+const ND_RACE_PROPORTIONS = {
```
jessex: Since these are used in multiple places now, we should either have a package with state-specific constants, or pull this from the API like we do with site offices. I'd prefer the latter. Just have a file with the race proportions for a given state. In time, we can update those breakdowns in the app easily without having to redeploy, and can even automate their update based on new census results, for example.

src/views/FreeThroughRecovery.js
```diff
@@ -0,0 +1,379 @@
...
+    return <Loading />;
+  }
+
+  return (
+    <main className="main-content bgc-grey-100">
+      <div id="mainContent">
+        <div className="row gap-20 pos-r">
+
+          {/* #FTR referral counts by month chart ==================== */}
+          <div className="col-md-6">
```
jessex: We should create something like a "visualization card component" which starts roughly here and includes the actual chart component (dynamically passed into the component as an argument, like how we pass `apiData` as an argument into a chart component), the chart headline (string argument), the methodology component (string array argument), and the export component (might need like an "export id" argument). Would make each page view much simpler to compose and much shorter, and ensure visual consistency across views.

Would you mind filing an issue for that?

src/components/SideBar.js
```diff
@@ -93,6 +93,14 @@ const SideBar = () => {
               <span className="title">Reincarcerations</span>
             </a>
           </li>
+          <li className="nav-item">
+            <a className="sidebar-link" href="free-through-recovery">
+              <span className="icon-holder">
+                <i className="c-red-500 ti-location-arrow" />
```
jessex: Since reincarcerations is also red, maybe this should be a different color. Is there a dominant color in the FTR branding?

src/views/FreeThroughRecovery.js
```diff
@@ -0,0 +1,379 @@
...
+                        Last 60 days
+                      </span>
+                    </div>
+                  </div>
+                </div>
+              </div>
+            </div>
+          </div>
+
+          {/* #FTR Referrals by LSI-R chart ==================== */}
```
jessex: I feel like this is more interesting than the gender breakdown, at least, and maybe also the age breakdown. Can we pull this further up above the fold?

src/views/FreeThroughRecovery.js
```diff
@@ -0,0 +1,379 @@
...
+                        Last 60 days
+                      </span>
+                    </div>
+                  </div>
+                </div>
+              </div>
+            </div>
+          </div>
+
+          {/* #FTR Referrals by Gender chart ==================== */}
```
jessex: Is there a reason the gender chart is structured in an inverted fashion relative to the age and LSI-R charts? Those have the different buckets of the dimension along the x-axis, with separate bars for referrals versus population. This one, on the other hand, has referrals and population along the x-axis with stacked bars for the genders. Is that a design choice owing to the fact that there are only two subsets of the population being visualized in this one (since ND doesn't have non-binary genders in the dataset)?

src/components/charts/freeThroughRecovery/FtrReferralsByLsir.js
```diff
@@ -0,0 +1,187 @@
...
+  const [ftrReferralCounts, setFtrReferralCounts] = useState([]);
+  const [stateSupervisionCounts, setStateSupervisionCounts] = useState([]);
+
+  const chartId = 'ftrReferralsByLsir';
+
+  const processResponse = () => {
+    const { ftrReferralsByLsir } = props;
+    const { supervisionPopulationByLsir } = props;
+
+    const lsirScoreBuckets = ['0-23', '24-29', '30-38', '39+'];
```
jessex: I think that since 34+ is what they _want_ the referral threshold to be, we should have a bucket that ends on that line. Is there a source for the buckets we've chosen here or did we choose them ourselves? If the latter, what if we just split this bucket into 30-34 and 35-38.

server.js
jjdressel11: Right now we are visualizing only FTR referral information, which I think is pretty specific to FTR. I originally had everything coming from the programEval api endpoint, but then changed the tab name to be just `Free Through Recovery` because at this stage we are only showing them descriptive stats about FTR and I don't want them to be confused about why there isn't information about any other program currently and didn't want it to fall under `Program Evaluation` until we were providing any kind of outcomes-based measurement. 

Eventually we will be evaluating FTR, so it should be under program eval. I think we should go with `/api/programEvaluation/freeThroughRecovery` because it has such a unique structure and outcomes measurement.

src/App.js
jjdressel11: Because it's so ND specific, should we have the app path be: 
`dashboard.recidiviz.org/programEvaluation/us_nd/freeThroughRecovery`?

src/components/charts/freeThroughRecovery/FtrReferralsByLsir.js
jjdressel11: I used these [LSI-R cut scores](https://drive.google.com/open?id=1W_WgvF80JBZCDAB-K_NS_LogM1KODWAI) that they gave us.

src/components/charts/freeThroughRecovery/FtrReferralCountByMonth.js
jjdressel11: I think it's better to organize like this:
`src/components/charts/programEvaluation/us_nd/free_through_recovery` ?
So that we have state-specific charts within our broader organization of shared components.

src/components/charts/freeThroughRecovery/FtrReferralsByRace.js
jjdressel11: Agreed! Moved to the api.

src/views/FreeThroughRecovery.js
jjdressel11: Oooo yes I love this. Filed #107

src/views/FreeThroughRecovery.js
jjdressel11: Moved to be chart #3

src/views/FreeThroughRecovery.js
jjdressel11: This was done for a few reasons. I think the age and LSI-R scores should be on the x-axis as opposed to the Referrals and Supervision rates because they are values that increase linearly, whereas the gender and race values are discrete (unordered) buckets. The gender chart is flipped this way because it only has two subsets and I thought having it stacked vertically made it easier to see differences in the proportions for only two groups.

I'm having Clementine do UXR on these charts on the ground in ND to validate the design choices I've made here!

src/App.js
jessex: This came up in that document with Clem the other day, re: our plan for multi-tenancy experience. We're going to try to not have state codes in the application paths. It should feel like the app you're using is totally dedicated to your jurisdiction/agency, and adding codes for specific states in the paths diminishes that effect. I think the path you're describing might make more sense for a backend API but not a frontend API.

src/components/charts/freeThroughRecovery/FtrReferralCountByMonth.js
jessex: That sounds good to me!

src/views/FreeThroughRecovery.js
jessex: Got it, this sounds like a good start then, and we'll see if UXR validates or invalidates. Thanks for the explanation! Makes sense.

src/App.js
jjdressel11: Great point. Will make this `dashboard.recidiviz.org/programEvaluation/freeThroughRecovery`.

server.js
jjdressel11: You're right that visualization for referrals to other programs will be structured in basically the same way. Also I buy that descriptive stats are a subset of program eval metrics.

Seems to me like we agree on the following:
- Keep the tab name as "Free Through Recovery"
- Have the API path be something under `/api/programEvaluation/`
- We should strive to never change API paths

Now we need to decide what we want the full api path to be:
- `/api/programEvaluation/freeThroughRecovery`
-`/api/programEvaluation?programName=free-through-recovery`
-`/api/programEvaluation/us_nd/freeThroughRecovery`
-`/api/programEvaluation?stateCode=us_nd&programName=free-through-recovery`

Or some other variation of any of these.

src/components/charts/freeThroughRecovery/FtrReferralsByLsir.js
jjdressel11: Happy to make this change once @terinpw confirms!

terinpw: It's >=34 for qualification into FTR. 

I'm torn - as it's a little weird to further divide up the buckets that ND already has and is familiar with - but again as you mentioned this is FTR specific, so ultimately I think I do like the split.

I can update the query in my PR

server.js
jessex: Thinking broadly, we will have APIs for returning program evaluation metrics, and they will _always_ be filtered by state code (this will be true of nearly all APIs in time). We will have APIs for returning program evaluation metrics that are not specific to a particular program, and we will have others that _are_ specific to a particular program. It is possible that we will have program names that are the same across state lines, so it will be important to specify both state code and program name when looking at a particular program.

With all of that in mind, I'd argue for an API like this:
- /api/snapshots/{stateCode}
- /api/reincarcerations/{stateCode}
- /api/revocations/{stateCode}
- /api/programEvaluation/{stateCode}
- /api/programEvaluation/{stateCode}/{programName}

So in this case, `/api/programEvaluation/us_nd/freeThroughRecovery`. What do you think?

src/components/charts/freeThroughRecovery/FtrReferralsByLsir.js
jessex: UXR will be the final arbiter here. I suggest we go with the further split but let's run it by Andrew or Clem tomorrow in person to see what they think before we merge. If UXR next week tells us to do the opposite, we'll change it.

server.js
jjdressel11: Looks good to me!

src/views/FreeThroughRecovery.js
```diff
@@ -0,0 +1,414 @@
...
+                      <button className="btn btn-link collapsed pL-0" type="button" data-toggle="collapse" data-target="#collapseMethodologyFtrReferralCountByMonth" aria-expanded="true" aria-controls="collapseMethodologyFtrReferralCountByMonth">
+                        <h6 className="lh-1 c-blue-500 mb-0">Methodology</h6>
+                      </button>
+                    </div>
+                  </div>
+                  <div id="collapseMethodologyFtrReferralCountByMonth" className="collapse" aria-labelledby="methodologyHeadingFtrReferralCountByMonth" data-parent="#methodologyFtrReferralCountByMonth">
+                    <div>
+                      <ul>
+                        <li>
+                          Referral counts include the number of people who were
```
terinpw: @jessex @jjdressel11 added methodology for all the graphs here. Take a look and let me know if you have thoughts/improvements.

src/assets/scripts/utils/strings.js
```diff
@@ -35,8 +35,35 @@ function normalizeLabel(str) {
   return capitalizeWords(label);
 }
 
+function normalizeAppPathToTitle(str) {
+  let noDash = replaceAll(str, '-', ' ');
+
+  while (noDash.indexOf('/', 1) !== -1) {
```
jessex: Instead of having a block for dealing with dashes and a block for dealing with slashes, would you get the same desired effect by replacing all punctuation with whitespace and then stripping whitespace from the ends?

src/assets/scripts/utils/strings.js
```diff
@@ -35,8 +35,35 @@ function normalizeLabel(str) {
...
+  }
+
+  let noSlash = replaceAll(noDash, '/', '');
+  if (!noSlash) {
+    noSlash = 'Snapshots';
+  }
+
+  const capitalLetters = noSlash.match(/[A-Z]/g);
+  let noSlashWithSpaces = '';
+  for (let i = 0; i < noSlash.length; i += 1) {
```
jessex: The only thing about this approach is it won't work as expected for acronyms, which we fortunately don't have yet. The Lodash library has a `startCase` function I've used in the past for this, but I don't know if that works for acronyms, either.

src/assets/scripts/utils/strings.js
```diff
@@ -35,8 +35,35 @@ function normalizeLabel(str) {
...
+  let noDash = replaceAll(str, '-', ' ');
+
+  while (noDash.indexOf('/', 1) !== -1) {
+    const nextSlash = noDash.indexOf('/', 1);
+    noDash = noDash.substring(nextSlash);
+  }
+
+  let noSlash = replaceAll(noDash, '/', '');
+  if (!noSlash) {
+    noSlash = 'Snapshots';
```
jessex: I think encoding Snapshots here as the default choice is unnecessary coupling. I'd rather return null and have the caller decide what to do in that case. The caller in this case would default to "Snapshots" but the top bar is better positioned to make that call than this specific utility class.

src/views/FreeThroughRecovery.js
```diff
@@ -0,0 +1,414 @@
...
+                      <button className="btn btn-link collapsed pL-0" type="button" data-toggle="collapse" data-target="#collapseMethodologyFtrReferralsByRace" aria-expanded="true" aria-controls="collapseMethodologyFtrReferralsByRace">
+                        <h6 className="lh-1 c-blue-500 mb-0">Methodology</h6>
+                      </button>
+                    </div>
+                  </div>
+                  <div className="collapse" id="collapseMethodologyFtrReferralsByRace" aria-labelledby="methodologyHeadingFtrReferralsByRace" data-parent="#methodologyFtrReferralsByRace">
+                    <div>
+                      <ul>
+                        <li>
+                          People with multiple races/ethnicities are counted once for each race/ethnicity.
```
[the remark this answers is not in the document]
jessex: Also, we should be explicit that this is for the referrals and supervision population bars, not for the ND population as a whole.

src/assets/scripts/utils/strings.js
jjdressel11: Cross that bridge later?

src/assets/scripts/utils/strings.js
jjdressel11: The reason I do this in two different steps is that we remove dashes, isolate the end of the path, and then add spaces where there are uppercase letters. This assumes that our paths will be camel case, and that the end of the path should be the title. (I imagine we may need a better solution for this eventually. The end of the path won't always be what we want the title of the page to be.)

If we ever have a route with dashes, then replacing punctuation with whitespace would cause this issue: 

`/programEval/freeThroughRecovery`-> `freeThroughRecovery` -> `Free Through Recovery`

`/someRoute/something-with-dashes` -> `someRoute something with dashes` -> ?

src/assets/scripts/utils/strings.js
jessex: Sounds reasonable.

src/assets/scripts/utils/strings.js
jessex: Okay, fair enough. Let's stick with this for now.

PR #108, opened 2019-10-16
Updating methodology text

jjdressel11:
An update to all of the methodology sections on the Revocations page to explicitly clarify that we are counting revocations that lead to incarceration in a DOCR facility. Also an update to the methodology text regarding violations in response to user feedback.

### DIFF
--- a/src/views/Revocations.js
+++ b/src/views/Revocations.js
@@ -105,12 +105,17 @@ const Revocations = () => {
                       <ul>
                         <li>
                           Revocation counts include the number of people who were incarcerated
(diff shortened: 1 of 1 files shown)


jjdressel11:
Discussed this with @cawarren and he suggested keeping this in Methodology for now, adding dynamic titles to the charts (#109), and then adding on-hover glossary tooltips (#110) that would help clarify what we mean by, say, the term `revocations`. This would free up the dynamic headers to be more descriptive of the content in the chart.

PR #111, opened 2019-10-17
Added details to methodology for supervision race chart to describe handling of multiple races/ethnicities

terinpw:


src/views/Revocations.js
```diff
@@ -609,6 +609,12 @@ const Revocations = () => {
                           The race proportions for the population of North Dakota were taken from
                           the U.S. Census Bureau.
                         </li>
+                        <li>
+                          People with multiple races/ethnicities are counted
+                          once for each race/ethnicity. This means the total
+                          number of data points in these charts can be larger
+                          than the total number of people.
```
jessex: "...total number of people, if there are individuals with different recorded races/ethnicities from different data systems." ?

jjdressel11:
```
If an individual has more than one race or ethnicity recorded from different data systems, then they are
counted once for each unique race and ethnicity. This means that the total count in this chart may be
larger than the total number of individuals it describes. This does not apply to the ND Population values.
```

terinpw:
> One suggestion. Are there any other charts where we need to note this?

Nope, just the FTR ones that are in Julia's PR #105

Otherwise we don't have any other race/ethnicity based ones.

PR #113, opened 2019-10-22
Revocations by P&P Office tooltip on hover

### LINKED ISSUE #106: Revocation count by P&P office tooltip should be on hover
The design pattern of the county choropleth and the P&P office charts should be consistent.

Update the P&P office chart to have a tooltip so that that both are using some visual indicator (color, size) to signify impact, and then both require a hover to see the actual value.

jjdressel11:
Closes #106

Removes the labels from the Revocations by P&P Office chart and changes the info to be in a tooltip on hover. Also includes some updates to the demo data for the FTR age chart that was broken locally.

### DIFF
--- a/src/components/charts/revocations/RevocationsByCounty.js
+++ b/src/components/charts/revocations/RevocationsByCounty.js
@@ -130,9 +130,9 @@ class RevocationsByCounty extends Component {
         >
           <ZoomableGroup center={[centerNDLong, centerNDLat]} zoom={7} disablePanning>
             <Geographies geography={geographyObject}>
-              {(geographies, projection) => geographies.map((geography, i) => (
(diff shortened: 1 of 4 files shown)


src/components/charts/revocations/RevocationsByCounty.js
```diff
@@ -130,9 +130,9 @@ class RevocationsByCounty extends Component {
         >
           <ZoomableGroup center={[centerNDLong, centerNDLat]} zoom={7} disablePanning>
             <Geographies geography={geographyObject}>
-              {(geographies, projection) => geographies.map((geography, i) => (
+              {(geographies, projection) => geographies.map((geography) => (
                 <Geography
-                  key={i}
+                  key={geography.properties.NAME}
```
jjdressel11: This was an eslint change

PR #119, opened 2019-10-23
Revocation rate by county of residence

jjdressel11:
Updates the Revocation by County chart to be Revocation Rate by County of Residence.

### DIFF
--- a/server/core/demo_data/revocation_rate_by_county_60_days.json
+++ b/server/core/demo_data/revocation_rate_by_county_60_days.json
@@ -1,0 +1,16 @@
+{"state_code":"US_DEMO","county_code":"US_DEMO_BENSON","revocation_rate":"0.01515151515", "revocation_count": "1", "population_count": "66"}
+{"state_code":"US_DEMO","county_code":"US_DEMO_BOTTINEAU","revocation_rate":"0.03448275862", "revocation_count": "1", "population_count": "29"}
+{"state_code":"US_DEMO","county_code":"US_DEMO_BURLEIGH","revocation_rate":"0.02599179207", "revocation_count": 	"38", "population_count": "1462"}
+{"state_code":"US_DEMO","county_code":"US_DEMO_CASS","revocation_rate":"0.008102633356", "revocation_count": "12", "population_count": "1481"}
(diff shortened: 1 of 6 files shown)


server/core/demo_data/revocation_rate_by_county_60_days.json
```diff
@@ -0,0 +1,53 @@
+{"state_code":"US_DEMO","county_code":"US_DEMO_MCLEAN","revocation_count":"3","population_count":"64","revocation_rate":0.015625}
```
jessex: All of these are fake fixtures, yes?

src/components/charts/revocations/RevocationRateByCounty.js
```diff
@@ -0,0 +1,192 @@
...
+    return revocationRate.toFixed(2);
+  }
+
+  return 0;
+}
+
+function valueCountsForCounty(officeValues, countyName) {
+  const officeNumbers = officeValues[countyName.toUpperCase()];
+  if (officeNumbers) {
+    return officeNumbers.revocationCount.concat('/', officeNumbers.populationCount);
```
jessex: Can we use the more standard format of `${officeNumbers.revocationCount}/${officeNumbers.populationCount}`?

src/components/charts/revocations/RevocationRateByCounty.js
```diff
@@ -0,0 +1,192 @@
...
+  }
+
+  return '0/0';
+}
+
+function colorForCounty(chartDataPoints, countyName, maxValue, useDark) {
```
jessex: A couple of these functions look identical to the ones in the `RevocationsByCounty` chart component. Can we move those out to some set of, say, county choropleth utilities?

server/core/demo_data/revocation_rate_by_county_60_days.json
jjdressel11: Yep

jjdressel11:
> Looks good, @jjdressel11! I know we were discussing the headline format for this chart but it looks like we haven't added a headline yet. Is that to come later?

Double checked with Clem and she agreed that not adding a headline is okay for now. It's not clear what the header should be for this chart because the county with the highest rate is often a county with a very low supervision population - which is not as important to highlight in a takeaway header.

src/components/charts/revocations/RevocationRateByCounty.js
jjdressel11: ugh fine

PR #126, opened 2019-10-30
Changes data export to be CSV

### LINKED ISSUE #96: Change data export to be CSV
It's doubtful that anyone in criminal justice is experienced dealing with JSON. We should change the `Export data` functionality to export as a CSV instead.

jjdressel11:
## Description of the change

- Changes the `Export data` button to download as a .csv instead of .json
- Updates to the `label` value in a bunch of datasets to make the CSV labels meaningful
- Renames `AdmissionTypeProportions` component to `AdmissionCountsByType` to be more accurate
- Re-organizes how the officer violation data was being downloaded so that it's optimized for CSV

I've tested the `Export data` for every chart and it is working as expected. It should also be noted that this does not break the existing functionality to download as a JSON. So, we could support multiple download file types if we wanted to. 

## Type of change

## Related issues

Closes #96 

### Development

- [X] Lint rules pass locally

### Code review

### DIFF
--- a/src/assets/scripts/utils/downloads.js
+++ b/src/assets/scripts/utils/downloads.js
@@ -16,7 +16,7 @@
 // =============================================================================
 
 import downloadjs from 'downloadjs';
-import * as jsonexport from 'jsonexport/dist';
+import * as csvExport from 'jsonexport/dist';
 import { timeStamp } from './time';
 
 function downloadCanvasImage(canvas, filename, chartTitle) {
@@ -40,14 +40,11 @@ function downloadCanvasImage(canvas, filename, chartTitle) {
 }
 
 function downloadObjectAsCsv(exportObj, exportName) {
-  const jsonString = JSON.stringify(exportObj.series);
-  const jsonObj = [JSON.parse(jsonString)];
-
   const options = {
     mapHeaders: (header) => header.replace(/label|values./, ''),
   };
 
-  jsonexport(jsonObj, options, (err, csv) => {
+  csvExport(exportObj.series, options, (err, csv) => {
     if (err) throw err;
     const dataStr = `data:text/csv;charset=utf-8,${csv}`;
     const filename = `${exportName}.csv`;
--- a/src/assets/scripts/utils/downloads.js
+++ b/src/assets/scripts/utils/downloads.js
@@ -16,6 +16,7 @@
 // =============================================================================
 
 import downloadjs from 'downloadjs';
+import * as jsonexport from 'jsonexport/dist';
 import { timeStamp } from './time';
 
 function downloadCanvasImage(canvas, filename, chartTitle) {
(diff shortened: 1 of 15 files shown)


src/assets/scripts/utils/downloads.js
```diff
@@ -38,6 +39,22 @@ function downloadCanvasImage(canvas, filename, chartTitle) {
...
 
+function downloadObjectAsCsv(exportObj, exportName) {
+  const jsonString = JSON.stringify(exportObj.series);
+  const jsonObj = [JSON.parse(jsonString)];
+
+  const options = {
+    mapHeaders: (header) => header.replace(/label|values./, ''),
+  };
+
+  jsonexport(jsonObj, options, (err, csv) => {
```
jessex: I'm curious why we're using a json export library to export csv?

src/assets/scripts/utils/downloads.js
```diff
@@ -38,6 +39,22 @@ function downloadCanvasImage(canvas, filename, chartTitle) {
   downloadjs(data, filename, 'image/png;base64');
 }
 
+function downloadObjectAsCsv(exportObj, exportName) {
+  const jsonString = JSON.stringify(exportObj.series);
+  const jsonObj = [JSON.parse(jsonString)];
+
+  const options = {
+    mapHeaders: (header) => header.replace(/label|values./, ''),
```
jessex: What does this do?

src/assets/scripts/utils/downloads.js
```diff
@@ -38,6 +39,22 @@ function downloadCanvasImage(canvas, filename, chartTitle) {
   downloadjs(data, filename, 'image/png;base64');
 }
 
+function downloadObjectAsCsv(exportObj, exportName) {
+  const jsonString = JSON.stringify(exportObj.series);
+  const jsonObj = [JSON.parse(jsonString)];
```
jessex: Why do we need to wrap this in an array?

src/components/charts/revocations/AdmissionCountsByType.js
```diff
@@ -67,6 +67,7 @@ const AdmissionTypeProportions = (props) => {
       id={chartId}
       data={{
         datasets: [{
+          label: 'Count',
```
jessex: Similar to changes elsewhere, should this read "Admission count"?

src/components/charts/revocations/RevocationCountByOfficer.js
```diff
@@ -83,47 +101,42 @@ const RevocationCountByOfficer = (props) => {
     setUnknownDataPoints(officerViolationCountsByType.UNKNOWN_VIOLATION_TYPE);
   }
 
-  function configureDownloads(chart, chartData, visibleOffice) {
+  function configureDownloads(
```
jessex: Nice cleanup in this file :+1:

src/assets/scripts/utils/downloads.js
jjdressel11: I found that converting to JSON and then to csv was the easiest given the nested structure that the data is often in - happy to explore other solutions if you think this is bad practice.

src/assets/scripts/utils/downloads.js
jessex: Got it, thanks!

src/assets/scripts/utils/downloads.js
jjdressel11: HA! This is completely unnecessary! `jsonexport` can work directly on the `exportObj.series` object. Fab.

src/assets/scripts/utils/downloads.js
jjdressel11: See my comment down below - turns out `jsonexport` can work directly with javascript objects. Seems like this library is poorly named. Thoughts on re-naming this import in the header so that it's less confusing that we're using a JSON export library to convert to csv?

jessex: I think that's a great idea.

PR #132, opened 2019-11-20
Multi-tenancy

### LINKED ISSUE #131: Enable multi-tenancy
Enable multi-tenancy for the dashboard, so that we can support users from different agencies and organizations, with different data and visualizations for those sets of users.

The main requirements here are:
* Users from North Dakota, Missouri, and future states each use the same deployed application
* Users from each state can register for an account, and no one else can
* This is already done, but new states need to be added to the pre-registration-hook domain whitelist in Auth0
* Users can see only their own data and no other states’ data
* We can configure different sets of visualizations and pages for different states

We could make the dashboard single-tenant by hosting different instances for each state, but this is a significant operational complexity. We could make the dashboard multi-tenant in the sense that there is a separate subdomain for each state or agency, e.g. north-dakota.dashboard.recidiviz.org, but our preferred user experience is for the dashboard to feel dedicated to the user.

**Additional context**
Check out the mini design doc for more context.

jessex:
## Description of the change

This adds multi-tenancy to the dashboard, i.e. makes it so that we can support users from different agencies and organizations, with different data and visualizations for those sets of users. This includes the following changes:
* Adds `state_code` as a path parameter to each backend API endpoint and propagates it throughout the backend
* Updates the metric refresh and cache code to ensure it caches metrics for different states separately
* Adds two new rules to Auth0 specifying the state code for a user in their `app_metadata` and making that metadata available in the id token, i.e. the user object on the client side itself (these are not yet in version control)
* Adds utilities to the app for working with a user's state code, determining which views are available for that state code, and authorizing a user to look at a specific view depending on their state code
* Adds a new component, `PrivateTenantRoute`, which uses the above utilities to authorize users for state-specific views and forward them to the correct routes
* Adds the ability for Recidiviz users to select which state they want to view through a dropdown menu on the _Profile_ page
* Creates a stub Missouri Snapshots view for testing

## Type of change

## Related issues

Closes #131 

### Development

- [x] Lint rules pass locally

### Code review

### DIFF
--- a/server/core/metricsRefresh.js
+++ b/server/core/metricsRefresh.js
@@ -31,7 +31,10 @@ const demoMode = require('../utils/demoMode');
 const isDemoMode = demoMode.isDemoMode();
 
 const METRIC_REFRESH_INTERVAL_MS = 1000 * 60 * 30; // Refresh metrics every 30 minutes
-const CACHED_STATE_CODES = ['US_ND'];
+const CACHED_STATE_CODES = [
+  'US_MO',
+  'US_ND',
+];
 
 /**
  * Performs a refresh of the free through recovery metrics cache, logging success or failure.
--- a/server/core/metricsRefresh.js
+++ b/server/core/metricsRefresh.js
@@ -31,18 +31,21 @@ const demoMode = require('../utils/demoMode');
 const isDemoMode = demoMode.isDemoMode();
 
 const METRIC_REFRESH_INTERVAL_MS = 1000 * 60 * 30; // Refresh metrics every 30 minutes
+const CACHED_STATE_CODES = ['US_ND']
 
 /**
  * Performs a refresh of the free through recovery metrics cache, logging success or failure.
  */
 function refreshFreeThroughRecoveryMetrics() {
-  metricsApi.fetchFreeThroughRecoveryMetrics(isDemoMode, (err, data) => {
-    if (err) {
-      console.log(`Encountered error during scheduled fetch-and-cache
-        of Free Through Recovery metrics: ${err}`);
-    } else {
-      console.log('Executed scheduled fetch-and-cache of Free Through Recovery metrics');
-    }
+  CACHED_STATE_CODES.forEach((stateCode) => {
+    metricsApi.fetchFreeThroughRecoveryMetrics(isDemoMode, stateCode, (err, data) => {
+      if (err) {
+        console.log(`Encountered error during scheduled fetch-and-cache
+          of Free Through Recovery metrics for ${stateCode}: ${err}`);
+      } else {
+        console.log(`Executed scheduled fetch-and-cache of Free Through Recovery metrics for ${stateCode}`);
+      }
+    });
   });
 }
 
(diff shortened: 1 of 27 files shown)


server/core/metricsRefresh.js
```diff
@@ -31,59 +31,69 @@ const demoMode = require('../utils/demoMode');
 const isDemoMode = demoMode.isDemoMode();
 
 const METRIC_REFRESH_INTERVAL_MS = 1000 * 60 * 30; // Refresh metrics every 30 minutes
+const CACHED_STATE_CODES = ['US_ND'];
```
jjdressel11: Why don't we want 'US_MO' in here?

src/views/stateViews.js
```diff
@@ -0,0 +1,92 @@
...
+  const views = STATE_VIEW_COMPONENTS[stateCode.toLowerCase()];
+  if (!views) {
+    return [];
+  }
+  return Object.keys(views);
+}
+
+const CURRENT_STATE_IN_SESSION = 'recidivizUserCurrentStateInSession';
+
+function getCurrentState() {
```
jjdressel11: Can you either rename or add a docstring that explains that this is only called when the user's state code is _definitely_ 'recidiviz' ? Otherwise it looks a little concerning that we're assigning random states to users.

src/components/StateSelector.js
```diff
@@ -0,0 +1,53 @@
...
+import React, { useState } from 'react';
+import Select from 'react-select';
+
+import { getStateNameForCode } from '../utils/authentication/user';
+import { getAvailableStates, getCurrentState, setCurrentState } from '../views/stateViews';
+
+const StateSelector = () => {
+  // Prepare state selector for Recidiviz users
+  const availableStateCodes = getAvailableStates();
+  const availableStates = availableStateCodes.map(
```
jjdressel11: Alphabetize the states in the selector.

server/core/metricsRefresh.js
jessex: Ah, mistake on my part. Will add!

server/core/metricsRefresh.js
```diff
@@ -31,59 +31,69 @@ const demoMode = require('../utils/demoMode');
...
-  metricsApi.fetchFreeThroughRecoveryMetrics(isDemoMode, (err, data) => {
-    if (err) {
-      console.log(`Encountered error during scheduled fetch-and-cache
-        of Free Through Recovery metrics: ${err}`);
-    } else {
-      console.log('Executed scheduled fetch-and-cache of Free Through Recovery metrics');
-    }
+  CACHED_STATE_CODES.forEach((stateCode) => {
+    metricsApi.fetchFreeThroughRecoveryMetrics(isDemoMode, stateCode, (err, data) => {
+      if (err) {
```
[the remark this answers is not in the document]
jessex: I think there's going to be a broader generalization of the Program Evaluation API, e.g. making `freeThroughRecovery` into a `:programName` path parameter and then properly handling cases where a given program doesn't exist for a given state. That's future work, but I think it's the right approach.

src/views/stateViews.js
jessex: Yes good call. Will add.

src/components/StateSelector.js
jessex: Done!

server/core/metricsRefresh.js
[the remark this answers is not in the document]
jjdressel11: Yes I agree with that approach! Sounds good

Issue #133, opened 2019-12-16
Revocation analysis matrix view


jessex:
**Is your feature request related to a problem? Please describe.**
Develop the revocation analysis matrix view as described in the Revocation Analysis PRD. This includes the interactive matrix, the static charts that are filterable through the matrix, and the case list.

**Describe the solution you'd like**
The solution as described in the PRD.

**Describe alternatives you've considered**
See the PRD for context.

**Additional context**
The calculations will be developed separately. Until they are available, we should test against local data fixtures.

PR #136, opened 2020-01-06
New revocations dashboard

aelnaiem:
## Description of the change

Adding react components for the new Revocations dashboard. I've added this as the revocations page for Missouri, along with some server changes to get all the fixture data needed. The styles in general might need to be generalized and moved out to other spec stylesheets.

I still need to update the revocations over time graph, test it with the new fixture data once that's ready, and update some of the static charts to be organized by risk level.

## Type of change

## Related issues

Takes care of most of #133 

### Development

- [x] Lint rules pass locally

### Code review


server.js
```diff
@@ -79,6 +79,7 @@ app.get('/api/:stateCode/programEvaluation/freeThroughRecovery', checkJwt, api.f
 app.get('/api/:stateCode/reincarcerations', checkJwt, api.reincarcerations);
 app.get('/api/:stateCode/revocations', checkJwt, api.revocations);
 app.get('/api/:stateCode/snapshots', checkJwt, api.snapshots);
+app.get('/api/:stateCode/newRevocations', checkJwt, api.newRevocations);
```
aelnaiem: I created a new endpoint but I wasn't sure if you wanted to combine this data with the other revocations endpoint.

src/views/stateViews.js
```diff
@@ -20,11 +20,13 @@ import UsNdReincarcerations from './tenants/us_nd/Reincarcerations';
 import UsNdRevocations from './tenants/us_nd/Revocations';
 import UsNdSnapshots from './tenants/us_nd/Snapshots';
 
+import UsMoRevocations from './tenants/us_mo/Revocations';
 import UsMoSnapshots from './tenants/us_mo/Snapshots';
 
 const STATE_VIEW_COMPONENTS = {
   us_mo: {
     '/snapshots': UsMoSnapshots,
+    '/revocations': UsMoRevocations
```
aelnaiem: I've added this as the revocations page for Missouri, but I'm not sure if that's correct

src/components/charts/new_revocations/RevocationsByRiskLevel.js
```diff
@@ -0,0 +1,95 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+// =============================================================================
+
+import React, { useState, useEffect } from 'react';
+import { Bar } from 'react-chartjs-2';
+
+import { COLORS } from '../../../assets/scripts/constants/colors';
+
+const RISK_LEVEL_TO_LABEL = {
```
aelnaiem: Should this live somewhere else? It can also be used when doing the racial and gender-based static chart grouping

src/components/charts/new_revocations/RevocationsByRace.js
```diff
@@ -0,0 +1,127 @@
...
+// =============================================================================
+
+import React, { useState, useEffect } from 'react';
+import { Bar } from 'react-chartjs-2';
+
+import { COLORS } from '../../../assets/scripts/constants/colors';
+
+const CHART_LABELS = ["Overall", "Low", "Moderate", "High", "Very high"];
+const RISK_LEVELS = ["LOW", "MODERATE", "HIGH", "VERY_HIGH"];
+const RACES = ["WHITE", "BLACK", "HISPANIC", "ASIAN", "NATIVE_AMERICAN", "PACIFIC_ISLANDER"];
```
aelnaiem: Should this be hard-coded? It can also be a mapping of value to label like I'm doing in the RevocationsByRiskLevel component

src/views/tenants/us_mo/Revocations.js
```diff
@@ -0,0 +1,248 @@
...
+import RevocationsByViolation
+  from '../../../components/charts/new_revocations/RevocationsByViolation';
+import RevocationsBySex
+  from '../../../components/charts/new_revocations/RevocationsBySex';
+import RevocationsByRace
+  from '../../../components/charts/new_revocations/RevocationsByRace';
+import CaseTable
+  from '../../../components/charts/new_revocations/CaseTable';
+
+const CHARGE_CATEGORIES = [
```
aelnaiem: I'm using a helpful React component for selecting these filters, and it requires the value and label data in this format. This is another example of where I'm hard-coding these values but it can be provided by the server as well if it's something that changes

src/views/stateViews.js
jessex: That is correct!

src/views/tenants/us_mo/Revocations.js
jessex: Very helpful, @aelnaiem! We'll likely shift this hard-coding out to the API in time, but this is the right approach for now.

yarn.lock
```diff
@@ -8534,6 +8534,7 @@ react-scripts@3.0.1:
 react-select@^3.0.8:
   version "3.0.8"
   resolved "https://registry.yarnpkg.com/react-select/-/react-select-3.0.8.tgz#06ff764e29db843bcec439ef13e196865242e0c1"
+  integrity sha512-v9LpOhckLlRmXN5A6/mGGEft4FMrfaBFTGAnuPHcUgVId7Je42kTq9y0Z+Ye5z8/j0XDT3zUqza8gaRaI1PZIg==
```
jessex: @aelnaiem: do you know what brought in this integrity check, out of curiosity? Did you explicitly add it, did it get added during your yarn install, something else?

server.js
jessex: This is fine as-is for now. Once we merge this in, I'll do some reorganization to move it out from under "new revocations" into something like "revocations/matrix", same with the packaging in the codebase. But you don't need to worry about that!

src/components/charts/new_revocations/RevocationsByRace.js
jessex: I think this is probably fine as-is for now. Although each region tends to have their own system for encoding protected demographics like race, we universalize them on ingest. So we know which enum values will be sent from backend to frontend. But we do need to think about how we display those as labels on the frontend, and whether that requires basically a case-by-case reversal of the what we did in ingest for display purposes. I'll give it more thought but it's a larger consideration than this PR.

src/components/charts/new_revocations/RevocationsByRiskLevel.js
jessex: Yeah if you place this and any helpers in a well-named utility class in `src/utils/transforms/`, I'd appreciate it.

yarn.lock
aelnaiem: It happened through my install, I didn't add it explicitly

src/components/TopBar.js
```diff
@@ -52,18 +55,22 @@ const TopBar = (props) => {
...
-            <a id="sidebar-toggle" className="sidebar-toggle" href="javascript:void(0);">
-              <i className="ti-menu" />
-            </a>
-          </li>
-          )}
-          <li style={{ paddingLeft: '20px', paddingTop: '22px' }}>
-            <h5 className="lh-1 mB-0 logo-text recidiviz-dark-green-text">{normalizedPath}</h5>
-          </li>
-        </ul>
+        {isLanternState(stateCode) ? (
```
jessex: Oh man I don't know why but it never occurred to me to use a ternary operator in JSX. Still learning the stack! :+1:

src/views/stateViews.js
```diff
@@ -36,6 +38,23 @@ const STATE_VIEW_COMPONENTS = {
   },
 };
 
+const LANTERN_STATE_CODES = ['recidiviz', 'us_mo'];
+
+/**
+ * Returns whether the given stateCode is a lantern state
+ */
+function isLanternState(stateCode) {
+  return LANTERN_STATE_CODES.includes(stateCode);
```
jessex: Is this actually the appropriate logic? This looks like it would apply for any state when a Recidiviz user is logged in (anyone with an @recidiviz.org email is going to have Recidiviz as a state code), even when the Recidiviz user is switched to a non-Lantern state. Or whenever someone is operating in demo mode, again even when the user is switched to a non-Lantern state. I think we need to look at which state is actually being viewed in the given moment, regardless of the user's state code, e.g. by checking `stateViews.getCurrentStateForRecidivizUsers` if the current state code is `recidiviz`.

src/assets/styles/spec/components/dashboard.scss
```diff
@@ -22,8 +22,12 @@
   height: 180px;
 }
 
+.matrix-container {
```
jessex: It looks like this new class is causing the matrix to be exactly 890 pixels in width regardless of the actual viewport, which means that on wider screens (including even just my 15" MBP), the "Using this chart" panel to the immediate right of the matrix can become very wide, and the relatively sparse content inside of it leaves a lot of unused vertical space. Is this by design or unexpected?

Here's a screenshot was fake data:
Screen Shot 2020-01-19 at 23 54 17

src/views/stateViews.js
jessex: I just confirmed this. When you go to _Profile_ as a demo user and select another state that is not Lantern, e.g. North Dakota, you still have the Lantern logo in the top and there is no sidebar. So we need to make the update I mentioned above.

jessex:
@aelnaiem one thing about the UX of removing the sidebar for Lantern states, there's no longer a link to click anywhere for Missouri to get back to the revocations view. I imagine that 1) the Lantern logo needs to be a link back to the homepage, and 2) we should remove the snapshots view for Missouri (at least for now), and 3) we should ensure that when you go back to the homepage for Missouri that it takes you to the revocations view.

src/views/stateViews.js
aelnaiem: Thanks, this should be straight forward if I understand how `getCurrentStateForRecificizUsers` works

src/assets/styles/spec/components/dashboard.scss
aelnaiem: Yea, I didn't actually know which of the two should expand, so I decided to make the using the chart section expand since the chart is a sort of fixed item, but it can instead be the other way. Let me know what the design would actually call for

jessex: This looks great now. Thanks, @aelnaiem!

Issue #140, opened 2020-01-06
Enable global and view-specific toggling for visualizations in the dashboard


jessex:
**Is your feature request related to a problem? Please describe.**
Following from the "Feature Completeness" PRD, implement toggling at the global and view-specific levels for dashboard visualizations. This should include global level toggles for:
* metric type
* time window
* supervision type
* district

Not all toggles are relevant to all charts, and the PRD includes visual treatment for conveying these cases. There is also a decent amount of dynamic visualization changes, e.g. not showing goal lines and trend lines for toggle states that are irrelevant to the specified goal, or not showing the dynamic generated descriptions in such cases.

**Describe the solution you'd like**
The toggling will happen on the client side. Metric json files will be expanded to include rows that have all of the combinations of supported dimensions, which will make them much larger in many cases. The client will iterate over the given rows and use those which are relevant to the current toggle states.

jessex:
This has been completed and released as part of https://github.com/Recidiviz/pulse-dashboard/releases/tag/v1.1.0.

Thanks to everyone for the incredible effort!

PR #155, opened 2020-01-29
Fixes the stickiness behavior for the toggle bar

jessex:
## Description of the change

Fixes the stickiness behavior of the toggle bar, preventing overlap with the `navbar` or with the dropdown menu from the user login. 

## Type of change

## Related issues

Towards #140 

### Development

- [x] Lint rules pass locally

### Code review

### DIFF
--- a/src/components/TopBar.js
+++ b/src/components/TopBar.js
@@ -62,7 +62,7 @@ const TopBar = (props) => {
     return isLanternState(stateCode);
   };
 
-  let navBarClass = 'header wide-navbar';
+  let navBarClass = 'header navbar wide-navbar';
   if (shouldLoadSidebar(isAuthenticated)) {
     navBarClass = 'header navbar';
   }
--- a/src/components/toggles/ToggleBar.js
+++ b/src/components/toggles/ToggleBar.js
@@ -22,8 +22,14 @@ import MetricTypeToggle from './MetricTypeToggle';
 import SupervisionTypeToggle from './SupervisionTypeToggle';
 import MetricPeriodToggle from './MetricPeriodToggle';
 
+const TOGGLE_STYLE = {
+  zIndex: 700,
+  position: 'sticky',
+  top: 55,
+};
+
 const ToggleBar = (props) => (
-  <div className="row gap-20">
+  <div className="row gap-20 pB-10" style={TOGGLE_STYLE}>
     {/* Figure out how to make sticky top work without hiding the dropdown */}
     <div className="col-md-12">
       <div className="bd bgc-white p-20">


src/components/TopBar.js
```diff
@@ -62,7 +62,7 @@ const TopBar = (props) => {
     return isLanternState(stateCode);
   };
 
-  let navBarClass = 'header wide-navbar';
+  let navBarClass = 'header navbar wide-navbar';
```
jessex: This was necessary because of an issue that was introduced with the login dropdown menu's behavior. The expected behavior is that you can click the top of the menu to make the dropdown appear, and click again to make it disappear. However, that depends on a jQuery callback that relies on the top bar being set with a class of `navbar`. With `navbar` removed, you could click to bring the menu into focus, but it would never disappear unless you refreshed or left the page.

src/components/toggles/ToggleBar.js
```diff
@@ -22,8 +22,14 @@ import MetricTypeToggle from './MetricTypeToggle';
 import SupervisionTypeToggle from './SupervisionTypeToggle';
 import MetricPeriodToggle from './MetricPeriodToggle';
 
+const TOGGLE_STYLE = {
+  zIndex: 700,
```
jessex: The `z-index` of the top bar is 800, and that of the login dropdown menu is 1000, and that of the various chart components on the page is 100, I think. 700 gets this such that chart components scroll underneath and past it, but the dropdown menu still appears over the top, and we don't clip the navbar by a few pixels in different display resolutions.

src/components/toggles/ToggleBar.js
```diff
@@ -22,8 +22,14 @@ import MetricTypeToggle from './MetricTypeToggle';
...
+  position: 'sticky',
+  top: 55,
```
jessex: This is the height of the top bar, so this ensures the toggle bar stays just below the top bar.

src/components/toggles/ToggleBar.js
```diff
@@ -22,8 +22,14 @@ import MetricTypeToggle from './MetricTypeToggle';
...
+};
+
 const ToggleBar = (props) => (
-  <div className="row gap-20">
+  <div className="row gap-20 pB-10" style={TOGGLE_STYLE}>
```
jessex: Switching from `position: relative` (the inherited default) to `position: sticky` causes some of the margin and padding properties to go a bit out of whack. Adding an extra padding along the bottom of 10 pixels ensures the first row of charts doesn't get too close. Keeps the grid even overall.

PR #185, opened 2020-02-18
Positive negative scales choropleth

### LINKED ISSUE #170: Add color to distinguish negative and positive numbers in map mode for admissions vs. releases
The `ADMISSIONS VERSUS RELEASES` chart in "map" view does not distinguish between positive and negative numbers. This is confusing to those who want to know which counties have more people being released (negative) and which have more people being admitted (positive)

Bubbles with positive numbers are in red scale, bubbles with negative numbers are in the current blue scale. The bubble size still conveys the magnitude of the number. 

- Transforming the bubble color & size minimum to be relative to most positive number, so the smallest / lightest bubble is the biggest positive number, biggest / darkest bubble is biggest negative number. This option is confusing to user as to why negative gets to be dark and positive is the smallest. This would also be true for a reverse scale form most negative to most positive
- Changing shape or adding shading to indicate positive / negative. This is not as obvious to users at a glance for which elements mean what, and adds an additional "dimensionality" from user perspective to understand. 

**Additional context**
I don't think we currently have a red scale. We would need to develop and, ideally, validate one could distinguish the red scale and blue scale if color blind (though those who are color blind likely wouldn't be worse off with differing pos/neg colors than they are now where the color is the same).

jessex:
## Description of the change

Configures choropleth charts to use different color ranges for positive and negative scales in charts that could feature both. Choropleths are already configured to use blue colors by default. Now if both positive and negative can occur, negatives will get the blue colors and positives will get the red colors. There could certainly be room for more general configuration/functionality here, but we'll omit that complexity until it's necessary.

Configures the Admissions vs Releases geo view to use this functionality.

## Type of change

## Related issues

Closes #170 

### Development

- [x] Lint rules pass locally

### Code review

### DIFF
--- a/src/utils/charts/choropleth.js
+++ b/src/utils/charts/choropleth.js
@@ -18,18 +18,49 @@ import { scaleLinear } from 'd3-scale';
 import { toHumanReadable } from '../transforms/labels';
 import { COLORS } from '../../assets/scripts/constants/colors';
 
-function colorForValue(value, maxValue, useDark) {
-  const scale = scaleLinear()
+const BLUES = ['#F5F6F7', '#9FB1E3', COLORS['blue-standard-2']];
+const DARK_BLUES = ['#CCD1DE', '#8897C4', '#1F2A3B'];
+const REDS = ['#F7F5F6', '#E39FB1', COLORS['red-standard']];
+const DARK_REDS = ['#DECCD1', '#C48897', '#3B1F2A'];
+
+function colorForValue(value, maxValue, useDark, possibleNegative) {
+  const scaleValue = possibleNegative ? Math.abs(value) : value;
+  const valueWasNegative = value < 0;
+
+  // For charts without negative values, default to using blues. If negative values can occur,
+  // use reds for positive and blues for negative, in line with convention elsewhere in the UI.
+  const positiveColors = possibleNegative ? REDS : BLUES;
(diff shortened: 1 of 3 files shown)


src/utils/charts/choropleth.js
```diff
@@ -18,18 +18,49 @@ import { scaleLinear } from 'd3-scale';
...
+const DARK_BLUES = ['#CCD1DE', '#8897C4', '#1F2A3B'];
+const REDS = ['#F7F5F6', '#E39FB1', COLORS['red-standard']];
+const DARK_REDS = ['#DECCD1', '#C48897', '#3B1F2A'];
+
+function colorForValue(value, maxValue, useDark, possibleNegative) {
+  const scaleValue = possibleNegative ? Math.abs(value) : value;
+  const valueWasNegative = value < 0;
+
+  // For charts without negative values, default to using blues. If negative values can occur,
+  // use reds for positive and blues for negative, in line with convention elsewhere in the UI.
```
jjdressel11: The convention in the UI is to use blue for good and red for bad, which currently happens to coincide with blue as negative and red for positive. However, this won't always be the case, so I think you should send in some value when `possibleNegative` is present about whether negative is good or bad and act accordingly.

This can be filed as an issue and done later if you want.

jessex: Yeah I had that exact same thought and decided to defer until we had a use case that called for more flexibility here. We'll leave it for a rainy, multi-colored day.

PR #202, opened 2020-02-27
Lantern ui launch blockers

### LINKED ISSUE #190: Manage the size of the charts in the revocation analysis matrix view
**Describe the bug**
The charts "below the fold" in the revocation analysis matrix view are quite large, and on some devices they cannot be fully viewed without scrolling. That's a fairly significant reduction in user experience, so we need to make these charts more manageable. This could mean 1) making them generally smaller, and/or 2) making them generally responsive (at present, they will expand if you expand the window, but not shrink if you shrink the window). 

**To Reproduce**
1. Go to /revocations when logged into a Lantern state
2. Scroll down to the charts below the matrix itself
3. See that they are generally quite large and some amount of scrolling is required to fully operate them

**Expected behavior**
Each chart below the fold should fit fully and neatly within the viewport provided by most users' devices. No scrolling should be required to, for example, see both the top of the y-axis and the labels on the x-axis.

 - OS [e.g. Windows 10]: All?
 - Browser [e.g. chrome, edge]: All?
 - Version [e.g. 22]: All?

### LINKED ISSUE #191: Improve the color scheme in the revocation analysis matrix view
**Describe the bug**
The Revocations by Race chart below the matrix in the revocation analysis matrix view has a color scheme which makes it difficult to understand the chart at a glance. There was a miscommunication between when the original medium-fi mocks were created and engineering took over, and the drop-in color scheme was never fully fleshed out.

@jjdressel11 has a proposed color scheme that still seems to fit within the Lantern branding, but would need to be broadly applied across the view, not just to this single chart. This seems wise, however, given that there are accessibility issues with the coloring elsewhere in the view anyway.

**To Reproduce**
1. Go to /revocations for a Lantern state
2. Observe accessibility issues related to color contrast in a variety of charts, including the matrix itself
3. Scroll down below the matrix and select "Race" to see Revocations by Race
4. Observe that the bars for different races and ethnicities in the chart are difficult to distinguish

**Expected behavior**
1. Accessibility scores for the view are passing
2. It is easy to differentiate between different races in the Revocations by Race chart, e.g. a user who has seen the view before can easily tell the race of a given bar by color alone

 - OS [e.g. Windows 10]: All
 - Browser [e.g. chrome, edge]: All
 - Version [e.g. 22]: All

### LINKED ISSUE #192: Provide a visual treatment for current month in progress for month-over-month charts
For charts which count data points month-over-month, the rightmost data point, for the current month still in progress, might visually trick viewers into invalid conclusions.

For example, if we are very early in a month, it may look like there is a general downward trend, even if the viewer is aware of the current date. If we are mid-to-late in a month, and the data point is roughly in line with the full results of previous recent months, it may lead the viewer into prematurely thinking that this month will wind up holding a somewhat constant trend, even though more data in the coming days may cause this month to be abnormally high.

These are of course not necessarily permanent issues as when the month is over the data point will generally stop moving, but it would be a nice usability improvement to have a visual reminder that the rightmost data point is still in progress and changing day-over-day. The more pressing need for this is for the revocation analysis matrix view's month-over-month chart, but the same solution could be applied to charts in other views, as well, and this was originally discussed in the PRD for #140.

Some sort of visual treatment on the chart itself, e.g. a text change or a color change, indicates that the current month is still in progress, at a glance. For example, we could simply append the words (in progress) to the current month label in the x-axis, and/or we could change the color of that label to a bright eye-catching color. We are limited in terms of what we can do with the actual chart visuals (bars, lines, etc.) by Chartjs, so we need to choose something practical.

An alternative, for some or all charts, is to not display the current month in progress. However, this has the significant downside of making it feel as though the chart only updates monthly, which may condition viewers to only look at the data once a month, not relying on it as a more fresh and constant tool for decision-making. That could have real outcome ramifications.

jessex:
## Description of the change

Addresses three launch blockers for the revocation analysis matrix tool:
- Makes the color scheme significantly more accessible
- Makes the static charts below the matrix more easily viewable by holding their size as a relative proportion of the viewport
- Adds a new visual treatment for the current month in progress for month-over-month charts

This is all still pending design review.

## Type of change

## Related issues

Closes #190 
Closes #191 
Closes #192 

### Development

- [x] Lint rules pass locally

### Code review

### DIFF
--- a/src/assets/scripts/constants/colors.js
+++ b/src/assets/scripts/constants/colors.js
@@ -277,6 +277,11 @@ const COLORS = {
   'blue-standard': '#809AE5',
   'blue-standard-2': '#3F4D62',
   'blue-standard-light': 'rgba(63, 77, 98, .7)',
+  'lantern-orange': '#F07132',
+  'lantern-light-blue': '#03A9F4',
+  'lantern-eggplant': '#5C384D',
+  'lantern-burnt-orange': '#8B2D21',
+  'lantern-blue': '#182B5E',
 };
 
 const GREYS = {
@@ -337,6 +342,15 @@ const COLORS_FIVE_VALUES_LIGHT = [
   '#fdb0a8',
 ];
 
+const COLORS_LANTERN_SET = [
+  '#F07132',
+  '#03A9F4',
+  '#5C384D',
+  '#8B2D21',
+  '#182B5E',
+  '#0E2C16',
+];
+
 export {
   COLORS,
   GREYS,
@@ -347,4 +361,5 @@ export {
   COLORS_THREE_VALUES,
   COLORS_FIVE_VALUES,
   COLORS_FIVE_VALUES_LIGHT,
+  COLORS_LANTERN_SET,
 };
--- a/src/assets/styles/spec/components/dashboard.scss
+++ b/src/assets/styles/spec/components/dashboard.scss
@@ -70,8 +70,8 @@
           background: white;
           border-radius: 25px;
           padding: 0 12px;
-          border: 1px solid rgb(240, 113, 50);
-          color: rgb(240, 113, 50);
+          border: 1px solid rgb(92, 56, 77);
+          color: rgb(92, 56, 77);
           line-height: 24px;
           font-size: 12px;
           font-weight: 600;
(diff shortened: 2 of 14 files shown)


jessex:
This is replaced by #203.

PR #208, opened 2020-03-04
Add "Important Notes" bar to all ND pages

### LINKED ISSUE #186: Add "things to know" bar to ND dashboard
ND currently uses different methodologies for counting revocations than they've asked us to use. It is non-obvious to users of the dashboard that we are use person-based methodologies and that we count revocations based on prison admissions, despite our explanations in the methodologies tabs. Not understanding the way we do calculations makes our numbers seem "wrong" in comparison with the current reports, when in fact we simply are calculating different numbers. 

This feature is intended to make key methodological choices (namely with person based counting and revocations based on admissions to prison) exceedingly obvious to ND users. 

A bar under the "explore bar" labeled "important" notes that expands when clicked into an explanation of important methodological choices. This bar is on all pages of the dashboard. 

The current methodology section contains this information. We could expand these sections, train users to use them more, or emphasize them more, however, it's nice to have the most critical information available at first glance. 

We could change page and chart titles to imply the differences in our methodologies. We may end up doing this, but overall this makes the dashboard feel less inviting and more complicated. 

We could change our methodology to be in line with the intuition of line staff in ND, but lose the opportunity to shift the state to new methodology that is more in line with state priorities and values. 

**Additional context**
May look into getting external front end work to add this, as it seems like a small front end ask well contained within itself.

itsjon:
## Description of the change

Adds an "Important Notes" bar to every ND page. This looks like a much larger change than it is because it also extracts a page template to reduce repeated code.

@kbinney feel free to patch this into your local and let me know if the spacing or anything else needs to be adjusted.

## Type of change

## Related issues

Closes #186

### Development

- [x] Lint rules pass locally

### Code review


jessex:
Hey @itsjon, small notekeeping point: Github's autoclose feature won't work if the "Closes #XXX" statement is in a blockquote, as it is in the PR template. You need to change it when issuing the PR to get the desired effect (slightly annoying, I know). I updated it as needed so you can see for future PRs.

src/views/tenants/us_nd/PageTemplate.js
```diff
@@ -0,0 +1,88 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+// =============================================================================
+
+import React from 'react';
+import '../../../assets/styles/index.scss';
+
+
+const ImportantNotesBar = () => (
```
jessex: Thanks so much for the extraction/decomposition! While we're at it, would you mind placing `ImportantNotesBar` in its own file?

src/views/tenants/us_nd/PageTemplate.js
```diff
@@ -0,0 +1,88 @@
...
+          </h5>
+        </button>
+        <div id="importantNotes" className="collapse" aria-labelledby="importantNotesBarHeading"
+             data-parent="#importantNotesBar">
+          <div className="bdT mT-10 pT-10">
+            <p>
+              <span className="font-weight-bold">PERSON-BASED COUNTING</span>: In collaboration with Central Office,
+              Recidiviz has decided to emphasize the counting of people rather than cases. Unless noted otherwise, all
+              our data focused on the number of people who successfully complete supervision, or the number of people
+              admitted to prison because of a revocation. If an individual has multiple cases terminated differently (ie
```
jessex: Typo nit: `ie` --> `e.g.`

src/views/tenants/us_nd/PageTemplate.js
```diff
@@ -0,0 +1,88 @@
...
+              least one case is relevant (in that case, the individual would count towards successful terminations
+              because of the expired case, and towards revocations because of the revoked case).
+            </p>
+            <p>
+              <span className="font-weight-bold">REVOCATIONS TO DOCR FACILITY</span>: In conjunction with North Dakota
+              priorities, Recidiviz looks specifically at revocations to DOCR facilities. Revocations resulting in other
+              dispositions are not counted. In addition, revocations are counted only when an individual’s admittance to
+              a facility is documented in Elite as a revocation. Individuals whose revocations result in continuation or
+              a sentence to a county jail are not included in revocation charts, unless noted otherwise for the chart.
+              Individuals sentenced to a DOCR facility but who do not wind up in that facility are not counted (this may
```
jessex: Subjective language nit. I would probably say something like "Individuals sentenced to a DOCR facility but who are not ultimately incarcerated in a DOCR facility are not counted..."

src/views/tenants/us_nd/FreeThroughRecovery.js
```diff
@@ -66,449 +67,447 @@ const FreeThroughRecovery = () => {
...
-                    />
-                  )}
-                </div>
-                <div className="layer bdT p-20 w-100 accordion" id="methodologyFtrReferralCountByMonth">
-                  <div className="mb-0" id="methodologyHeadingFtrReferralCountByMonth">
-                    <div className="mb-0">
-                      <button className="btn btn-link collapsed pL-0" type="button" data-toggle="collapse" data-target="#collapseMethodologyFtrReferralCountByMonth" aria-expanded="true" aria-controls="collapseMethodologyFtrReferralCountByMonth">
-                        <h6 className="lh-1 c-blue-500 mb-0">Methodology</h6>
-                      </button>
                     </div>
```
jessex: Looks like the story here is that I was missing a `</div>`. Is that right?

itsjon:
> Great work, @itsjon! I didn't test this locally because I'm short on time today but the code and the decomposition looks great (looking forward to much more decomposition). I'll let @kbinney take a look at styling and copy and comment there.

@jessex oops, thank you for fixing!

itsjon:
D'oh, quoted the wrong comment in my reply. Thank you for fixing the GitHub autoclose component of the PR description, @jessex.

Checking out the other comments now.

src/views/tenants/us_nd/PageTemplate.js
itsjon: done

src/views/tenants/us_nd/PageTemplate.js
itsjon: fixed

src/views/tenants/us_nd/PageTemplate.js
itsjon: That sounds better to me as well, and I've updated it accordingly. @kbinney let me know if there are any copy changes you'd like made before this gets merged.

src/views/tenants/us_nd/FreeThroughRecovery.js
itsjon: I don't think so, but the HTML is definitely valid now. GitHub's diff visualization couldn't make much sense of this change, but the HTML inside `<React.Fragment>` should be unchanged from before aside from indentation.

itsjon:
@jessex thank you for the review! Excited for more decomposition as well 😀

src/views/tenants/us_nd/PageTemplate.js
jessex: It doesn't look like this has been done yet, unless I am sorely mistaken?

jessex:
@itsjon: I don't think you actually pushed up any new changes prior to re-requesting my review. Are you sure you pushed any more commits?

src/views/tenants/us_nd/FreeThroughRecovery.js
jessex: Yeah when you select the setting to "Hide whitespace changes" on the PR, it becomes _significantly_ smaller and it doesn't actually appear that any closing element was missing. It's just a weird diff.

itsjon:
:sigh: optimistic command use there. _Now_ it's pushed!

kbinney:
This works great and is exactly what I specified, but I unfortunately have some last minute change requests.

I talked about the content with Andrew and have some changes to the copy, including copy specific to each page (ie snapshots copy vs. revocation copy). Expect them by the end of the day. I also did some quick user research amongst my co-workers and found the bar appears less "clickable" than I had anticipated. Can you add in an arrow similar to the below? Co-worker UXR suggests it makes a difference. 

When bar is closed: (ignore the blue border around the important facts box)
Screen Shot 2020-03-09 at 2 52 07 PM

After clicking it (more text showing): (ignore the orange line, it's from prototyping in sketch)
Screen Shot 2020-03-09 at 2 52 14 PM

Sorry for the last minute change requests!

itsjon:
@kbinney all good! I'll take a look at adding that arrow and can update the copy when you have it.

itsjon:
@kbinney I used a couple icons from the icons library already in use in the app. Let me know what you think!

Collapsed:

Screen Shot 2020-03-09 at 5 47 14 PM

Expanded:
Screen Shot 2020-03-09 at 5 47 22 PM

PR #225, opened 2020-03-11
Update default value of filters at the top of the page

### LINKED ISSUE #210: Update default filter menu strings to the currently selected option
The ‘District’, ‘Supervision Level’, and ‘Supervision Type’ filters at the top of the page all use the default string ‘Select…’, when the option ‘All’ is actually already selected by default. 

We should just have them defaulted to ‘All’, since that makes the current selection clearer and is more consistent.

We could leave them as-is, but this would be inconsistent with the 'Time Period' filter, which shows the currently selected option rather than 'Select...'. The 'Time Period' filter is more transparent behavior, however.

**Additional context**
Screenshot attached.
image

NikiO-INO:
## Description of the change

Update default selects value as being discussed in task. Fixed compatibility with IE 11.

## Type of change

## Related issues

Closes #210

### Development

- [ ] Lint rules pass locally

### Code review


package.json
```diff
@@ -90,7 +90,8 @@
     "development": [
       "last 1 chrome version",
       "last 1 firefox version",
-      "last 1 safari version"
+      "last 1 safari version",
+      "ie 11"
```
jessex: Was there a specific issue that this fixed?

src/views/tenants/us_mo/Revocations.js
```diff
@@ -257,20 +257,23 @@ const Revocations = () => {
             <Select
               options={districts}
               onChange={(option) => updateFilters({ district: option.value })}
+              defaultValue={{ value: 'All', label: 'All districts' }}
```
jessex: Since we reference this in two separate places, let's extract it out to a constant near the top of the file:
```
const DEFAULT_DISTRICT = { value: 'All', label: 'All districts' };
```

package.json
NikiO-INO: We couldn't open dashboard localy on IE 11, with such errors on image. This solution remove all errors.
image (1)

src/views/tenants/us_mo/Revocations.js
NikiO-INO: Yep, I fixed that, but we already have "DEFAULT_DISTRICT" const, I named it like "DEFAULT_BASE_DISTRICT"

package.json
jessex: Oh interesting. Because I use a Mac I have never tried to open it _locally_ in IE11. But I have kept the app compatible with IE11 in staging and production. Thanks for taking care of this!

PR #240, opened 2020-03-18
Extract re-usable code from the Revocations by Officer chart and use it to add a new 'Case Terminations by Officer' chart

### LINKED ISSUE #188: Add front end support for "Case terminations by officer" chart
ND currently tracks officer revocation rates by looking at case terminations that were revocations  / all case terminations in a given month. Our dashboard moves to a people based methodology using Elite data as the source of truth for revocations. As a result, our `by officer` chart appears to "miss" many revocations POs & their supervisors would have counted in officers' revocation rates. 

We don't want to shift our methodology to exactly mirror existing ND methodology, as ND priorities and values seem to have shifted since their report generating tools (and the underlying methodology) were created. However, we do want POs and their supervisors to trust the numbers they see in the dashboard. One way to do that is to provide more numbers calculated in ways similar to their existing numbers. 

Add front end support for a "Case terminations by officer" chart. This chart should look like the existing `revocations by officer` chart. However, it will look at case terminations by officer, color-divided by type of termination (death, discharge, expiration, suspension, revocation, absconsion, unknown), instead of revocations by office. 

It should support all toggles. Selecting "rate" mode will show for each officer the percent of their terminations that were of each type. The percent of terminations that revocations should be close to what they currently calculate as their revocation rate. It should not support a map mode.  

This chart should be placed under `revocations by violation type` next to the new `case terminations by month` chart. 

- Changing the methodology underlying the revocations by officer chart. Undesirable because the methodology used in that chart was developed in collaboration with North Dakota and preferred in the long run if we can get users to trust the numbers in that chart. 

**Additional context**
- The `revocations by officer` chart does not show officer numbers when all districts are selected, as it is too many #s on the x axis. This chart should have the same behavior as that chart in that officer IDs are only shown when a particular district is selected, and the alert will pop up to explain that when all districts are selected. 

We are also looking into an easier way for ND staff to map Officer IDs to Officer names (https://github.com/Recidiviz/pulse-dashboard/issues/176). The solution decided on for `revocations by officer` should apply to this chart as well.

itsjon:
## Description of the change

Similarly to https://github.com/Recidiviz/pulse-dashboard/pull/231, this PR adds snapshot tests for `RevocationCountByOfficer.js` (this time zipping up the snapshots so they don't dominate the diff), then extracts the chart creation code from that component and makes it reusable. That reusable code is used in a new 'Case Terminations by Officer' chart.

I'd originally had in mind to expand and reuse the existing generic bar chart function in `BarCharts.js`, but I deprioritized that for now in the interest of time. I've created https://github.com/Recidiviz/pulse-dashboard/issues/241 to track that improvement.

## Type of change

## Related issues

Closes #188

### Development

- [x] Lint rules pass locally

### Code review


itsjon:
Cc @kbinney

src/components/charts/revocations/test_data/.gitignore
```diff
@@ -0,0 +1 @@
+RevocationCountByOfficer/snapshots
```
jessex: Why are we ignoring this? Don't snapshots need to be committed to ensure consistent testing across developer machines?

src/components/charts/revocations/test_data/snapshotChartVariants.js
```diff
@@ -21,28 +21,42 @@
...
-import * as revocationCountByViolationType from '../RevocationCountByViolationType';
+// import * as revocationCountBySupervisionType from '../RevocationCountBySupervisionType';
+// import * as revocationCountByViolationType from '../RevocationCountByViolationType';
+import * as revocationCountByOfficer from '../RevocationCountByOfficer';
 
 const fs = require('fs');
 const path = require('path');
 
 const charts = [
+  // {
```
jessex: Are we planning to uncomment this before merging?

src/components/charts/revocations/test_data/snapshotChartVariants.js
```diff
@@ -21,28 +21,42 @@
 
 import { readJsonLinesFile } from '../../../../utils/testing';
 
-import * as revocationCountBySupervisionType from '../RevocationCountBySupervisionType';
-import * as revocationCountByViolationType from '../RevocationCountByViolationType';
+// import * as revocationCountBySupervisionType from '../RevocationCountBySupervisionType';
```
jessex: Same here: uncomment before merging?

src/components/charts/revocations/test_data/.gitignore
jessex: Ah, I see. Is this because we are committing the `snapshots.zip` file directly?

itsjon: Correct! I just didn't want another bunch of files cluttering the PR.

src/components/charts/revocations/test_data/snapshotChartVariants.js
itsjon: No, I just kept those in the code (commented-out) for the record, but I'm also happy to delete them.

Actually, now that I'm thinking on it more, I will go ahead and delete it. We shouldn't need them again, and there's always git history.

src/components/charts/revocations/test_data/snapshotChartVariants.js
itsjon: Also deleted these commented-out lines

jessex: Why are they not necessary, though? Because it's a tool for a one-time generation of many snapshots for a given chart?

itsjon: Right, the primary purpose of the snapshots is to make sure nothing's broken during code extraction/deduplication, so I don't expect to re-run the snapshotting process for any one chart. But I do want the snapshotting code checked in so we have a record of how the snapshots were generated.

itsjon:
@jessex sounds good. Thank you for the quick review!

kbinney:
Looks great!

PR #254, opened 2020-03-23
Handle DUAL and ALL supervision type

jjdressel11:
## Description of the change

Adds "Dual Supervision" to the Supervision Type dropdown menu, which supports `supervision_type=DUAL` in the data. Also now accepts `supervision_type=ALL` as the default value when the dropdown is not set to filter to a specific supervision type. This change means that we are no longer doing sums on the front-end across the various supervision types.

This fixes an issue where the revocation counts were around 2-3x too high across the board for views that didn't filter by supervision type, because people on DUAL supervision were being counted in the PAROLE, PROBATION, and DUAL counts, which were all being summed to produce the aggregate counts.

## Type of change

### Development

- [X] Lint rules pass locally

### Code review

### DIFF
--- a/src/views/tenants/us_mo/Revocations.js
+++ b/src/views/tenants/us_mo/Revocations.js
@@ -70,7 +70,6 @@ const SUPERVISION_TYPES = [
   { value: 'All', label: 'All' },
   { value: 'PROBATION', label: 'Probation' },
   { value: 'PAROLE', label: 'Parole' },
-  // TODO: Enable dual supervision filtering when supported in calculation methodology
   { value: 'DUAL', label: 'Dual Supervision' },
 ];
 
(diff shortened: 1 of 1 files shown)


src/views/tenants/us_mo/Revocations.js
```diff
@@ -67,14 +67,15 @@ const CHARGE_CATEGORIES = [
 const DEFAULT_CHARGE_CATEGORY = 'All';
 
 const SUPERVISION_TYPES = [
-  { value: '', label: 'All' },
+  { value: 'All', label: 'All' },
   { value: 'PROBATION', label: 'Probation' },
   { value: 'PAROLE', label: 'Parole' },
   // TODO: Enable dual supervision filtering when supported in calculation methodology
```
jessex: We can remove this TODO! <3

jjdressel11:
> What testing have you done, before we merge?

I've tested that all of the Supervision Type drop-down menu items works (entire page loads with data), and that when you go back to the ALL/Default, that the values are the same as they were on load (the values are not being summed across supervision types). Have also tested that all of the other filter toggles work as expected.

I'm going to hold on merging this until we've finalized all of the calc changes to support these changes so that we can do more testing to verify that the correct number of people are showing up in the caseload table at the bottom with every filter/clicking on a cell in the matrix.

PR #256, opened 2020-03-24
Metric file decompression

### LINKED ISSUE #255: Add ability to handle compressed metric files
Some of our metric files are now untenably large. We will update the backend to compress these on upload to the dashboard data location, but the Node server will need to decompress them before handing content to the frontend.

Decompress content in memory once downloaded from GCS (or the local file system if in demo mode). Use the built-in zlib utilities.

We could write the files to tmp on the file system of the Node server, then decompress them there, then stream the content through a buffer to the frontend. This would alleviate memory issues we will eventually face (and have faced previously) associated with holding all of the uncompressed data in memory, but would require more time to do the extra file I/O.

We could also pass the compressed content all the way to the frontend server and let the frontend decompress, which would mean sending fewer bytes over the wire overall. But it seems it would be wiser to use built-in request compression with HTTP headers and let the browser handle it.

jessex:
## Description of the change

This has been tested successfully in stage and locally, working with both compressed and decompressed files, though there are a couple of areas where I think it can be improved. It's currently deployed in staging.

## Type of change

## Related issues

Closes #255 

### Development

- [x] Lint rules pass locally

### Code review

### DIFF
--- a/server/core/metricsApi.js
+++ b/server/core/metricsApi.js
@@ -95,16 +95,19 @@ const FILES_BY_METRIC_TYPE = {
 /**
  * Converts the given contents, a Buffer of bytes, into a JS object or array.
  */
-function convertDownloadToJson(contents) {
-  let contentBuffer = '';
+function convertDownloadToJson(contents, contentsWrappedInArray) {
+  let contentString = '';
+  const contentBytes = contentsWrappedInArray ? contents[0] : contents;
   try {
     console.log('Attempting decompression of contents...');
-    contentBuffer = unzipSync(contents);
+    const unzippedBuffer = unzipSync(contentBytes);
+    contentString = unzippedBuffer.toString();
   } catch (error) {
-    console.error('An error occurred during decompression, assuming already decompressed...');
-    contentBuffer = contents;
+    console.error('An error occurred during decompression, assuming already decompressed...',
+      error);
+    contentString = contents;
   }
-  const stringContents = contentBuffer.toString();
+  const stringContents = contentString.toString();
 
   if (!stringContents || stringContents.length === 0) {
     return null;
@@ -192,7 +195,7 @@ function fetchMetrics(stateCode, metricType, isDemo, callback) {
       const results = {};
       allFileContents.forEach((contents) => {
         console.log(`Fetched contents for fileKey: ${contents.fileKey}`);
-        const deserializedFile = convertDownloadToJson(contents.contents);
+        const deserializedFile = convertDownloadToJson(contents.contents, source === 'GCS');
         results[contents.fileKey] = deserializedFile;
       });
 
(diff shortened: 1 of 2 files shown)


server/core/metricsApi.js
```diff
@@ -94,8 +95,20 @@ const FILES_BY_METRIC_TYPE = {
...
  * Converts the given contents, a Buffer of bytes, into a JS object or array.
  */
-function convertDownloadToJson(contents) {
-  const stringContents = contents.toString();
+function convertDownloadToJson(contents, contentsWrappedInArray) {
+  let contentString = '';
+  const contentBytes = contentsWrappedInArray ? contents[0] : contents;
+  try {
+    console.log('Attempting decompression of contents...');
+    const unzippedBuffer = unzipSync(contentBytes);
```
jessex: This is the synchronous unzip operation, but I'd like if we unzipped each file asynchronously of each other. I had some issues structuring that. Any ideas, @macfarlandian?

server/core/metricsApi.js
```diff
@@ -94,8 +95,20 @@ const FILES_BY_METRIC_TYPE = {
...
+    contentString = unzippedBuffer.toString();
+  } catch (error) {
+    console.error('An error occurred during decompression, assuming already decompressed...',
```
jessex: Our way of determining if a file is already decompressed and doesn't need compression is try-catch, and I hate using try-catch for execution flow. But I wanted to unblock staging and couldn't figure out how to use the zlib API to, e.g. look at file headers and make a call. Any ideas, @macfarlandian?

server/core/metricsRefresh.js
```diff
@@ -30,7 +30,7 @@ const demoMode = require('../utils/demoMode');
 
 const isDemoMode = demoMode.isDemoMode();
 
-const METRIC_REFRESH_INTERVAL_MS = 1000 * 60 * 5; // Refresh metrics every 5 minutes
+const METRIC_REFRESH_INTERVAL_MS = 1000 * 60 * 15; // Refresh metrics every 15 minutes
```
jessex: Just a hunch this was causing undue load on the server...

jessex:
I just pushed up an update which tried to restructure the code to ensure the decompression would happen asynchronously, but I'm not 100% sure it's working and there was on wall-time improvement. Definitely in _Callback Hell_ over here.

server/core/metricsApi.js
macfarlandian: It doesn't look like there is an easy API for this, you would have to read the first couple of bytes off the raw stream

jessex: Got it. Thanks for taking a look!

jessex:
Closing this because it doesn't look like it's going to be used any time in the near future. We can revive it if necessary.

Issue #286, opened 2020-04-07
Migrate non-Lantern views to API per page


jessex:
**Is your feature request related to a problem? Please describe.**
As part of recent stability and performance improvements, the data-intensive Lantern views were migrated from a more monolithic API-request-per-page pattern to a more nimble and performant API-request-per-chart pattern. We should use this same pattern in the non-Lantern views:
1. Even if they are not so data intensive, it will still have at least a minor performance improvement
1. It will permit more flexible error handling in the event of API failure. See #285 
1. It will ensure that API request patterns are consistent across the app, reducing dev friction

**Describe the solution you'd like**
Each chart on each non-Lantern view should use common functionality for initiating and processing its own API request. The page itself should show loading spinners in the place of charts which are still loading.

**Describe alternatives you've considered**
The chief alternative is not to do this, with the reasoning being that there is simply no compelling user-facing need at present, and that this will increase non-essential complexity. I'm not sure I buy that argument entirely, given that we have seen it can be difficult to predict when performance issues will actually materialize as new states come online and existing charts are asked to handle more (or more complex) data. So I think this is a safety measure to be defensive against eventual performance issues in the non-Lantern views.

jessex:
This is no longer a relevant goal.

Issue #107, opened 2019-10-16
Build reusable component for visualization cards


jjdressel11:
We should create a "visualization card component" which takes in props for the chart component, the chart headline, the methodology component, and the export id. This will make each page much easier to build, much shorter, much cleaner, and easier to update.

jessex:
This has been gradually performed over several PRs over the last few months. Closing this out.

PR #287, opened 2020-04-07
API per chart in Lantern views

jessex:
## Description of the change

This changes the Lantern Revocation Analysis Matrix view from a request pattern of API per page to API per chart. The major changes here are:
- Moving control of API requests from Revocations.js into the individual chart components it utilizes
- Adds a loading spinner per chart while each one is individually loading, and makes the spinner visually consistent with the rest of the app
- Extracting out a utility function in metricClient.js to handle fetching and processing an API request for a particular metric file
- Updating the server code to support fetching by a single file (though this can be generalized further)
- Simplifying the caching mechanism to rely on the native refresh functionality in the cache manager we use, which 1) removes unnecessary complexity from our codebase, 2) avoids a double-caching issue where we might naively cache by both metric type and metric type + file thus doubling the memory footprint, and 3) makes the cache refresh functionality more sane by tying it to cache expiration and not absolute intervals in time

The end result is a 10x-100x improvement in overall/perceived page load time, greater flexibility in the face of networking issues, and the removal of (hopefully) most timeout errors that might occur over poor connections.

## Type of change

## Related issues

Inspiration for #286 

### Development

### Code review


jessex:
@macfarlandian definitely I want to generalize the whole dataflow piece into its own component for easier reuse. I had filed a general-purpose issue for decomposition of chart components at #107 and I imagine this could slot into that work. But I do want to get this ready for the next launch so I'd like to defer that work if it's fine by you. I think it would be a good candidate for a tech debt week coming up! Let me know if you disagree strongly, otherwise feel free to approve.

PR #290, opened 2020-04-10
Fixes a memory leak / infinite regression of page reloads in the MO v…

jessex:
## Description of the change

Not sure how this wasn't spotted during testing. My fault!

## Type of change

### Development

### Code review

### DIFF
--- a/src/components/charts/new_revocations/RevocationMatrix.js
+++ b/src/components/charts/new_revocations/RevocationMatrix.js
@@ -94,7 +94,6 @@ const RevocationMatrix = (props) => {
     processResponse();
   }, [
     apiData,
-    filteredData,
     awaitingApi,
     props.filterStates,
     props.metricPeriodMonths,


src/components/charts/new_revocations/RevocationMatrix.js
```diff
@@ -94,7 +94,6 @@ const RevocationMatrix = (props) => {
     processResponse();
   }, [
     apiData,
-    filteredData,
```
macfarlandian, line 97: please add a comment here about why we're excluding this dependency (and why we think it's safe to do so) to prevent us from coming back and inadvertently breaking this again later

jessex: I'm not sure about that, actually. This is removing something that never should have been added in the first place. And commenting on why this isn't here would logically seem to require commenting on why a host of other stateful variables are not here either. Does that seem sound to you or do you still want the comment?

macfarlandian: oh i see ... the fact that we are overloading the variable name `filteredData` inside this function confused me! so it's not actually a dependency, nevermind, carry on

PR #303, opened 2020-04-29
Add export image opportunity to svg charts

### LINKED ISSUE #87: Export as image for svg charts
Currently the export as image functionality only works with canvas elements. Our map charts render as SVGs, so we need to expand the export functionality to support svg chart exports.

DmitryJakubovsky:
## Description of the change

I added the "Export Image" button to all svg charts with map. I find out that to resolve this issue already exists html2canvas library and used it.

Also I find out new bug: on `/revocations` page when you click on Map button and back on "Graph" in REVOCATIONS BY MONTH chart, chart on the right (REVOCATIONS BY OFFICER) is rebuilt. Looks a little bit annoying. I didn't find this in open issues.

## Type of change

## Related issues

Closes #87

### Development

### Code review


DmitryJakubovsky:
@jessex My editor made a few related small code style fixes. Is it ok?

jessex:
@DmitryJakubovsky thanks for calling out the issue you found! I have noticed it myself but for some reason have not yet filed a bug for it. I believe it's due to one of the props that is being passed into that by-officer chart being a dependency for a `useEffect` hook in the chart component, but the data structure of that prop being touched in the view itself, thus triggering a reload of the chart that otherwise shouldn't need to happen. I'll file a task for it.

src/views/tenants/us_nd/FreeThroughRecovery.js
```diff
@@ -119,9 +119,7 @@ const FreeThroughRecovery = () => {
                         Export
                       </a>
                       <div className="dropdown-menu" aria-labelledby="exportDropdownMenuButton-ftrReferralCountByMonth">
-                        {geoViewEnabledRCOT === false && (
-                          <a className="dropdown-item" id="downloadChartAsImage-ftrReferralCountByMonth" href="javascript:void(0);">Export image</a>
-                        )}
+                        <a className="dropdown-item" id={`${geoViewEnabledRCOT ? 'downloadHtmlElementAsImage' : 'downloadChartAsImage'}-ftrReferralCountByMonth`} href="javascript:void(0);">Export image</a>
```
jessex, line 122: Since we perform this id-switching logic in a variety of places, let's extract it out to a utility function somewhere that takes in the chart suffix as an argument, e.g. 
```
function resolveImageDownloadId(geoViewEnabled, chartSuffix) {
  const prefix = geoViewEnabled ? 'downloadHtmlElementAsImage' : downloadChartAsImage';
  return `${prefix}-${chartSuffix}`;
}
```

DmitryJakubovsky: I fixed map charts. 
About resolving id: maybe better to use conditional operator? It will make markup more declarative and implicit. And it looks like code [below](https://github.com/Recidiviz/pulse-dashboard/blob/b91b53b7084161b7a908ef0791422ab94369e541/src/views/tenants/us_nd/Reincarcerations.js#L261).

DmitryJakubovsky:
@jessex And one more issue. There is used html2canvas library, but it does not support converting svg to image for IE11. Is it critical for this moment?

jessex:
Yes, IE11 support is important as a somewhat significant portion of our userbase is on IE11. I had not tested this in IE11 myself just yet but was planning to. If these map exports are not working in IE11, then we should look to another library to perform this for us.

DmitryJakubovsky:
To fix the IE11 problem I add one of html2canvas versions depends on the browser. You can read more about it [here](https://github.com/niklasvh/html2canvas/issues/1756#issuecomment-483971384). If you think that it looks then I will try to find another library.
I have only Mac so I tested it on the Virtualbox for IE11 and Edge.

public/index.html
```diff
@@ -65,5 +78,6 @@
       To begin the development, run `npm start` or `yarn start`.
       To create a production bundle, use `npm run build` or `yarn build`.
     -->
-  </body>
-</html>
+</body>
+
+</html>
```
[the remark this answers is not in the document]
DmitryJakubovsky: Sorry for this. I've already turned off autoformat in my editor.

PR #308, opened 2020-04-30
Aligned all cards

### LINKED ISSUE #50: Ensure consistent visualization card height
When we have rows with multiple cards per row, the card heights can be different heights when the content inside (for example, header text) is of different lengths. Ensure height consistency in a way that is dynamic and responsive.

DmitryJakubovsky:
## Description of the change

Aligned all cards on the North Dakota pages.

## Type of change

## Related issues

Closes #50

### Development

### Code review


DmitryJakubovsky:
I have not found a universal solution for the alignment problem. We should use an individual solution for each card, because all cards have different content, and the content of the related cards will never look the same.

If we expand the header, all became good while we didn't click the `Methodology` link - all content of the neighbor card will dive to the bottom.
If we expand any other block, the current problem will stay with us.

So I think the current solution the most optimal at this moment.

jessex:
Thanks for the thoughts, @DmitryJakubovsky. I'll sync up with Product and see if we want to go forward with this one. It might be that this situation is less optimal than the current one, wherein each card has reasonable internal spacing but occasionally are misaligned in slight ways.

jessex:
@DmitryJakubovsky after conferring with Product, we decided to leave the current layout as is. Thanks for tackling and exploring this, though!

PR #316, opened 2020-05-07
Added extended sort opportunity to CaseTable

### LINKED ISSUE #312: Make the "Revoked Individuals" list sortable by column
No specific problem, but a requested "upgrade" to the current MO dash that increases the user-friendly problem-solving because they can do things like sort by officer and risk level.

Next to each column heading, add "sort" icons to indicate that it is possible to sort the chart by that particular column. 

N/A - an alternate would be not addressing this and letting the table remain unsortable.

DmitryJakubovsky:
## Description of the change

A little bit refactored CaseTable.js: extracted pagination, added useSort hook.

## Type of change

## Related issues

Closes #312

### Development

### Code review


jessex:
Copying over from the issue for posterity:

- Yes, click 3 times to cycle through ascending, descending, and default
  - _Nothing to change here, this is great_
- Yes, the sort icons should always appear in the header. Your UX suggestion works, where default is both light grey, then top arrow becomes dark grey for ascending, then flip to bottom arrow becomes dark grey for descending, then both return to light grey for default.
  - _The change here is to make it so the icons are always visible, with highlighting for whichever sort is enabled at a given time, i.e. across all columns, only one of the arrows should ever be highlighted at a given time_
- Single column sorting is good, no need to add multi-column sorting
  - _Nothing to change here_
- We want to be able to sort on all of them. For violation record, let's try sorting by total number of violations.
  - _The first change here is to add sorting for all columns_
  - _The second change is the only change in this list that adds some real complexity. Basically, we want to be able to sort the violation record column by the total number of violations listed in that string. I think the way you'd probably do this, @DmitryJakubovsky, is to update the `parseViolationRecord` function to also sum up the `record.number` values that get parsed by the regex, and return that alongside the final violation record string. Then we'd keep that summed number in memory with its respective row and sort on that number. But crucially, we would make sure not to display that number anywhere and not to include it in data exports. It would be very nice to have this sort feature work, but if you find that there is too much complexity @DmitryJakubovsky then let's just not have sorting enabled for this column._

src/hooks/usePrevious.js
```diff
@@ -0,0 +1,30 @@
...
+  const ref = useRef();
+
+  useEffect(() => {
+    ref.current = value;
+  }, [value]);
+
+  return ref.current;
+}
+
+export default usePrevious;
```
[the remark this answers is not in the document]
DmitryJakubovsky: Fixed.

src/utils/charts/violationRecord.js
```diff
@@ -0,0 +1,93 @@
...
+  const aRecords = parseViolationRecord(aRecordLabel).sort(violationComparator);
+  const bRecords = parseViolationRecord(bRecordLabel).sort(violationComparator);
+
+  const aSum = sumViolationRecords(aRecords);
+  const bSum = sumViolationRecords(bRecords);
+
+  if (aSum > bSum) return 1;
+  if (aSum < bSum) return -1;
+
+  for (let i = 0; i < aRecords.length; i += 1) {
```
jessex, line 52: Nice job coming up with a secondary sort here! But I think we may have the secondary sort a bit incorrect. For example, if I sort in descending order, it will show me those with the _most_ violation records first. Within those with the same sum of violation records, it will then sort on severity in ascending order, so those with less severity will show up before those with more severity. However, that's a bit counterintuitive for users because more violations is also considered more severe, so the flow of violation severity should be in the same direction as that of violation record sum.

To be clear, that means when you're descending (higher sums first), the more severe violations should show up first when you're in a tie for overall sum, and when you're ascending (lower sums first), the less severe violations should show up first when you're in a tie for overall sum. I _think_ that just means reversing how you're treating severity in here right now, but I could be wrong.

src/components/charts/new_revocations/CaseTable/useSort.js
```diff
@@ -17,14 +17,22 @@
...
+
+const RISK_LEVEL_PRIORITY = [
+  "NOT_ASSESSED",
+  "LOW",
+  "MEDIUM",
+  "HIGH",
+  "VERY_HIGH",
+];
+
+const OFFICER_RECOMENDATION_PRIORITY = [
```
jessex, line 30: I imagine we'll want this to be in severity order, in which case it should go (in ascending order): WARNING, CONTINUATION, CITATION, REVOCATION

DmitryJakubovsky: Thanks! Didn't know the right order so by default sort it alphabetically. Fixed.

src/utils/charts/violationRecord.js
DmitryJakubovsky: Fixed the secondary sort.
P.S. In the test data I found out violation records like `3misd; 2misd;1absc`. Should we aggregate it to "5misd, 1absc" or not?

src/components/charts/new_revocations/CaseTable/useSort.js
```diff
@@ -16,15 +16,73 @@
...
+  if (RISK_LEVEL_PRIORITY.indexOf(a) > RISK_LEVEL_PRIORITY.indexOf(b)) {
+    return 1;
+  }
+  if (RISK_LEVEL_PRIORITY.indexOf(a) < RISK_LEVEL_PRIORITY.indexOf(b)) {
+    return -1;
+  }
+  return 0;
+}
+
+function compareOffecerRecomendations(a, b) {
```
jessex, line 71: Nit: typo in "officer" but I won't block merging on this if there are no other issues.

src/components/charts/new_revocations/CaseTable/useSort.js
```diff
@@ -16,15 +16,73 @@
...
+
+const OFFICER_RECOMENDATION_PRIORITY = [
+  "WARNING",
+  "CONTINUATION",
+  "CITATION",
+  "REVOCATION",
+];
+
+function comparePersonExternalIds(a, b) {
+  if (parseInt(a, 10) > parseInt(b, 10)) return 1;
```
jessex, line 38: We changed from a simple string sort to an integer sort, I guess because the fixture data only has integers. But the actual data in production may include characters in the state ids, so we do _not_ want to change the functionality for this one.

src/components/charts/new_revocations/CaseTable/useSort.js
```diff
@@ -16,15 +16,73 @@
...
+
+  if (!aOfficer && bOfficer) return 1;
+  if (!bOfficer && aOfficer) return -1;
+
+  if (aOfficer.toLowerCase() > bOfficer.toLowerCase()) return 1;
+  if (aOfficer.toLowerCase() < bOfficer.toLowerCase()) return -1;
+  return 0;
+}
+
+function compareRiskLevel(a, b) {
```
jessex, line 61: Can we not simplify this function to `return RISK_LEVEL_PRIORITY.indexOf(a) - RISK_LEVEL_PRIORITY.indexOf(b)`?

src/components/charts/new_revocations/CaseTable/useSort.js
```diff
@@ -60,37 +118,31 @@ function useSort() {
...
-    if (String(a2.state_id) > String(b2.state_id)) return 1;
-    if (String(a2.state_id) < String(b2.state_id)) return -1;
-
-    return 0;
+    return (
+      (sort.field === "state_id" &&
+        comparePersonExternalIds(a2.state_id, b2.state_id)) ||
+      (sort.field === "district" &&
+        compareDistricts(a2.district, b2.district)) ||
+      (sort.field === "offecer" && compareDistricts(a2.officer, b2.officer)) ||
```
jessex, line 126: Another typo: "officer" not "offecer"

The sort field will never be "offecer" so this shouldn't actually work as-is

src/utils/charts/violationRecord.js
jessex: Thanks for the heads up, @DmitryJakubovsky. That's an issue in the fixture data that does not happen in production. We don't need to do anything for this.

src/components/charts/new_revocations/CaseTable/useSort.js
```diff
@@ -0,0 +1,155 @@
...
+        )) ||
+      (sort.field === "violation_record" &&
+        compareViolationRecords(
+          a2.violation_record,
+          b2.violation_record,
+          sort.order
+        )) ||
+      // default sorts
+      compareDistricts(a2.district, b2.district) ||
+      compareOfficers(a2.officer, b2.officer) ||
```
jessex, line 142: I'm not sure this is working correctly. When you look at the default sort, you see everything sorted by district, but then the officer names within a district are _not_ sorted. Take a look at the example below. It's functioning as though it goes from districts straight to person ids, skipping officer names. So there must be something with the implementation here.

Related: when you try to sort by officer specifically, it doesn't work at all. I don't know if this is because of the "offecer" typos in a few places or what.

src/components/charts/new_revocations/CaseTable/useSort.js
```diff
@@ -0,0 +1,155 @@
...
+];
+
+function comparePersonExternalIds(a, b) {
+  if (parseInt(a, 10) > parseInt(b, 10)) return 1;
+  if (parseInt(a, 10) < parseInt(b, 10)) return -1;
+  return 0;
+}
+
+function compareDistricts(a, b) {
+  if (a.padStart(2, "0") > b.padStart(2, "0")) return 1;
```
jessex, line 44: This breaks on data in staging or production because there can be cells with empty values for almost every field, including `district`. So this can get called with either `a` or `b` as null or undefined, and as-is it just breaks. This is why we do the null-safety around cells in the `CaseTable` component.

jessex:
@DmitryJakubovsky Github accidentally split my comments across a couple of reviews, so don't just look at the most recent review that requested changes. There are more important comments just above that review.

src/components/charts/new_revocations/CaseTable/useSort.js
DmitryJakubovsky: Fixed.

src/components/charts/new_revocations/CaseTable/useSort.js
DmitryJakubovsky: Yes, most part of the time I work with test data. Fixed.

src/components/charts/new_revocations/CaseTable/useSort.js
DmitryJakubovsky: You are right. Done.

src/components/charts/new_revocations/CaseTable/useSort.js
DmitryJakubovsky: Fixed.

src/components/charts/new_revocations/CaseTable/useSort.js
DmitryJakubovsky: Fixed all sorts.

src/components/charts/new_revocations/CaseTable/useSort.js
DmitryJakubovsky: Returned everything as it was.

src/components/charts/new_revocations/CaseTable/useSort.js
```diff
@@ -35,53 +35,47 @@ const OFFICER_RECOMENDATION_PRIORITY = [
...
-    OFFICER_RECOMENDATION_PRIORITY.indexOf(a) >
-    OFFICER_RECOMENDATION_PRIORITY.indexOf(b)
-  ) {
-    return 1;
-  }
-  if (
-    OFFICER_RECOMENDATION_PRIORITY.indexOf(a) <
+function compareOfficerRecomendations(a, b) {
+  return (
+    OFFICER_RECOMENDATION_PRIORITY.indexOf(a) -
```
jessex, line 76: It looks like there are some additional values in staging/production that are not accounted for here. My bad! The list is actually fairly long and may grow over time, so I think we should go back to just sorting this column alphabetically. Sorry for misleading you!

DmitryJakubovsky: Returned alphabetical sort.

Issue #326, opened 2020-05-11
[Nav Redesign] Adjust app navigation elements (non-Lantern)


jessex:
**Is your feature request related to a problem? Please describe.**
We will roll out the navigation redesign feature iteratively, in a way that will not break functionality for the app between commits, in case a release needs to go out in the interim.

The mini-PRD only goes so far as reorganizing existing charts into new views that match the updated conceptual hierarchy--it does not yet propose new charts. This second task updates the actual navigation elements, building on #325. It updates the sidebar (visible only in the non-Lantern views for now), updates app routes for the non-Lantern views, 

**Describe the solution you'd like**
1. Match the sidebar UI spec in the PRD
1. Remap the app routes to an organizational hierarchy matching the nav hierarchy
    1. dashboard.recidiviz.org/community/goals, dashboard.recidiviz.org/community/explore, dashboard.recidiviz.org/facilities/explore 
    1. This only applies to the non-Lantern views for now. A later task handles the Lantern views
1. Point the new app routes at the new views
1. Ensure the existing logic for specifying available views for a state can handle this nesting (_shouldn’t_ be an issue)
1. Decide on the default landing points for each existing state
1. Remove the legacy views and routes
1. **Do we want existing app routes to be kept around but redirect to a particular page to ensure bookmarks and links aren’t broken?** How would they be mapped? To relevant explore pages?

**Describe alternatives you've considered**
N/A

**Additional context**
This is a child of #324.

jessex:
@kbinney there is a question for you, bolded in the description.

kbinney:
Keeping existing app routes and redirecting would be great. 

I'd map: 

- dashboard.recidiviz.org -> dashboard.recidiviz.org/community/goals
- dashboard.recidiviz.org/snapshots -> dashboard.recidiviz.org/community/goals
- dashboard.recidiviz.org/revocations -> dashboard.recidiviz.org/community/goals
- dashboard.recidiviz.org/reincarcerations -> dashboard.recidiviz.org/facilities/goals
- dashboard.recidiviz.org/programEvaluation/freeThroughRecovery -> dashboard.recidiviz.org/programming/explore

kbinney:
Currently, the collapsed dashboard shows page icons users can switch between. The nested sidebar (new version described in this PRD) will work as follows:

- The top level headings (Community, Explore, Programming) will have an icon like the current sidebar. We will use the same icons. 
    - The “Snapshots” icon will become the community icon
[...]

PR #353, opened 2020-05-21
Feature redesigned navigation

DmitryJakubovsky:
## Description of the change

* Changed navigation menu on no Lantern pages.
* Added redirects to old routes (except `/revocations` endpoint because the same endpoint exists on the Lantern page)

## Type of change

## Related issues

> Closes [#326]

### Development

### Code review


src/components/PrivateTenantRedirect.js
```diff
@@ -0,0 +1,49 @@
+// Recidiviz - a data platform for criminal justice reform
+// Copyright (C) 2019 Recidiviz, Inc.
```
jessex, line 2: Nit: 2020

src/components/topbar/TopBar.js
```diff
@@ -0,0 +1,76 @@
...
+        <TopBarLogo />
+      ) : (
+        <ul className="nav-left">
+          {isAuthenticated && <TopBarHamburgerMenu onClick={toggleSideBar} />}
+          <TopBarTitle title={normalizedPath} />
+        </ul>
+      )}
+      <ul className="nav-right">
+        {isAuthenticated ? (
+          <TopBarForAuthenticatedUser user={user} onLogout={onLogout} />
```
jessex, line 63: The names `TopBarForAuthenticatedUser` and `TopBarForGuest` seem like they're going to be alternative versions of the top bar for different kinds of users, but actually they are alternative versions of a specific component, namely the user menu in the top right. I would suggest renaming these to `TopBarUserMenuForAuthenticatedUser` and `TopBarUserMenuForGuest`, or something like that.

src/views/tenants/us_mo/community/Revocations.js
```diff
@@ -20,31 +20,32 @@ import RenderInBrowser from 'react-render-in-browser';
...
 
-import Loading from '../../../components/Loading';
-import '../../../assets/styles/index.scss';
-import { useAuth0 } from '../../../react-auth0-spa';
-import { callMetricsApi, awaitingResults } from '../../../utils/metricsClient';
-import { matrixViolationTypeToLabel, toInt, violationCountLabel } from '../../../utils/transforms/labels';
-import useTopBarShrinking from '../../../hooks/useTopBarShrinking'
+import Loading from '../../../../components/Loading';
+import '../../../../assets/styles/index.scss';
+// import '../../../../assets/styles/index.scss';
```
jessex, line 25: Remember to delete this.

src/hooks/useLayout.js
```diff
@@ -0,0 +1,53 @@
...
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+// =============================================================================
+
+import { useAuth0 } from "../react-auth0-spa";
+import { getUserStateCode } from "../utils/authentication/user";
+import { hasSideBar } from "../utils/layout/filters";
+import { isLanternState } from "../views/stateViews";
+
+function checkIsLantern(user, isAuthenticated) {
+  if (!isAuthenticated) {
```
jessex, line 24: Pursuant to my previous comment about authentication breaking, I tested locally and it appears that changing this conditional to `if (!user || !isAuthenticated)`, and doing the same in `checkIsWide` just below, fixes the login issue!

DmitryJakubovsky: Thanks for the research! Fixed.

src/views/tenants/us_mo/community/Revocations.js
DmitryJakubovsky: Removed.

src/components/topbar/TopBar.js
DmitryJakubovsky: Yep, you right. Renamed components.

src/components/PrivateTenantRedirect.js
DmitryJakubovsky: Done.

DmitryJakubovsky:
I moved the styles for TopBar and SideBar to corresponding component folders, because they are called only there. If you think that it is a bad idea I will return as it was.

kbinney:
This is looking great!
 Noticed a few things when running locally:

1. The side bar has more space between a section header and section component than between the last section component and the next section header. This should be the reverse. IE in the below picture, "Goals" appears farther from "Community" than "explore" from "facilities." I'd like there to be less space between goals and community and more space between explore and facilities. 

Screen Shot 2020-05-27 at 9 41 09 AM

2. The section headers currently change color when moused over (see below for mouse over of Community). They should not do that. Having them change color makes them seem "clickable" and they aren't clickable. 

Screen Shot 2020-05-27 at 9 48 07 AM

3. The current page's bolding should be removed when the user is mousing over other page options. In the below photo, the user is mousing over the Explore page of the Facilities section. Since the use is no longer trying to select the Goals page of the Community section, the bolding should disappear, but it is still bold in the photo below.

Screen Shot 2020-05-27 at 9 50 09 AM

4. The right carrot would ideally be the same height as the page title. Right now, the top of the carrot seems to be taller than the text (ie top of carrot is above the G and bottom below it in the carrot next to Goals in the photo above). 

5. (Change) Realizing now I hadn't specified anything about the Map option on the charts on the two goals pages. I like that the map is still an option, but without the explore bar at the top, the time period in Map view is unclear. If this doesn't expand the scope too much, I'd like period information to be added to the map view for the 4 charts on the community goals page and the one chart with map view on the facilities goal page. See below for screen shots of one of these charts, the map view of it, and an example of adding period information from another chart. The charts show 3 years of information so the period should read "Month/1/year - present" with the month & year from 3 years ago / the first data point of the x axis in chart mode. For example, if 3 years of data started in June 2017, the period would be (6/1/2017 - present)

If adding the period information adds to the scope of this change, I'd like the map option removed from charts on the two goal pages. 
Screen Shot 2020-05-27 at 9 54 58 AM
Screen Shot 2020-05-27 at 9 55 02 AM
Screen Shot 2020-05-27 at 9 59 26 AM

6. The LSI-R score change (average) chart has a lot of white circles when in Map mode (see below). I suspect this comes from the average sometimes being positive and sometimes negative. The positive/negative coloring should actually be red / blue like in the Admissions versus releases chart (also see below). Ideally, for the LSI-R score change chart, positive changes would be red and negative changes would be blue. Higher magnitudes = darker colors. 

Screen Shot 2020-05-27 at 10 04 18 AM
Screen Shot 2020-05-27 at 10 07 47 AM

jessex:
Let's make numbers 5 and 6 separate issues, @kbinney, since they are expansions/deviations of scope and require non-trivial engineering effort.

@DmitryJakubovsky, we should include the feedback for points 1-4 from Katherine's comment in your current PR. Once those points are incorporated, we can approve and merge.

DmitryJakubovsky:
@jessex I fixed points 1,2,4. With point 3 I have some trouble because I don't have ideas about how to make it beautiful in react style. Using only CSS does not help.

jessex:
@DmitryJakubovsky you will see I just pushed a commit to your branch that fixes the auth issue I noted. It wasn't that the app was automatically logging back in after attempting to logout. It's that logout was just not working to begin with, which I figured out after looking at the Auth0 logs. It is a one-line fix, as you'll see.

All we need to do to wrap this up, then, is to fix the feedback component styling, as I noted in my most recent review.

Issue #339, opened 2020-05-14
[Lantern] When time period = 1 month, show week over week data for the top chart


sychang:
**Is your feature request related to a problem? Please describe.**
A clear and concise description of what the problem is, e.g. we don't currently convey...

**Describe the solution you'd like**
For the chart, "Number of people revoked to prison per month," when the Time Period selected is "1 month," display week over week data. 

The rightmost date should always be the current day. 

The week that is currently in progress should be grayed, similar to how the month that is currently in progress is grayed for the other time periods.
![image](https://user-images.githubusercontent.com/6633887/81893970-631fe100-9574-11ea-8d4f-05860eb5990e.png)

The x-axis markers should be on the 7th, 14th, 21st, and 28th of every month. The rightmost marker should be the date closest to today's date that has not yet passed. For example, if today is May 13, the rightmost edge of the chart should be May 14, with the section between May 7 and May 14 grayed out because this week is currently in progress.

**Describe alternatives you've considered**
N/A

**Additional context**
This could enable adding a trendline when the time period is 1-month (#337).

jessex:
@sychang I previously told this to @cawarren, but this is not something we can easily do. We would need to make significant changes to our calculation infrastructure to support week-over-week calculations. This is not something we can do in the near term unless we prioritize that calculation work.

sychang:
Got it and makes sense - I will drop it to the ice box for now and communicate out the challenges.

PR #358, opened 2020-05-26
Date range preview (only ui)

### LINKED ISSUE #335: [TO ESTIMATE] Add "choose a date range" as an option for "Time Period" filter
We currently allow users to change the time period on the dashboard to view data for 1 month, 3 months, 1 year, and 3 years. However, we don't allow users to choose a custom date range, so they can't easily answer questions such as:
- How many people were revoked in November 2019?
- What was the district with the highest percent of individuals revoked between May and August 2019 (they might be interested in that date because it coincides with a key policy change)?

At the bottom of the "Time Period" filter on the Lantern dashboard, add an option to "Choose a Date Range"

image

TODO: Needs UX work on date picker and how to display the custom date range once selected (unless we happen to already have a component to select a date range elsewhere in the codebase that we can reuse).

Once a date range is picked, update the charts to represent that date range. The subtitles of the charts should also be updated to reflect the time range. For example, if the date range is May 1, 2019 to August 30, 2019:
- The "Number of people revoked to prison per month" chart should begin on the left at May 1, 2019 and end on the right at August 30, 2019
- The subtitle should say "May 1, 2019 to August 30, 2019." Previously, it would have said something like "Last 6 months (12/1/2019 to present)" if the time period selected was "6 months."

image

When "1 month" or "3 months" are selected, allowing users to view the "previous month" or "previous 3 months." That requires a bit more UX revision than this option.

**Additional context**
Requested by MO stakeholder: "It would be nice to have the time period filter to be customizable or at least in one month increments. I can’t pull the April numbers now just for April as it will only do it for May. I have to set the filter to three months and then look at each individually to see what month they were revoked. It might be helpful as well when exporting that it pulls the date of the expiration into the spreadsheet or at least the month and year."

DmitryJakubovsky:
## Description of the change

Only proof of concept. We can override every subcomponent of this datepicker.

## Type of change

## Related issues

> Closes #335

### Development

### Code review


sychang:
1. Because we don't have granularity finer than a month, I would suggest only allowing the user to select down to the month so they don't get the impression that they can make more granular selections than that. With that in mind, here is a proposed flow - the copy is final but the visuals are just illustrative, not exact:

image

Additional notes: 
- There are 2 options for showing the "Custom Date Range" selection - I'm happy to go with whichever option requires less engineering work. If equivalent, let's go with option 2 to create a nicer menu UX if a user selects "Custom Date Range" but then changes their mind and closes the modal without making a date selection (we'd revert to the previous selection).
- Do not let the calendar pickers go further than 36 months in the past from today. 
- Once a start date is selected, make sure the user does not select an end date before the start date by making all dates before the start date inactive, something like this:
image

2. Agree we want to keep the 3 month, 6 month, 12 month, and 36 month options. However, I like what you have done with saying "Last X months" - this helps clarify the meaning of those selections after we add this custom date range.

3. I don't think this is an issue anymore with the new flow but let me know if it still is.

4. Agree we want the text field to be a fixed width. The copy should say "Custom: MM/YYYY - MM/YYYY," for example: "05/2018 - 05/2019"

5. This question may not be relevant anymore given the above mock but let me know if it is.

DmitryJakubovsky:
@jessex @sychang  thanks for clarification! Тow I have the whole picture.
@jessex But I have one question. Now all records in the HTTP data has `metric_period_months` field which can be one of "1", "3", "6", "12", "36". Ho do I filter data if the user will want to use Time Period from 4 to 8 months? Maybe real data has additional fields with ISO dates?

jessex:
That's a great and very important point, @DmitryJakubovsky, that I had not originally considered.

@sychang for any metrics that don't use year+month but instead use metric_period_months (i.e. this is a deduplicated/rationalized metric encompassing this number of months terminating at this month), we would need to generate another data point with metric_period_months=1, =2, =3, =4, ..., =36, for all months in range. This applies to every file except for the Revocations Over Time chart--all other charts use metric_period_months. Right now, we only generate for metric_period_months=1, =3, =6, =12, =36, terminating at the current month. Naively, this would lead to those files which use it being multiplied by 2-3 orders of magnitude in size to support this. So it's similar in spirit to #339: we need to either do some frontend work to support this significant increase in file size that this requires, or do some frontend and calc work to have data fetched in real-time as filters change (which has significant UX implications), or some non-trivial calculation change that alters the format of the data to make this viable.

(The reason we have metric_period_months instead of just summing up month-over-month values is principally because we have to deduplicate these metrics by person, which you can't do if you sum up month-over-month since each month is going to include a variety of data points that correspond to a given person.)

I think this one has to go to the icebox. It turns out to be significant work to support.

sychang:
Is there a way to avoid the file size increase by introducing a page refresh after the user clicks "Apply" in the custom date range modal? This gives us a small UX hit but a modal is already popping up over the page so a refresh is less jarring upon modal close. Or is this off base because we'd still need to pre-generate another data point for all the files? 

If that doesn't work, here is my understanding of the options. This is fairly high priority for MO so I want to make sure I have the right assessment for prioritizing this against other work:

1. Implement a solution that could lead to an up to 1-month delay for other frontend and/or calc work, depending on the solution. If we went with a longer-term approach of changing the way data is organized, would this make future additions, like adding an admission type filter, more straightforward? 

2. Work with states to understand if there is a limited, lower-cost version of this we could build. In one of the requests, a stakeholder specifically asked for a way to look at the month prior to this month (April 2020). Instead of allowing a custom date range selection, allow the addition of:
- [First priority] Last month (If today is May 27, 2020, show April 2020)
- [Lower priority] Last year (show 2019 in full, as opposed to past 12 months)
Would that introduce a small enough increase to file size that would not warrant significant work haul?

3. Send to icebox 

Is that right?

jessex:
Discussed with @sychang offline and we're going to devote a bit of time to triage potential solutions that let us drastically scale up the amount of data handled by the dashboard, which impacts both this issue and a couple of others, and then make a prioritization call for the work accordingly. In the meantime, this PR will sit on ice.

PR #447, opened 2020-08-13
Feature lantern pa specific violation types

### LINKED ISSUE #434: Update Lantern Violation chart with PA-specific violation types
PA has a different set of violations than MO that it's tracking on the Admissions Plot. We would like to reflect these violations in the UI.

In the "Relative Frequency of Violation Types" bar chart, violations on the bar chart (and their colors), from left to right on the x-axis, should be:
- Low tech. (blue)
- Med. tech. (blue)
- Elec. monitoring (blue)
- Subs. use (blue)
- Absconding (blue)
- High tech (blue)
- Summary offense (orange)
- Misdemeanor (orange)
- Felony (orange)

N/A

**Additional context**
These are the same violation types and ordering as #433.

DmitryJakubovsky:
## Description of the change

* A little bit refactored Violation chart
* Composited revocation charts for more flexibility
* Added PA-specific violation types

## Type of change

## Related issues

Closes #434

### Development

### Code review


src/views/tenants/us_pa/community/Revocations.js
```diff
@@ -137,15 +143,77 @@ const Revocations = () => {
...
+            timeDescription={timeDescription}
+          />
+        }
+        violationChart={
+          <RevocationsByViolation
+            dataFilter={allDataFilter}
+            filterStates={filters}
+            stateCode={stateCode}
+            timeDescription={timeDescription}
+            technicalViolationTypes={[
```
jessex, line 160: There are a couple things I don't love about these data structures for capturing the set of violation types for a given state:
1. We need to repeat every value: it's always in `allViolationTypes` and always in one of `technicalViolationTypes` or `lawViolationTypes`.
2. We need to repeat the values between these `Revocations.js` files and the `labels.js` file and they have to be exactly the same or else they won't render correctly in the chart.

To clean this up a bit, what if we got rid of `violationTypeToLabel` in `labels.js` entirely? There's no super great reason to keep it, since it only buys us a bit of de-duplication for states who want to use the same labels in their violation distribution charts. But we can assume that there will eventually be two states that want to use different labels for the same violation type, and then we'd need to revise or remove this anyway.

Instead, I would just pass in a single property to the `RevocationsByViolation` chart that takes in an ordered list of objects that looks something like this:
```
violationTypes = [
    { key: "low_tech_count", label: "Low tech.", type: "TECHNICAL" },
    ...
    { key: "felony_count", label: "Felony", type: "LAW" },
]
```

You could even create a type for this object so that we can ensure each state view that tries to create a `RevocationsByViolation` component abides by the same prop format.

What do you think?

DmitryJakubovsky: I think this is a great idea! 
I wanted to do something similar, but I thought that a lot of code would have to be duplicated. Now, if it is not a problem, I'm glad to make these changes!

src/views/tenants/us_mo/community/Revocations.js
```diff
@@ -48,16 +48,38 @@ import {
...
-  { value: "BOARDS_RETURN", label: "Board Returns" },
+  {
+    key: "intervention_fee_count",
+    label: "Intervention Fees",
+    type: "TECHNICAL",
+  },
+  { key: "special_count", label: "Special Conditions", type: "TECHNICAL" },
+  { key: "weapon_count", label: "Weapons", type: "TECHNICAL" },
+  { key: "substance_count", label: "Substance Use", type: "TECHNICAL" },
+  { key: "municipal_count", label: "Municipal", type: "LOW" },
```
jessex, line 79: Nit: these should be `LAW`, not `LOW`.

src/views/tenants/us_pa/community/Revocations.js
```diff
@@ -65,13 +64,28 @@ const admissionTypeOptions = [
...
+  { key: "med_tech_count", label: "Med tech.", type: "TECHNICAL" },
+  {
+    key: "elec_monitoring_count",
+    label: "Elec. monitoring",
+    type: "TECHNICAL",
+  },
+  { key: "subs_use_count", label: "Subs. use", type: "TECHNICAL" },
+  { key: "absconding_count", label: "Absconding", type: "TECHNICAL" },
+  { key: "high_tech_count", label: "High tech.", type: "TECHNICAL" },
+  { key: "summary_offense_count", label: "Summary offense", type: "LOW" },
```
jessex, line 78: Same here: these should be `LAW`.

src/views/tenants/us_pa/community/Revocations.js
jessex: Love how this turned out!

PR #461, opened 2020-08-31
Feature lantern specific district or region

### LINKED ISSUE #437: [Lantern] Restrict Lantern district filters for users based on user metadata
Full details in PRD: https://docs.google.com/document/d/1bxSg8JgQHFFZZDSjBUKM-fOtU-8r3FVieQD5J81crkA/edit#heading=h.c29ll7o2tvsa

The current Lantern tool is currently restricted to Department of Corrections agency leadership. MO leadership would like to enable regional- and district-level supervisors to be able to access the tool because they are best-positioned to drive operational change in their correctional system. Supervisors are in charge of one particular district or a region (a group of districts), so we want to only show them Lantern data for their own district/region. 

Enable distribution of the Lantern tool beyond MO agency leadership to District Administrators and Regional Supervisors. We would like to do this by restricting the “district” functionality of the Lantern dashboard to the district(s) that a particular supervisor is in charge of. 

District administrators should see the same Lantern dashboard that exists today as if their district filter was selected. They should not be able to change the district that is selected. For simplicity, the UI can remain the same, but with the default district selected and the page filtered accordingly. If a user clicks the dropdown, their district will be the only option present:

Regional supervisors should see the same Lantern dashboard that exists today, with the default showing all of the districts in their region, with copy updated from “All districts” to “All districts in region,” and the ability to select individual districts in their region (in the below example, districts 01, 02, 03, and 04 belong to the regional supervisor’s region and those should be the only four districts that appear in the dropdown). With multi-district select functionality (#424), regional supervisors can select multiple districts at the same time. They should not be able to select a district or region that is not their own.

This requires the following frontend changes:
* The ability to filter which options are selectable in the district dropdown based on the kind of user
* The ability to change which districts count as "ALL" based on the kind of user
* The ability to determine the kind of user based on the user_metadata, similar to how we determine the user's state_code. This should be accomplished by checking for two new fields in the metadata:
- If there is a `region` field in the user metadata, we should look at a static, state-specific mapping from region to districts (with MO's mapping defined in the parent epic) and then limit what's in the dropdown to just those districts and make sure that the ALL option only includes those districts
- If there is a `district` field in the user metadata, we should limit what's in the dropdown to just that district and make sure that's what their data is filtered to at all times

UI alternate
- We could update the district UI to remove the dropdown for district supervisors and replace it with static text indicating the district. This is a minor UX improvement (prevents a potential dropdown click because it more clearly indicates that they cannot change districts) but does not appear to justify the additional design and FE work. Open to discussion.

**Additional context**
Full context in [PRD](https://docs.google.com/document/d/1bxSg8JgQHFFZZDSjBUKM-fOtU-8r3FVieQD5J81crkA/edit#). Feel free to drop any detail comments directly in the doc for easier discussion.

This is a child of the parent epic #430.

### LINKED ISSUE #438: [Lantern] Add ability to specify specific districts or regions in user metadata
Full details in PRD: https://docs.google.com/document/d/1bxSg8JgQHFFZZDSjBUKM-fOtU-8r3FVieQD5J81crkA/edit#heading=h.c29ll7o2tvsa

The current Lantern tool is currently restricted to Department of Corrections agency leadership. MO leadership would like to enable regional- and district-level supervisors to be able to access the tool because they are best-positioned to drive operational change in their correctional system. Supervisors are in charge of one particular district or a region (a group of districts), so we want to only show them Lantern data for their own district/region. 

Enable distribution of the Lantern tool beyond MO agency leadership to District Administrators and Supervisors. We would like to do this by restricting the “district” functionality of the Lantern dashboard to the district(s) that a particular supervisor is in charge of. MO has created a file of the .gov accounts and corresponding districts/regions in the file **"LANTERN.DA_RA_LIST"** in their data warehouse.

We should add two optional new fields to the user metadata in Auth0:
- `region` - if this is present, then we check against a static mapping from region to districts in the codebase to determine which districts to filter to
- `district` - if `region` is not present but `district` is, then we filter to only this district

We should add logic similar to that which fetches state code from user metadata to check these fields and return a list of districts the user should be able to see. This will then be called in #437.

N/A

**Additional context**
Full context in [PRD](https://docs.google.com/document/d/1bxSg8JgQHFFZZDSjBUKM-fOtU-8r3FVieQD5J81crkA/edit#). Feel free to drop any detail comments directly in the doc for easier discussion.

This is a child of the parent epic #430.

DmitryJakubovsky:
## Description of the change

* Added ability for district and region administrators to filter by their jurisdiction.

## Type of change

## Related issues

Closes #437
Closes #438

### Development

### Code review


src/utils/authentication/viewAuthentication.js
```diff
@@ -24,8 +24,13 @@ function getDemoUser() {
     name: 'Demo Jones',
     email: 'notarealemail@recidiviz.org',
     'https://dashboard.recidiviz.org/app_metadata': {
-      state_code: 'RECIDIVIZ',
+      state_code: 'recidiviz',
+      // state_code: 'us_mo',
     },
+    // 'https://dashboard.recidiviz.org/user_metadata': {
```
jessex, line 30: What's the status of this PR? I see this commented out bit and I see some merge conflicts. Is this ready for review?

DmitryJakubovsky: Yes, this PR is ready now. You can uncomment this metadata info for testing functionality as a district or region administrator.

src/utils/authentication/user.js
```diff
@@ -74,3 +82,21 @@ export function getAvailableStateCodes(user) {
 export function doesUserHaveAccess(user, stateCode) {
   return getAvailableStateCodes(user).includes(stateCode);
 }
+
+/**
+ * Returns Auth0 user districts or null if it missed
```
jessex, line 87: Let's reword this docstring. "Returns the district or districts that a user should be limited to viewing data for, if there is user metadata in Auth0 specifying a district or region (group of districts) for the user."

src/views/tenants/us_mo/community/Revocations.js
```diff
@@ -91,10 +96,14 @@ const violationTypes = [
 ];
 
 const Revocations = () => {
+  const { user } = useAuth0();
+  const { district } = getUserMetadata(user);
+  const userDistricts = getUserDistricts(user);
```
jessex, line 101: It doesn't appear this filtering is working quite in the way you expect it, particularly when there is a `region` but not a `district`. Although the district dropdown filter does only include the districts for that region and although the Revocation Matrix chart does appear to be filtered to only include data points for that region's set of districts, it does _not_ appear that all of the other charts are being filtered to only that set of districts. For example, comment your testing block in `viewAuthentication` so that the `region` is set but not the district, and you'll see that the charts for Admissions by Month and the Admitted Individuals list (and potentially other charts, as well) are showing the same data, and that it appears to be the same data as when there is no specified `region`. Furthermore, for the Admissions by District chart, it does not appear to be highlighting the bars for the specified `district` or `region`, as called for in the PRD.

So I think that the district filtering is not being applied everywhere as expected.

src/utils/authentication/user.js
```diff
@@ -27,6 +27,14 @@ export function getUserAppMetadata(user) {
   return user[appMetadataKey];
 }
 
+/**
+ * Returns the Auth0 user_metadata for the given user.
+ */
+export function getUserMetadata(user) {
```
jessex, line 33: I think I misled you by mistake with my comments in the two issues. We don't want to use `user_metadata` for this feature, but rather the existing `app_metadata` that we already use for `state_code`. Sorry about that!

We should be able to get rid of this function and retrieve the district and region from the response to `getUserAppMetadata`.

src/components/charts/new_revocations/ToggleBar/DistrictFilter.js
```diff
@@ -26,74 +26,90 @@ import sortBy from "lodash/fp/sortBy";
...
-      )(apiData),
-    [apiData, defaultValue]
-  );
+  const { district, region } = getUserMetadata(user);
+  const userDistricts = getUserDistricts(user);
+
+  if (district) {
+    return (
+      <FilterField label="District">
+        <Select
```
jessex, line 50: I noticed that two errors are printed to the console when we specify a particular district, pointing to this line:

Screen Shot 2020-09-02 at 13 11 15

src/views/tenants/us_pa/community/Revocations.js
```diff
@@ -87,19 +92,24 @@ const violationTypes = [
 ];
 
 const Revocations = () => {
+  const { user } = useAuth0();
```
jessex, line 95: I suspect that this pattern of invoking Auth0 and then immediately trying to get data off of the user may not work how you expect. That `useAuth0` hook executes asynchronously, right? The `user` that gets passed immediately into the next two calls may not be resolved yet, right?

jessex (changes requested):
Thanks, @DmitryJakubovsky! Some code comments and also some issues noted in my testing.

src/components/charts/new_revocations/helpers.js
```diff
@@ -135,3 +135,11 @@ export const formattedMatrixFilters = (filters) => {
   }
   return parts.join(", ");
 };
+
+export const transformFilters = (filters, userDistricts) => {
```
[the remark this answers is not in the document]
DmitryJakubovsky: Yes, you are right, renamed this function name

src/utils/authentication/user.js
DmitryJakubovsky: Renamed. Sorry for the meager description.

src/views/tenants/us_mo/community/Revocations.js
DmitryJakubovsky: Fixed all defects.

src/views/tenants/us_pa/community/Revocations.js
jessex: You are right, and I was mistaken. I tested with a test user with live authentication and this worked as expected. Thanks!

src/views/tenants/us_mo/community/Revocations.js
jessex: All appears to be working as expected. Thanks!

PR #470, opened 2020-09-07
Completely removed jquery

Wmaileq:
## Description of the change

- removed unused deps that use jquery as dep
- replaced bootstrap-jquery functions in favor of react-bootstrap components
- replaced perfect-scrollbar by react-perfect-scrollbar
- removed masonry-layout unused dep (not part of this PR)

## Type of change

- [x] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Configuration change (adjusts configuration to achieve some end related to functionality, development, performance, or security)

## Related issues

> Closes #24 

### Development

These boxes should be checked by the submitter prior to merging:

- [x] Manual testing against realistic data has been performed locally

### Code review

These boxes should be checked by reviewers prior to merging:

- [x] This pull request has a descriptive title and information useful to a reviewer
- [x] This pull request has been moved out of a Draft state, has no "Work In Progress" label, and has assigned reviewers
- [x] Potential security implications or infrastructural changes have been considered, if relevant

PR #472, opened 2020-09-08
added missing prop-types

### LINKED ISSUE #143: Add props validation to all React components
We are not using [props validation](https://reactjs.org/docs/typechecking-with-proptypes.html) in any of our custom React components. It's not only best practice, it's also the most common linter warning in the codebase as of time of writing.

Wmaileq:
## Description of the change

- Added missing prop-types
- Added shape prop-types according to api data

## Type of change

## Related issues

> Closes #143

### Development

### Code review


Wmaileq:
I did not cover next files with prop-types:
src/components/charts/common/** - as this files are not used
src/components/charts/ExportMenu.js - as file was changed in another pr
src/components/charts/GeoViewTimeChart.js - maybe we need to refactor this file to functional component and separate it into several components?

src/components/charts/programming/FtrReferralCountByMonth.js
```diff
@@ -204,7 +205,27 @@ const FtrReferralCountByMonth = ({
   return chart;
 };
 
-FtrReferralCountByMonth.defaultProps = {};
-FtrReferralCountByMonth.propTypes = {};
+FtrReferralCountByMonth.defaultProps = {
+  header: undefined,
```
jessex, line 209: Question: why is this defaulted to `undefined`? It doesn't look like that's a general pattern we're following here for other non-required prop types.

jessex (changes requested):
Thanks for this, @Wmaileq!

For the files you noted in your comment:
- Would you mind removing the following charts from `chart/common`: `ProgramCostEffectiveness`, `RecidivismRateByProgram`, `ReincarcerationRateByReleaseFacility`, `ReincarcerationRateByTransitionFacility`?
- Is that other PR adding prop types for `ExportMenu` directly?
- Let's skip over refactoring `GeoViewTimeChart` for now--it's not high priority to clean up that code, though I agree it can be improved. Instead, let's just make sure it's covered 

Finally, I noticed two errors from the new prop types while running the app in demo mode locally:

The Community > Goals page for ND has two prop type errors:
Screen Shot 2020-09-08 at 13 52 10

The Lantern view for PA has several prop type errors specifically related to supervisionType, since the PA view doesn't have a supervision type filter. That should probably be made optional like admission type? Here are a couple that you see when opening up the app for PA:
Screen Shot 2020-09-08 at 13 55 05

src/components/charts/programming/FtrReferralCountByMonth.js
Wmaileq: Actually, i dont think that default to undefined is good practice, but in other places, e.g.:
`src/components/charts/community/SupervisionSuccessSnapshot.js`
`src/components/charts/community/LsirScoreChangeSnapshot.js` and 25 other matches undefined is used in defaultProps.
I agree that it is not good pattern and i think i have to fix this issue by replacing `undefined` with `null`

Wmaileq:
@jessex 

- [ ] Would you mind removing the following charts from chart/common: ProgramCostEffectiveness, RecidivismRateByProgram, ReincarcerationRateByReleaseFacility, ReincarcerationRateByTransitionFacility? 
- removed

- [ ] Is that other PR adding prop types for ExportMenu directly? 
- yes, this propTypes added in #470 

- [ ] Let's skip over refactoring `GeoViewTimeChart` for now--it's not high priority to clean up that code, though I agree it can be improved. Instead, let's just make sure it's covered 
- propTypes added

Additionally generilized` metricType ('counts', 'rates')` into constant and replaced all usages.

One thing left - remove `undefined` from `defaultProps`

jessex (commented):
@Wmaileq everything is looking good! Just need to wait for your `undefined` change and then we should be ready to merge this one.

jessex (changes requested):
It looks like there is still one prop type error on the Facilities Explore page:
Screen Shot 2020-09-10 at 14 16 24

jessex (approved):
Thanks much, @Wmaileq!

PR #477, opened 2020-09-10
Supervision level filter

### LINKED ISSUE #431: Separate "Supervision Level" dropdown for PA
More details and context in this doc: https://docs.google.com/document/d/1ndwZw8PSAIJfLQkuUZmWEKeeqVH6S2PxaGeGXRsuKds/edit#heading=h.6pe088gkbuk8.

PA would like to see supervision levels in addition to case types in the admissions plot.

Add a Supervision Level dropdown to the right of the Case Type filter. The dropdown should contain the following options:

- All (selected by default)
- Enhanced supervision
- Maximum supervision
- Medium supervision
- Minimum supervision
- Special circumstance supervision
- Monitored supervision

PA originally proposed for us to include the supervision levels under the "case type" dropdown, but that does not map well to our schema, since there can be multiple case types per supervision period, but only one supervision level per supervision period. After discussing with @ageiduschek we decided it was best to keep them as separate dropdowns.

**Additional context**
N/A

Wmaileq:
## Description of the change

- Added supervision level filter

## Type of change

## Related issues

> Closes #431

### Development

### Code review


jessex:
Yeah I noticed that myself, too, @Wmaileq. I am going to need to do a bit more optimization on the client side. What's happening is that although the server is passing the optimized metric file format to the client, the client is expanding it back into the original format so that each chart component can keep processing the data points in the same way it is now, which means we are consuming way more memory in the browser than we need to. We knew that this would be a necessary further optimization but we weren't sure how quickly it would come. It's also true that the fixture files themselves are bigger than we expect the production data files to be because of their contrived nature: I am just multiplying the number of rows when adding in new dimensions to the fixtures, but in reality not all of these combinations actually exist.

Either way, thanks for getting this in place. I'll take this one over from here and merge this when we have the client side performing better with the updated data.

jessex (approved):
Thanks much, @Wmaileq! With my latest commits, this no longer halts the page in demo mode and I've tested that the supervision level filter functionality is working as expected.

jessex:
Build is passing on Travis, merging this.

PR #492, opened 2020-09-28
Add Intercom support for Lantern dashboard (MO and PA only)

### LINKED ISSUE #452: Add Intercom support for Lantern dashboard
Before merging, we should ensure that @chrispmlee gets a chance to clone the branch and test locally.

We are trying to set up a universal doc center and support channel for Recidiviz products via Intercom. This will enable users to have a unified portal to contact us with questions / feedback and access Recidiviz product documentation. Lantern is the first test product for the integration.

Here are directions from the Intercom website to get started with the code necessary. The authentication should not be needed in order to see the instructions (Please let @chrispmlee know otherwise).
https://app.intercom.com/invites/6318f2cd-dcf0-478e-a074-65371ae5af53

We are interested in capturing the current State_ID (i.e. whether it's North Dakota, Washington, Idaho, etc.) so that we can appropriately show user's the correct documentation. This should be included as a custom attribute.

The portal should look something like this when active:
image

N/A

**Additional context**
@chrispmlee  would like to test Intercom before deploying 👍

jovergaag:
## Description of the change

Adds Intercom support for the MO and PA dashboards. The Intercom widget is present on the bottom right corner of the app for authenticated users on those dashboards. 

Functionality has been tested and the Intercom users config has been modified to include the custom attribute `state_code`.

Note: I will work with @chrispmlee on local testing to get QA/sign off from requester. 

## Type of change

## Related issues

> Closes #452 

### Development

### Code review


src/utils/initIntercomSettings.js
```diff
@@ -0,0 +1,37 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+// =============================================================================
+
+import { getUserStateCode } from "../utils/authentication/user"
+
+const APP_ID = "pqcz3ikn";
```
jessex, line 20: Is this sensitive / should it be in an environment variable / should we rotate it now that it has been checked in?

jessex (changes requested):
Looks good, @jovergaag! In local testing, though, I notice that the Intercom widget is still enabled even when I switch over to a non-Lantern state like ND. I think we will want to make sure that's _not_ the case.

src/utils/initIntercomSettings.js
jovergaag: Oh yes, I meant to add the environment variable that but in all the excitement forgot to. I'll do that now.

@chrispmlee and I had discussed that if a user logs in with a Lantern state email, they would have the widget available to them as they navigate between state dashboards (not sure if that is possible or not, maybe a CSG user?) If we want to be very specific about not having the widget available on ND despite the user's state code, then I can disable Intercom specifically on that page. Thoughts?

jessex: I think that makes sense for sure, but there are currently three groups of users to be aware of:
- Actual end-users who always have access to one state, e.g. internal users in MO or ND, and should be irrelevant to this discussion since they cannot switch states
- CSG users who have access to all of the _Lantern_ states, and because they can only switch states among Lantern states, they should always see the Intercom widget (and we'd want to be sure that the state code passed to Intercom was accurate both in their initial state they load into and the state that they switch to)
- Recidiviz users who have access to all states, and because they can switch between Lantern and non-Lantern states, if we wanted to be perfect, we would make sure that the widget only showed up for the Lantern states and went away when switching to non-Lantern states

So I would suggest doing the specific thing and not showing the widget on non-Lantern states. I believe Vlad had done this successfully, if I recall, on his Help Scout branch/PR.

jovergaag: That makes sense. I had thought CSG had access to the ND dashboard but it makes sense that they would not.

I've updated it so that the widget only appears on MO and PA, not on ND no matter which user is logged in. For Recidiviz users, the widget is disabled when the user selects ND on their profile page and is re-enabled when they select PA or MO. That should satisfy the use cases you outlined above.

I'm still waiting on support from Intercom to change the APP_ID. Once that is finished I'll put this up for re-review.

jovergaag:
Update on this. The Intercom has been incredibly slow responding to support by both @chrispmlee  and I. Changing the app id is not something we can do without support. @sychang confirms that this is not necessary for PA launch since there are only 10 DOC leadership and they all have direct email access to us. 

I will continue to keep on top of Intercom to help us change out app id.

jessex:
Yes their support team in general has been quite slow, disappointingly. But thanks for staying on top of this!

chrispmlee:
Yes still no response from Intercom, we have pinged them every day. I did local testing on my side, and it looks good to me (aside from APP_ID stuff) Will need to work with product a little bit more to figure out how we will use intercom's Help Center. Also submitted a ticket for design to help revamp and make it look ~sparkly~ for when we do have documents up on Help Center.

jovergaag:
@jessex I think we're finally ready! Moved the new app id to env files. The styling of the chat drawer has also been updated.

FYI - the app id in the .env.development.local and .env.development are for the test workspace, so messages generated from non production apps do not end up on the main workspace. Only .env.production has the production workspace id. 

Important note about the test workspace: the launcher icon at the bottom RH corner of the APP is an on-brand green. For the test workspace it is black. So if you see a jarring black icon, just know that in production it will nicer. Or you can try it with the production id locally and see the colors for yourself.

README.md
```diff
@@ -36,7 +36,7 @@ That's it! We suggest installing a linting package for your preferred code edito
 
 Second and last, set up your environment variables.
 
-For Recidiviz staff, download and unzip the `pulse_dashboard_env_vars.zip` from the shared 1Password vault and copy the files into the project directory. 
+For Recidiviz staff, download and unzip the `pulse_dashboard_env_vars.zip` from the shared 1Password vault and copy the files into the project directory.
```
jessex, line 39: Just want to make sure you've updated this in 1Password, @jovergaag!

jessex (approved):
Beautiful. Thanks so much for driving this to ground, @jovergaag!

README.md
jovergaag: I did indeed.

PR #493, opened 2020-09-29
Update PA risk charts with PA specific risk categories

### LINKED ISSUE #491: Update PA risk chart with PA-specific categories
PA has different risk level categories than MO. The risk level chart on the revocations matrix needs to be updated to reflect that. Charts that need to be updated
- Risk level chart
- Race chart
- Gender chart

The Risk Level column in the "Admitted individuals" list at the bottom should also reflect this state-specific mapping (show MEDIUM instead of MODERATE for PA).

**Additional context**

jovergaag:
## Description of the change

PA and MO have different risk categories. This changes the risk categories for PA in the Risk Level, Gender, and Race charts, as well as in the Case Table.

I also added some constant files to use constants instead of string state codes and Core/Lantern state arrays.

## Type of change

## Related issues

> Closes #491

### Development

### Code review


src/tenants.js
```diff
@@ -0,0 +1,50 @@
...
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+// =============================================================================
+
+import * as lantern from "./utils/lanternConstants";
+import * as core from "./utils/coreConstants";
+
+export default {
+  [lantern.MO]: {
+    "name": "Missouri",
+    "avaiableStateCodes": [lantern.MO],
```
jessex, line 24: Whoooooops, typo: should be `availableStateCodes`

src/utils/transforms/labels.js
```diff
@@ -18,32 +18,47 @@
...
-
-const riskLevelValuetoLabel = {
-  NOT_ASSESSED: 'Not Assessed',
-  LOW: 'Low',
-  MEDIUM: 'Moderate',
-  HIGH: 'High',
-  VERY_HIGH: 'Very high',
+]
+
+const riskLevelValueToLabelByStateCode = {
```
jessex, line 31: I think in the long-run we will want to declare state-specific manifests for these sorts of things, maybe even as an expansion of `tenants.js` or something like it, but I don't think it's a requirement just yet.

jessex (changes requested):
Thanks, @jovergaag! I dig the cleanup/rationalization here. Just a nit and an idle thought from me.

src/utils/transforms/labels.js
jovergaag: I totally agree. I was thinking of heading down that route in this PR but given the time pressure just wanted to make sure we can get this in before PA launch which *might?* be Thursday.

src/tenants.js
jovergaag: great catch.

jessex (approved):
🙌

PR #499, opened 2020-10-01
Production data bug - MO matrix technical revocation counts

### LINKED ISSUE #498: [Lantern] Revocations matrix showing doubled numbers
**Describe the bug**
Serena has identified a production issue with the MO dashboard where it appears we are double-counting numbers somewhere in the front end. For a 1 year window, technical revocation count is 2X in the dashboard, but the number we’re sending to the frontend is X.

This is the relevant query  in  recidiviz-123.dashboard_views.revocations_matrix_cells which shows the X that is getting sent to the frontend (look at first row):

SELECT *
FROM `recidiviz-123.dashboard_views.revocations_matrix_cells`
WHERE state_code = 'US_MO' AND violation_type = 'TECHNICAL' AND metric_period_months = 12
ORDER BY total_revocations DESC

**To Reproduce**
1. Go to Lantern dashboard in prod and see that the number in the 1-year technical cell is 2X and not X.

**Expected behavior**
Value should be X to match the output in our dashboard metrics file.

**Screenshots**
See above

 - OS [e.g. Windows 10]: All
 - Browser [e.g. chrome, edge]: All
 - Version [e.g. 22]: All

 - Device: [e.g. iPhone6]: 
 - OS: [e.g. iOS8.1]: 
 - Version [e.g. 22]: 

**Additional context**

jovergaag:
## Description of the change

Fixes a bug in production where revocations were being counted twice due to filtering logic not taking into account the new `supervisionLevel` rows in the MO data.

I agree with @jessex that this is not the most elegant solution, but the more elegant solution will take more time so lets get it in and I'll look at it more soon.

## Type of change

## Related issues

> Closes #498

### Development

### Code review

### DIFF
--- a/src/components/charts/new_revocations/helpers.js
+++ b/src/components/charts/new_revocations/helpers.js
@@ -46,7 +46,8 @@ const includesAllItemFirst = (items) => {
 export const applyTopLevelFilters = (filters) => (
   data,
   skippedFilters = [],
-  treatCategoryAllAsAbsent = false
+  treatCategoryAllAsAbsent = false,
+  applySupervisionLevel = false
 ) =>
   data.filter((item) => {
     if (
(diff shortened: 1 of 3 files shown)


jessex (changes requested):
I don't think applySupervisionLevel is supposed to be true for MO. The point of the fix is that we want only the supervision_level=ALL data points for MO. Setting that arg to true makes MO the same as PA, but we need them to be treated in the opposite fashion.

Interestingly, when I ran your branch locally, I noticed that the matrix itself appeared to have accurate counts but the other charts were still duplicated. I don't know how that is the case and I may be crazy, but that was my observation. I think you may want to pair up with Anna or Julia in the AM to make sure your implementation is getting the right counts for all charts.

jovergaag:
@jessex This is ready for review now. We were just passing the boolean in to the `applyTopLeveFilters` incorrectly. And the violations chart was broken because it doesn't have the `supervision_level` attribute. So I put in a fix for that.

You are right, the below the fold charts *looked correct, but in fact the data was all doubled but since they are percentages it was not easy to spot. That is fixed now as well.

jessex (approved):
We're good to go here from my own testing. Thanks so much for spearheading this, @jovergaag!

PR #515, opened 2020-10-06
Copy changes for PA

### LINKED ISSUE #495: PA Lantern dashboard copy changes
All mentions of "Officer" should be updated to "Agent"

All mentions of "Gender" should be updated to "Sex"

This applies to PA only. Each state moving forward will have some state-specific copy tweaks like this.

**Additional context**
This change should be updated prior to the beta.

jovergaag:
## Description of the change

There are several copy changes to be applied to the PA dashboard. The full list is [here](https://docs.google.com/document/d/1o1mJD0Cbb8B9M-Dt10uWkbPbhzbnSNN3XO9_gQ4bEmQ/edit#)

I ended up using the I18n package to use for our tenant copy translations. This will help us keep all of these copy translations in one place, to make updates down the road, and to add new tenants in the future as well. I chose to go with I18n instead of just a static object lookup because I18n can be quite powerful, it can easily manage pluralize/singularize, as well as set the "locale" based on the route so that all of this just magically happens depending on what page the user is on. It has a fairly simple config right now, but we can expand on it as we go.

I also ended up moving some of the tenant specific constants into a `tenants/utils` directory to try to keep those all centralized and clean. @Wmaileq may have also been looking into centralizing these so let me know if you ended up taking a different approach and we can adjust.

NOTE: I am fully aware that none of this is tested. I chose to prioritize getting this in, along with several other FE priorities for the PA launch. I will make an issue to write tests for these once I am no longer blocking the launch.

## Type of change

## Related issues

> Closes #495 

### Development

### Code review


package.json
```diff
@@ -54,6 +54,7 @@
     "helmet": "^3.21.2",
     "highlight.js": "^9.15.6",
     "http-errors": "~1.6.2",
+    "i18n": "^0.13.2",
```
jovergaag, line 57: @jessex I saw on a previous PR that you had made a comment that a package that was added didn't pass the security checks. Where can I find that check so I can run new packages through it to verify?

src/components/charts/new_revocations/RevocationsByDistrict/helpers.js
```diff
@@ -75,13 +76,21 @@ export const sumCounts = (key, data) =>
...
-      { label: "Revocation count", value: "counts" },
-      { label: "Percent revoked of standing population", value: "rates" },
-      { label: "Percent revoked of exits", value: "exits" },
-    ]
-  : [
-      { label: "Revocation count", value: "counts" },
-      { label: "Percent revoked of standing population", value: "rates" },
-    ];
+export const modeButtons = () => {
+  return flags.enableRevocationRateByExit
```
jovergaag, line 80: It seems to me like this concept of `RevocationsByExit` no longer exists. I need to verify that, but believe that this is another area for refactor/cleanup.

package.json
jessex: We were talking about [Help Scout](helpscout.com) the vendor, not a particular dependency/library. Each vendor we want to use needs to go through a security assessment and we were not comfortable moving forward with Help Scout.

As it pertains to specific dependencies/libraries, we rely on Github's automated CVE scanning tool to tell us if we need to upgrade the version of a particular dependency, so it's important to generally rely on the most recent stable version of anything we add to minimize the chance that we'll get an alert immediately upon merge.

src/assets/styles/spec/components/dashboard.scss
```diff
@@ -316,23 +316,6 @@
       color: inherit;
     }
   }
-
-  .separation-column {
-    display: none;
-  }
-}
-
-.separate-filter-title {
```
jessex, line 325: Were these CSS classes no longer necessary? Not even `.long-title`?

src/components/charts/new_revocations/RevocationsByDistrict/helpers.js
jessex: No, it does exist. There is calculation work blocking our being able to enable the feature flag. @jjdressel11 can share more about timeline.

src/components/charts/new_revocations/RevocationsByRiskLevel/RevocationsByRiskLevel.js
```diff
@@ -195,6 +195,7 @@ const RevocationsByRiskLevel = ({
     />
   );
 
+  // TODO do revocationsRateByExit still exist?
```
jessex, line 198: Yep! See my reply above.

jessex (changes requested):
Great work, @jovergaag! Just some minor remarks and questions for you. Love this use of `i18n`.

src/utils/charts/info.js
```diff
@@ -14,304 +14,313 @@
...
+    revocationsOverTime: [
+      {
+        header: "What this chart shows",
+        body: `This chart displays the total number of people who were admitted to prison during
+  each month due to revocation by the parole board or the court.`,
+      },
+      {
+        header: "Who is included",
+        body: `This chart includes only people who were revoked to prison. Counts do not include (1)
+  people who were revoked to termination or to a new form of supervision (i.e., SIS to SES);
```
jessex, line 31: @jovergaag I think @sychang may have missed these. SIS and SES are MO-specific and irrelevant to PA. Also, a lot of the language in this particular file appears to have not been adapted to PA-specific needs (e.g. still generally referring to revocations as opposed to admissions to SCI). I think Product may want to take another pass through the additional info blurbs to make sure they're ready to go.

jessex (commented):
@jovergaag I think the code looks great and I double-checked that your styling change in the case table column headers looks reasonable. However, I think some of the copy in the additional info text across the page may still need to be updated. See my comment below.

jessex (approved):
Good to go! 🚀

PR #519, opened 2020-10-06
Bug - Supervision Level filters

### LINKED ISSUE #514: Fix supervision level double-counting in PA
The issue addressed last week to resolve MO's double counting issue is present for PA. The bubble chart should not be double-counting when "all" is selected as a supervision level.

MO issue: 
https://app.zenhub.com/workspaces/-lantern--5f2cdf04d4d9670013da86ed/issues/recidiviz/pulse-dashboard/498

**Additional context**

jovergaag:
## Description of the change

On the PA dashboard, when the `supervisionLevel = 'ALL'` filter was applied, revocations were being counted twice because all rows were being counted instead of only the 'ALL' row. This work changes how that filter is applied. It includes tests which hopefully more clearly lay out the expectation of what the filter result should be based on input data and filter variations.

## Type of change

## Related issues

> Closes #514

### Development

### Code review


src/views/tenants/us_mo/community/Revocations.js
```diff
@@ -112,6 +113,7 @@ const Revocations = () => {
       : {}),
     reportedViolations: "",
     violationType: "",
+    supervisionLevel: SUPERVISION_LEVELS[0].value,
```
jessex, line 116: If I'm not mistaken, this is the linchpin of the whole change, right? You're hard-coding the filter set for MO to include `supervision_level=ALL` so it's always going to disregard every row where that's not the case? If so, I'm sad that I didn't realize how simple the fix was _supposed_ to be and am happy that you did!

src/components/charts/new_revocations/__tests__/helpers.test.js
```diff
@@ -0,0 +1,124 @@
...
+
+    it("does not double count the 'ALL' item", () => {
+      expect(filtered).not.toContain(data[0]);
+    });
+  });
+
+  describe("with supervision_level = 'ALL' filter applied (default)", () => {
+    beforeEach(() => {
+      filters = { supervisionLevel: "ALL" };
+      filtered = applyTopLevelFilters(filters, true)(data);
```
jessex, line 98: You should remove the `true` arg now, right? We're back to only a single argument into that call?

jessex: Same goes for the other `applyTopLevelFilters` calls in these tests.

jessex (changes requested):
Very very good simplification, @jovergaag! One nit and one question.

src/views/tenants/us_mo/community/Revocations.js
jovergaag: Yes that was part of the secret sauce. That was the route I was going down that afternoon...but I was missing the other important piece. That piece was removing `!isAllItem(filters.supervisionLevel)` on line 96 of the helpers. That line was removing the 'ALL' row from the data before it got to the `nullSafeComparison` which is what is ultimately doing the filtering work. 

So, this particular filter was actually working differently than we thought it was. with `supervisionLevel = 'ALL'` filter, it was returning all rows except the 'ALL' row and summing them which is the opposite of what our discussion on Slack suggested was the desired result. I'm pretty sure the other filters are still doing filtering that way, so I hope to look at them more closely later this week.

jessex (approved):

PR #530, opened 2020-10-08
Feature: added pre push testing

Wmaileq:
## Description of the change

- Pre-push testing added

## Type of change

## Related issues

No related issues

### Development

### Code review


package.json
```diff
@@ -111,7 +111,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "lint-staged"
+      "pre-commit": "lint-staged && yarn test"
```
jovergaag, line 114: I'm a little hesitant to add this hook. For two reasons:

1. As our test coverage continues to grow for the repo the test suite will take longer to run which means longer to commit.
2. I often will break a large task up into smaller chunks and commit throughout, even if the tests aren't passing yet. This wouldn't allow me to do this.

I would love to hear your thoughts on this and why you want to add it. Maybe I can be convinced!

jovergaag (changes requested):
Comments in line.

package.json
Wmaileq: For the first case I see the solution in running tests only for changed files, is this reasonable?

For the second case: I think any committed code should never fail tests
Actually is this the case?
When you add new code it could be not covered with tests but in this case pre-commit hook will not fail and you can add tests in next commit.
But I can't imagine situation when you need to commit code that FAILS tests, I think test fixes should be included.

jovergaag: What if you changed this to a pre-push hook instead of pre-commit?

If I'm working on a larger chunk of work, I will often make many smaller commits so that I can revert that chunk if it ends up not being applicable. Often times the commits will not be passing tests, or, if doing TDD and I have written one or more tests for a feature that is not finished yet the test(s) will not be passing until the end.

I think a pre-commit hook isn't flexible to support different people's personal workflow styles.

But I would be open to the pre-push test hook!

jovergaag (changes requested):
See my note on previous comments.

package.json
Wmaileq: @jovergaag This sounds logical! I think we have come to a common opinion.

package.json
```diff
@@ -111,7 +111,8 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "lint-staged"
+      "pre-commit": "lint-staged",
+      "pre-push": "yarn test --only-changed"
```
jovergaag, line 115: I changed a test to ensure that it failed, committed it, then pushed it up to this branch. The push was successful. I removed the `--only-changed` flag and did the same thing and it triggered the tests, which failed, and push failed. I think we should remove the `-only-changed` flag here. 

I think the 'changed' files are ones which are staged, so after a commit there is no longer anything staged.

jovergaag (changes requested):

package.json
Wmaileq: My fault, fixed

jovergaag (approved):
👍

PR #555, opened 2020-10-20
Normalize officer recommendation in case table

### LINKED ISSUE #547: Fix string normalization for officer_recommendation string in Lantern dashboard
Currently the Lantern frontend uses a "district name normalizer" on the `officer_recommendation` field that comes from the `revocations_matrix_filtered_caseload` metric file, which translates strings like "TREATMENT_IN_PRISON" to "Treatment in prison" for display in the dashboard (e.g. it removes punctuation and lowercase all except first character). The normalization is not what we want for certain strings:

Current strings that are incorrectly normalized:
"CODS" -> "Cods" (only used in MO)
"PLACEMENT_IN_DOC_FACILITY" -> "Placement in doc facility" (value added in #4325, will not exist in staging data until deployed and export rerun)

Desired normalization:
"CODS" -> "CODS" (keep capitalized)
"PLACEMENT_IN_DOC_FACILITY" -> "Placement in DOC facility"

**Additional context**
You can query for full list of officer_recommendation values present in prod with this query:
```
SELECT DISTINCT officer_recommendation 
FROM `recidiviz-123.dashboard_views.revocations_matrix_filtered_caseload`
```

jovergaag:
There are two special cases for how we want to normalize the officer recommendation string to be displayed in the case table (and associated export csv).

This adds that normalization, along with some organization of the `CaseTable` helper methods (which eventually will be removed to a reducer) and tests. I hope to move this component (not sure why, but it is the chosen one) to the new reducer and styled component architecture next week, pulling this formatting logic out of the component is a step in that direction.

Was able to quiet a few React key warnings as well.

## Description of the change

> Description here

## Type of change

## Related issues

> Closes #547 

### Development

### Code review


src/components/charts/new_revocations/CaseTable/helpers.js
```diff
@@ -0,0 +1,92 @@
...
+import { translate } from "../../../../views/tenants/utils/i18nSettings";
+import { COLORS } from "../../../../assets/scripts/constants/colors";
+
+const nullSafeLabel = (label) => label || "Unknown";
+
+const normalizeOfficerRecommendation = (value) => {
+  switch (value) {
+    case "CODS":
+      return value;
+    case "PLACEMENT_IN_DOC_FACILITY":
```
jessex, line 31: Is this case because `humanReadableTitleCase(PLACEMENT_IN_DOC_FACILITY)` produces `Placement In Doc Facility` instead of the all-caps DOC? If so, bummer.

src/setupTests.js
```diff
@@ -16,3 +16,12 @@
...
 import "@testing-library/jest-dom/extend-expect";
+
+import {
+  setTranslateLocale,
+  initI18n,
+} from "./views/tenants/utils/i18nSettings";
+import * as lanternTenant from "./views/tenants/utils/lanternTenants";
+
+initI18n();
+setTranslateLocale(lanternTenant.PA);
```
jessex, line 27: I don't know if this would be a tad much, but did you consider creating a test tenant instead of referencing a real live production tenant in the test code? I don't think you necessarily need to, but it means that potentially unrelated config changes for PA might lead to churn in test code.

jessex (approved):
Great work, @jovergaag! Two minor questions for you.

src/components/charts/new_revocations/CaseTable/helpers.js
jovergaag: Yes, that is exactly the issue. They want to maintain the all caps DOC

src/setupTests.js
jovergaag: I think we actually want to keep it pointed at an actual tenant. That way we can be testing these copy translations simultaneously with the helper function. If we end up changing the translation - if we are more fully covered, it will be super apparent where that translation is being used because theoretically we would have those tests failing. 

Is that what you were getting at?

jessex: Mm, I see what you mind. Makes sense!

ageiduschek:
Thank you!!

Issue #25, opened 2019-07-24
Add comprehensive unit testing


jessex:
There is a single test file, `App.test.js`, but it is commented out. We need unit tests for the following:
* Views
* Components
* Basic logic under `src/`
* Backend API logic

jessex:
As part of this, we want to reenable/fix the Coveralls code coverage integration and ensure we get a coverage report with each local build and on each pull request.

jessex:
@Wmaileq, we'd like to continue to make improvements in our test coverage. Here are a few things to tackle next:
- Add unit tests for the backend, i.e. the code under `/server`
- Enable Coveralls for the repository so we get test coverage reports with our build
- Fill in some missing test coverage on the frontend, including `/src/utils/authentication`, `/src/auth0/hooks`, and `/src/auth0/rules`

We can do those three first with separate pull requests. If we make it through those, my next target for improved testing would be `/src/components` but let's see how quickly those first three go.

jessex:
@Wmaileq, I think at this point, you should take a look at the Coveralls report and target any areas that can be high-value additions of test coverage, i.e. high improvements in coverage for relatively less effort.

jovergaag:
@Wmaileq Let's also re-enable the `App.test.js`

jovergaag:
@Wmaileq Since we have so much code churning right now, please prioritize this issue over your other refactor issues (other than the open PRs) for now. We are launching the product to the PA folks on Thursday so I think we should focus on adding test coverage instead of new features/refactors for the next few days to allow any bug fixes and fast follows from that launch to roll through quickly.

`TenantRoutes` would be a great place to start.

Thanks you!

Wmaileq:
Working on `src/react-auth0-spa.js`

ageiduschek:
Closing this! Going to open new issues as test needs arise

PR #575, opened 2020-10-26
App tests

Wmaileq:
## Description of the change

- App.js tests
- Coverage increased from 26.52% to 35.669%

## Type of change

## Related issues

> Part of #25 

### Development

### Code review


src/App.test.js
```diff
@@ -15,18 +15,106 @@
...
+      "/community/revocations"
+    );
+    const user = {
+      "https://dashboard.recidiviz.org/app_metadata": {
+        state_code: MO,
+      },
+    };
+    useAuth0.mockReturnValue({
+      user,
+      isAuthenticated: true,
```
jovergaag, line 62: Can you add a test case for when `isAuthenticated` is false and when `loading` is true. We just want to be sure that nothing is rendered before the user is authenticated. We could also probably want a test to ensure that the user is redirected to the login page.

If this is too difficult to write tests for in demo mode, let me know and I can take over those last couple.

jovergaag (changes requested):
Just a couple of non-happy-path tests that we should look into.

Thanks!

src/App.test.js
Wmaileq: Tests added!
Thinking about what we need here (integrational or unit tests) came to the idea that unit tests here would just cover internal react-router logic, what we don't really need.
So I've covered `App` in a bundle with the `TenantRoutes` and the `StateCodeProvider` components that resulted in having almost 100% coverage on the `TenantRoutes`.
Despite this, I think we need to do unit tests for `TenantRoutes` and `StateCodeProvider` anyways.

jovergaag (changes requested):
Just getting clarification on `TenantRoutes` test.

src/components/TenantRoutes.js
```diff
@@ -56,21 +56,15 @@ const TenantRoutes = ({ children }) => {
     return null;
   }
 
-  let element = null;
-
-  Children.forEach(children, (child) => {
+  return Children.toArray(children).reduce((node, child) => {
```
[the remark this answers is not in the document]
Wmaileq: @jovergaag 
image
TenantRoutes completely covered by this PR :)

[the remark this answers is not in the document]
jovergaag: Oh! I haven't actually logged into Coveralls, or worked with it before. I'm a little surprised that `TenantRoutes` is considered covered. 

Let's go ahead and merge this and you can continue adding tests for the uncovered files, then we can go back and add unit tests for `TenantRoutes` later. We will be continuing to refactor some of the other components so it feels more important to have them covered as we refactor. 

Eventually I would like to change the routes to add `:state_code` to the route and maybe even the filter state - at that point the `TenantRoute` tests will be more important but that is much further down the road. I actually don't think `TenantRoutes` is a great implementation and would like to find a different solution overall...

__helpers__/mockWithTestId.js
```diff
@@ -0,0 +1,5 @@
+import React from "react";
```
jovergaag, line 1: Great idea.

jovergaag (approved):
Thank you!!

PR #582, opened 2020-10-27
Security vulnerability fix and bug fix

### LINKED ISSUE #561: Security advisory: upgrade create-react-app to resolve security alert in object-path
Upgrading create-react-app should upgrade object-path to 0.11.5 and resolve the security alert.

Relevant info: https://github.com/facebook/create-react-app/issues/9842

**Additional context**

jovergaag:
## Description of the change

Addresses 3 issues:
1. Security vulnerability in react-scripts 3.4.3. Upgraded the module to [3.4.4](https://github.com/facebook/create-react-app/releases/tag/v3.4.4). I tried upgrading to `react-scripts@latest` which was 4.0.0 but it was not working with our version of node. So until we upgrade node we can stick with 3.4.4. #561 
2. While on the node issue I remembered that @macfarlandian had had trouble installing packages a couple weeks back due to a node version, changing node versions resolved his issue. I added a `.nvmrc` to help avoid this hiccup in the future, and added it to the README.
3. QA today found a MO data bug, it came from the Revocations refactor that was merged yesterday. This adds a fix for that along with a test.

I'm hoping that this will be the last code changes before our final deploy to Production for the PA launch. @sychang has finished a round of QA this morning which uncovered #3 but nothing else. We will both do another round this afternoon after this is deployed to staging.

## Type of change

## Related issues

> Closes #561

### Development

### Code review


src/views/tenants/constants/__tests__/filterOptions.test.js
```diff
@@ -0,0 +1,481 @@
...
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+// =======
+
+import filterOptionsMap from "../filterOptions";
+
+describe("filterOptionsMap", () => {
+  describe("when the stateCode is us_mo", () => {
+    it("returns the correct filter options", () => {
+      const result = filterOptionsMap.us_mo;
+      const expected = {
```
macfarlandian, line 24: i wouldn't block approval on this but it seems like there is a fair amount of repetition across these two giant objects ... I wonder whether breaking this up into smaller tests would make it easier to digest or lead to more granular and useful failures if something breaks, and let you factor out the subsets that overlap?

macfarlandian (approved):
🙌 good call specifying the node version

README.md
```diff
@@ -20,13 +20,17 @@ Bringing criminal justice analysis to decision makers to help reduce incarcerati
 
    `git clone git@github.com:Recidiviz/pulse-dashboard.git`
 
-1. Install Yarn package manager:
+2. Ensure you are using the correct version of Node (if you don't use NVM, just check the .nvmrc file and ensure you are using that version).
```
jessex, line 23: Just a protip in case you're not seen this before: in Markdown, you can number things as 1, 1, 1, 1 instead of 1, 2, 3, 4 and it will get rendered correctly as the latter. This is useful because it means that if you later add more items to the list or remove existing ones, you won't have to change all of the numbers to ensure the list still renders sensibly.

jessex (approved):
Thanks, @jovergaag!

src/views/tenants/constants/__tests__/filterOptions.test.js
jovergaag: @macfarlandian I broke it down and made it clear which options are the same across tenants. I would be hesitant to break down the objects any more and generalize any more because we have had some trouble with the filters due to these filter options and the tiny differences between them for tenants, so I want to leave no room for the imagination about what these are supposed to look like. LMK what you think!

jovergaag:
@jessex @macfarlandian We ended up finding one more small IE UI issue which I decided to add on here since Travis takes so long to build. If you could give this another look I would appreciate it!

jessex (approved):
Looks good!

src/views/tenants/constants/__tests__/filterOptions.test.js
macfarlandian: fantastic!

macfarlandian (approved):
🎉

PR #583, opened 2020-10-28
Tests/ Layouts and Intercom

Wmaileq:
## Description of the change

- Intercom slightly refactored from common function to hook and now should be used only on pages where it should be visible, instead of toggling on every page.
- Layout tests added
- Intercom utils\hook tests added

## Type of change

## Related issues

> Part of #25

### Development

### Code review


src/utils/initIntercomSettings.js
```diff
@@ -15,32 +15,14 @@
...
-
 const APP_ID = process.env.REACT_APP_INTERCOM_APP_ID;
 
-export function initIntercomSettings() {
+function initIntercomSettings() {
   window.intercomSettings = {
     app_id: APP_ID,
   };
 
-  window.Intercom("boot", window.IntercomSettings);
```
Wmaileq, line 27: @jovergaag Was it working? 
As I have no access I can't check, but here we set settings to `window.intercomSettings` and then boot Intercom with `window.IntercomSettings` which should be `undefined`.
Is it an error or have I missed something?

jovergaag: When I add `  console.log(window.intercomSettings)` to line 26 I see an object with an app_id. This seems to be working as expected.

package.json
```diff
@@ -133,14 +134,16 @@
   },
   "jest": {
     "setupFilesAfterEnv": [
-      "./src/setupTests.js"
+      "./src/setupTests.js",
+      "@testing-library/react-hooks/dont-cleanup-after-each.js"
```
jovergaag, line 138: I'm curious why you added this. Did you have tests failing? Or just for performance?

src/components/layouts/__tests__/CoreLayout.test.js
```diff
@@ -0,0 +1,62 @@
...
+    const { container } = render(<CoreLayout>{mockChildren}</CoreLayout>);
+
+    expect(useSideBar).toHaveBeenCalled();
+    expect(container.firstChild.className).not.toContain("is-collapsed");
+  });
+
+  it("should show right pathname in TopBarTitle", () => {
+    const { getByText } = render(<CoreLayout>{mockChildren}</CoreLayout>);
+
+    expect(getByText("Some > Nested > Pathname")).toBeInTheDocument();
```
jovergaag, line 60: 👍

src/utils/initIntercomSettings.js
jovergaag: I realized I didn't answer your question - yes it is working.

src/utils/initIntercomSettings.js
```diff
@@ -15,32 +15,14 @@
...
-    email: user.email,
-    user_id: user.sub,
-    hide_default_launcher: false,
-  };
-
-  window.Intercom("update", intercomSettings);
+  window.Intercom("boot", window.intercomSettings);
 }
 
-export function disableIntercomLauncher() {
```
jovergaag, line 44: We need to be able to disable the Intercom launcher for the ND page. Intercom should only be available on the LanternLayout, not on the CoreLayout. Since it is initialized in `App.js`, the launcher is enabled but it just hasn't been updated with the user settings in `useIntercom` when on the ND page. I think the best thing to do is set `hide_default_launcher: true` on line 22.5 here to disable it by default. I tested that locally and it worked when switching back and forth between tenants.

The Intercom launcher is an `<iframe>` so it is going to be hard to write a test for it not having it in your browser DOM for reference.  I will in the meantime figure out if we can send you the app_id for Intercom so you can see what is happening.

jovergaag (changes requested):
One small change to disable Intercom on ND

src/utils/initIntercomSettings.js
Wmaileq: @jovergaag yeah, `console.log(window.intercomSettings)` working now :) 
But I am curious about `master`, is intercom working on master now? It seems to be an error here, which I was trying to point out. 

In master we assign `appId` to `window.intercomSettings`, and then read them with `window.Intercom("boot", window.IntercomSettings`). (Upper and lower cased **I**)

package.json
Wmaileq: https://github.com/Recidiviz/pulse-dashboard/pull/583/files#diff-b31924d36fcfd13a11cf4636cec0a249ff3f5fa63221a1ebef9be6dfa3acca76R29

Hooks are unmounting every test (after every `it`), without this setup, what is not what we expect, in my opinion. 
In tests I want to take control of mounting, rerender, and unmounting of hooks to get conditions closer to the real.

src/utils/initIntercomSettings.js
Wmaileq: >  I think the best thing to do is set hide_default_launcher: true on line 22.5 here to disable it by default.

Yes! I think this is the best solution. Now it should work as expected, thanks for the hint!

Wmaileq:
Fixed the issue and rebased to the masters current head.

jovergaag (approved):
Awesome. Thank you!

PR #601, opened 2020-11-03
Fix vertical alignment of filter chip - label and delete button

### LINKED ISSUE #595: Filter chips are vertically misaligned
The text in the filter chips in the Lantern dashboard (both MO and PA) do not appear to be vertically centered - text looks slightly higher than center in these examples:

Note, in this second example, the vertical offset changes when the page scrolled all the way to the top. When I scroll down and the additional filter bar changes orientation, the text appears more center-aligned (the "X" appears slightly closer to the top, but is less severe than the above screenshots):

**Additional context**
I'm think I started noticing this maybe 10/27 or 10/28 before training. It didn't appear like that during other QA rounds, otherwise I'm pretty sure I would have noticed.

This is not beta exit blocking, and can be prioritized alongside any UI / tech debt tasks. Placing in 11/2 for discussion purposes.

jovergaag:
## Description of the change

The filter chips were vertically misaligned. This was the case in the 'Additional Filters' filter bar as well as the filter buttons in the revocations by district chart.

The chips after the fix:
Screen Shot 2020-11-03 at 1 16 40 PM
Screen Shot 2020-11-03 at 1 41 31 PM
Screen Shot 2020-11-03 at 1 41 36 PM

## Type of change

## Related issues

> Closes #595

### Development

### Code review

### DIFF
--- a/src/components/charts/new_revocations/ToggleBar/ViolationFilter.js
+++ b/src/components/charts/new_revocations/ToggleBar/ViolationFilter.js
@@ -17,6 +17,7 @@
 
 import React, { useMemo } from "react";
 import PropTypes from "prop-types";
+import cn from "classnames";
 
 import FilterField from "./FilterField";
 import Chip from "../Chip";
@@ -25,8 +26,10 @@ import {
   matrixViolationTypeToLabel,
   pluralize,
 } from "../../../../utils/transforms/labels";
+import { usePageState } from "../../../../contexts/PageContext";
 
 const ViolationFilter = ({ reportedViolations, violationType, onClick }) => {
+  const { isTopBarShrinking } = usePageState();
   const formattedMatrixFilters = useMemo(() => {
     const parts = [];
     if (violationType) {
@@ -39,7 +42,6 @@ const ViolationFilter = ({ reportedViolations, violationType, onClick }) => {
     }
     return parts.join(", ");
   }, [reportedViolations, violationType]);
-
   if (!formattedMatrixFilters) return null;
 
   return (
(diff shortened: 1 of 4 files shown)


jovergaag:
@sychang "am i ok"? :)

sychang:
You are so ok!

src/components/charts/new_revocations/ToggleBar/ViolationFilter.js
```diff
@@ -50,6 +52,7 @@ const ViolationFilter = ({ reportedViolations, violationType, onClick }) => {
       >
         <Chip
           label={formattedMatrixFilters}
+          className={cn({ "Chip--shrinking": isTopBarShrinking })}
```
macfarlandian, line 55: what's the reasoning behind moving this logic out here? it kind of seems to me that it was better off encapsulated within the `Chip` component rather than the parent having to know it needs to apply this classname. Especially since `Chip` is still conditionally applying other classnames to itself based on props

macfarlandian (changes requested):

src/components/charts/new_revocations/ToggleBar/ViolationFilter.js
jovergaag: There is only one Chip on the page (out of 15ish) that shrinks based on the scroll of the page. Prior to this PR the other chips were shrinking but that was a bug. 

Other options are to pass a boolean into Chip that specifies whether a particular instance is shrinkable. Or shrink the Chip when it is nested under `.FilterField--shrink`. I'm not a fan of the second option. I'll push up a commit to see what the first option looks like.

jovergaag: @macfarlandian That is much better, good suggestion. Ready for another look.

macfarlandian (approved):
🎸 🎶

PR #603, opened 2020-11-04
Agent chart added #490

### LINKED ISSUE #490: Front-end for Agent chart
We would like DOC leaders to be able to easily identify anomalies and standout agents to identify and understand anomaly agent behavior and properly allocate resources. The agency can work with individual districts and/or agents to address anomalies or set new agency policy to correct imbalances.

This will be a new middle chart in the Lantern dash:
image

**Additional context**

### LINKED ISSUE #540: [Tech-debt] ExportMenu should accept only datasets and labels fields, rather than the whole chart object
ExportMenu component should not accept chart node, it should accept only datasets and labels fields.

Also, please rename `TenantExportMenu` to `ChartCardExportMenu`. 

**Additional context**

Wmaileq:
## Description of the change

- RevocationCharts BEM implemented
- Added some translations "revoked", "admissions"
- Added Agent Chart

## Type of change

## Related issues

> Closes #490 
> Closes #540 

### Development

### Code review


src/components/charts/new_revocations/RevocationsByDistrict/RevocationsByDistrict.js
```diff
@@ -20,63 +20,70 @@ import PropTypes from "prop-types";
...
-    modes={
-      flags.enableRevocationRateByExit
-        ? ["counts", "rates", "exits"]
-        : ["counts", "rates"]
-    }
-    defaultMode="counts"
-  />
-);
+}) => {
+  const chartTitle = `${translate("Revocations")} by district`;
```
jovergaag, line 36: Can you do the same for the Gender, Race, and Risk Level charts please. Somehow we missed this translation in the refactor.

src/components/charts/new_revocations/RevocationCharts/RevocationCharts.js
```diff
@@ -33,33 +37,39 @@ const RevocationCharts = ({
...
+    conditionallyHide(selectedChart, "Gender", genderChart),
+    conditionallyHide(selectedChart, "Race", raceChart),
+    conditionallyHide(selectedChart, "District", districtChart),
   ];
 
   const renderSelectedChartSingularLoad = () => {
     switch (selectedChart) {
       case "Risk level":
         return riskLevelChart;
+      case "Agent":
```
jovergaag, line 64: The button for this chart in the labels menu should translate for the different tenants - Officer (MO)/Agent (PA)

jovergaag (changes requested):
This is looking great! When we were looking through this we were able to flesh out a few more requirements. I made comments below.

Wmaileq:
@jovergaag fixed issues.
Also toggled treatCategoryAllAsAbsent to true.
Curious whether all other charts should be toggled in the same way or am I missing something?

src/components/charts/new_revocations/RevocationCharts/RevocationCharts.js
```diff
@@ -82,29 +92,30 @@ const RevocationCharts = ({
...
         {CHARTS.map((chart) => (
-          <div key={chart}>
+          <div className="RevocationCharts__label" key={chart}>
             <button
               type="button"
-              className={`chart-type-label ${
-                selectedChart === chart ? "selected" : ""
-              }`}
+              className={cn("RevocationCharts__button", {
+                "RevocationCharts__button--selected": selectedChart === chart,
```
jovergaag, line 102: We are not seeing the chart in PA - in either live data or demo mode.

src/components/charts/new_revocations/RevocationCharts/RevocationCharts.js
```diff
@@ -33,33 +37,39 @@ const RevocationCharts = ({
...
   );
 
   const renderSelectedChartSimultaneousLoad = () => [
-    conditionallyHide(selectedChart, "Risk level", riskLevelChart, 0),
-    conditionallyHide(selectedChart, "Violation", violationChart, 1),
-    conditionallyHide(selectedChart, "Gender", genderChart, 2),
-    conditionallyHide(selectedChart, "Race", raceChart, 3),
-    conditionallyHide(selectedChart, "District", districtChart, 4),
+    conditionallyHide(selectedChart, "Risk level", riskLevelChart),
+    conditionallyHide(selectedChart, translate("Officer"), officerChart),
```
jovergaag, line 53: I don't think we need to translate this here - the copy translation is done on line 106. Maybe this is why the agent chart isn't rendering? Same goes for line 64. Sorry - I may have led you astray on this in my original review.

src/components/charts/new_revocations/RevocationsByOfficer/RevocationsByOfficer.js
```diff
@@ -0,0 +1,94 @@
...
+  dataFilter,
+  filterStates,
+  stateCode,
+  timeDescription,
+}) => {
+  const chartTitle = `${translate("Revocations")} by ${translate("officer")}`;
+
+  return (
+    <RevocationsByDimension
+      chartId="revocationsByAgent"
```
jovergaag, line 39: This should be `revocationsByOfficer`

src/components/charts/new_revocations/RevocationsByOfficer/createGenerateChartData.js
```diff
@@ -0,0 +1,145 @@
...
+      rate: calculateRate(dataPoint.count, dataPoint[fieldName]),
+    })),
+    orderBy(["rate"], ["desc"]),
+    (dataset) => dataset.slice(0, MAX_OFFICERS_COUNT)
+  )(apiData);
+
+  const dataPoints = map((item) => item.rate.toFixed(2), filteredData);
+
+  const labels = map("officer", filteredData);
+  const denominators = map("supervision_count", filteredData);
```
jovergaag, line 67: We are having a bug in the counts such that most of the officers are showing 100% - with a multiple of 8 in the denominator. The chart with demo data looks fine so this is almost certainly a backend bug. I'm going to look into the metric views and see if I can figure out what is going on, unfortunately that is lower priority than the IE11 performance issue so it might be a few days.

jovergaag (changes requested):
There are a few things to continue to look into.
* PA chart
* translations
* I'm going to look into the percent data to figure out why we're seeing strange results on the percent chart
* Export - this is something I didn't comment on below. We will want to export ALL of the data that matches the filter instead of truncating it to 50. Also, can you please make the data in the export rows instead of columns for officer names? That will make it much easier to read.

Re: `treatCategoryAllAsAbsent` - no we should not set that to true for this chart (yet). That means that there are no 'ALL' rows in the data coming from the API and so the data passes through the filters differently. For this view we do have the 'ALL' rows right now. However, this metric file is over 4 times larger than any of the other files and running this locally the performance impact is significant. I think we will have to trim down the metric file, and that might mean removing the ALL rows. But for now, do not include that boolean.

Wmaileq:
@jovergaag All done except 

> Also, can you please make the data in the export rows instead of columns for officer names? That will make it much easier to read.

I think maybe it is not the last chart where we should rotate, and the solution is the `ExportMenu` prop. It is not 5 minutes task so it would be easier to make an extra issue for this.

jovergaag:
@Wmaileq This is no longer rendering any data in the chart - even in demo mode. Maybe due to the merge master?

The data won't be ready to support this for another, so there is no rush to get this in. Considering that, I think it is worth adding the ability to `rotateData` within the `ExportMenu` component. If you want to do that in a separate PR to avoid this PR getting too big that is fine.

Wmaileq:
@jovergaag for me it is working, but you need to specify District. Because flag `treatCategoryAllAsAbsent` is disabled.

jovergaag (commented):
This looks great for now. Let's put it on hold while we sort out the backend.

I chatted with the product manager and she agrees that it is best to pivot the axis of the data export for most of the charts. I'll create a new issue for this so we can get it merge and ready for when this is ready to go.

jovergaag:
@Wmaileq Sorry for taking so long to get back to you on this. I was on vacation on Friday.

I just pushed up a commit that lets us see data in the officer chart. We were filtering for `district != 'ALL'` in `createGenerateChartData`. Removing that fixed the issue on live data immediately, and I just updated the fixture file so you can see a better set of data on the default filters.

jovergaag (approved):
Great! This will be followed up by a few other tasks to address remaining issues.

PR #612, opened 2020-11-11
Abortable useChartData promise and concurrent requests handling

### LINKED ISSUE #564: [Tech-debt] Do not fetch revocation_matrix_cells twice
We are currently fetching `revocation_matrix_cells` twice, once for the matrix chart and once for the district filters. Instead, hold this data in a context and allow both components to access it.

**Additional context**

### LINKED ISSUE #606: Memory leak - cleanup in useEffect, check props passed to children, 'use strict' mode?
Nearly all of our users on the PA dashboard use IE11. They have been reporting difficult sessions and errors loading the charts after using the app for a while. I have spent the day today running some performance analysis on IE11 and Chrome. With the optimizations in the optimization PR ([#486](https://github.com/Recidiviz/pulse-dashboard/pull/486) the memory usage in Chrome when loading the MO dashboard is approximately 100MB - and does not grow much when using the filters generously. On IE11 - it is around 722MB and easily grows to 1.8GB within about a 15 minute session. Some of this is probably the performance analyzer using memory in IE11 but regardless, after using the app for a while and refreshing the page - eventually the charts fail to load because IE11 has run out of memory.  I'm going to work on the backend and see if we can consolidate some of the metric files so that we are downloading less data to begin with. I'm hoping you can hunt down some ways to optimize memory and ensure we are cleaning up in our effect hooks and closures.

I often see this warning in the console when switching between the dimensions charts. Those might be a good place to start ensuring we are cleaning up our effect hooks. Maybe start there but we really should check all of them.
`Warning: Can’t call setState (or forceUpdate) on an unmounted component. This is a no-op, but it indicates a memory leak in your application. To fix, cancel all subscriptions and asynchronous tasks in the componentWillUnmount method.`

A few other things to look into
* 'use strict' mode
* de-structuring objects and only passing necessary objects into children components.

I found this helpful:
https://blog.appsignal.com/2020/05/06/avoiding-memory-leaks-in-nodejs-best-practices-for-performance.html

**Additional context**

Wmaileq:
## Description of the change

- Made abortable promises
- The same file does not load twice

## Type of change

## Related issues

> Closes #606 
> Closes #564 

### Development

### Code review


jovergaag (approved):
Nice. Thanks for breaking this out so we can get it merged while we come up with a better cacheing solution.

PR #619, opened 2020-11-18
downloads refactor

### LINKED ISSUE #616: Swap the axis of the data export for specific charts
Data exports for the following charts are difficult to read due to the single row/multiple column format. Let's swap the axis of the export so that it is more rows than columns.

Apply this change for the following exports:
-revocationsByDistrict
-revocationsOverTime
-revocationsByViolation
-revocationsByOfficer
-revocationsByGender
-revocationsByRace
-revocationsByRiskLevel

**Additional context**

Wmaileq:
## Description of the change

- Refactored `downloads` functions
- Added export csv rotate if columns count more than rows count. 

## Type of change

## Related issues

> Closes #616 

### Development

### Code review


Wmaileq:
@jovergaag 
Almost finished here.

Need to test all the exports and then it will be ready for review. Also tests needed here.

Also, tomorrow I will investigate if we can replace this `document.getElementById`s with react approach.

src/components/charts/common/bars/utils.js
```diff
@@ -99,31 +99,22 @@ export function configureDownloads(
...
     label: bar.label,
     data: countsByType[bar.key],
   }));
 
   const chartTitle = upperCase(exportLabel);
 
-  const convertValuesToNumbers = false;
-
-  configureDownloadButtons(
+  configureDownloadButtons({
```
jovergaag, line 109: This is so much nicer with the props object, thank you.

src/utils/downloads/configureFileName.js
```diff
@@ -0,0 +1,24 @@
+import getTimeStamp from "./getTimeStamp";
+
+function configureFilename(chartId, filters, withoutFilters) {
```
jovergaag, line 3: It looks like the third param is `shouldZipDownload` when this is being used in downloads, lines 44, 117, etc. Mismatch withoutFilters/shouldZipDownload

jovergaag: Might want to update it here or there to make it less confusing.

src/utils/downloads/createMethodologyFile.js
```diff
@@ -0,0 +1,43 @@
+import moment from "moment";
```
jovergaag, line 1: Can you add a test for this, just to ensure the major sections are there.

It would also be nice to include a test on the chartTitle per locale to ensure our translations stay intact.

src/utils/downloads/downloadCanvasAsImage.js
```diff
@@ -0,0 +1,76 @@
...
+
+function transformCanvasToBase64(canvas, chartTitle, filters) {
+  const topPadding = 120;
+  const temporaryCanvas = document.createElement("canvas");
+  temporaryCanvas.width = canvas.width;
+  temporaryCanvas.height = canvas.height + topPadding;
+
+  // Fill the canvas with a white background and the original image
+  const destinationCtx = temporaryCanvas.getContext("2d");
+  destinationCtx.fillStyle = "#FFFFFF";
```
jovergaag, line 16: There's a lot of static styling in here. Are any of these styles shared with the charts themselves? Can we point both of them at a single stylesheet to ensure the download/chart stay in sync, and make it easier to update in the future.

src/utils/downloads/downloads.js
```diff
@@ -0,0 +1,208 @@
...
+  chartDatasets,
+  chartLabels,
+  chartBox,
+  filters,
+  convertValuesToNumbers,
+  timeWindowDescription,
+  shouldZipDownload,
+}) {
+  const filename = configureFilename(chartId, filters, shouldZipDownload);
+  const downloadChartAsImageButton = document.getElementById(
```
jovergaag, line 118: Were you not able to find another way to do this?

src/utils/downloads/transformChartDataToCsv.js
```diff
@@ -0,0 +1,55 @@
...
+    });
+  }
+
+  try {
+    return await csvExport(formattedData, {
+      mapHeaders: (header) => header.replace("dimension", ""),
+    });
+  } catch (e) {
+    // eslint-disable-next-line no-console
+    console.log(e);
```
jovergaag, line 50: Should we `console.error` here instead?

src/utils/downloads/transformChartDataToCsv.js
```diff
@@ -0,0 +1,55 @@
...
+  convertValuesToNumbers,
+  isTable
+) {
+  const datasetsWithoutTrendLine = datasets.filter(
+    (dataset) => dataset.label !== "trendline"
+  );
+
+  let formattedData;
+
+  if (!isTable && labels.length > datasetsWithoutTrendLine.length) {
```
jovergaag, line 15: This approach is working great other than for the matrix. The matrix axis got flipped which is confusing. I think that one should never be flipped no matter how many labels there are.

jovergaag (changes requested):
These exports are so much easier to read now! A few requested changes:

- For the district names, please convert them to strings so they will all be left justified. Most programs opening a csv will left justify strings and right justify numbers so for district names they are all over the place.
- Please add the name of the dimension as the column header for the first column.

A few other comments in line. 

Great work!

jovergaag:
One more thing, the `TenantExportMenu` and `ExportMenu` continue to make me think whenever I need to remember which is which. Can you rename `TenantExportMenu` -> `ChartCartExportMenu` please.

src/utils/downloads/configureFileName.js
Wmaileq: Yes, I've done it on purpose, because, in my opinion, function `configureFilename` does not need to know about `shouldZipDownload` I thought this parameter name would be more descriptive.
When you read `withoutFilters` you know that this parameter disables filters, but when you see `shouldZipDownload` I think it is a bit confusing, and you need to go through func body to realize what this parameter do.

src/utils/downloads/downloads.js
Wmaileq: Yes, I think this should be done in new PR, because this refactor will result in a lot of changes, but PR is big already.

src/utils/downloads/createMethodologyFile.js
Wmaileq: I am not sure that I got it, looks like chartTitle goes from props, not from translations...
Added unit tests, so all the external functions mocked.

jovergaag: Oh you are right. Nevermind :)

src/utils/downloads/downloadCanvasAsImage.js
Wmaileq: All exported charts consist of: heading, filters applied, and the whole copied chart canvas.
1. We have no filters on the website charts
2. I am not sure that the heading should inherit from the website, but if this is the case I can do it.
3. the whole chart canvas is just copied to export canvas, so no differences here.

Maybe it is a good idea to inherit background from the source.

Wmaileq:
Added dimensions to all the export data, tested all the charts, all should work as expected!

src/components/charts/GeoViewTimeChart.js
```diff
@@ -325,6 +325,7 @@ class GeoViewTimeChart extends Component {
       chartLabels: officeNames,
       chartBox: document.getElementById(this.props.chartId),
       filters: this.props,
+      dimension: "Region",
```
jovergaag, line 328: Let's call this 'County'

jovergaag (changes requested):
I have just a few small nit picks on the labels. We also still need to convert all the row labels to strings to ensure they are left justified.

Thanks!

src/components/charts/common/bars/utils.js
```diff
@@ -90,15 +90,15 @@ export const groupByMonth = (barKeys) => (dataset) =>
     }))
   )(dataset);
 
-export function configureDownloads(
+export function configureDownloads({
```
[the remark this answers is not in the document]
Wmaileq: Yeah, forgot to comment but I investigated here, looks like CSV doesn't support types, so we cannot set the type of cell to string implicitly, different CSV readers have different technics.
We can add `'` to the start of the cell, this converts cell to string in Excel, but in `Mac's Numbers app` it will look like this:
image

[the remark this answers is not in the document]
jovergaag: Oh. Right that makes sense. Let's not add the `'` and just leave it the way it is. Thanks for looking into it!

jovergaag (approved):
Wonderful, thank you! 🎉

PR #620, opened 2020-11-20
[IE11] multi-select onBlur bug

### LINKED ISSUE #596: [fullstory] District filter remains open after interacting with other parts of the page
From viewing a Fullstory session, it appears that one users's district filter remains open even after they click on the case type filter and interact with other parts of the page. I did not see this behavior occur for any other users. The user was using IE11.

**Additional context**
Fullstory is sometimes a little janky, so I'm not sure if this was actually occurring, or the video was off. Can close if this ends up being not reproducible.

jovergaag:
## Description of the change

We had a bug in IE11 where clicking outside of the multi-select menu did not close the menu. This bug only existed react-scripts builds and not in webpack builds which was strange.

The bug was fixed by cleaning up how we are searching for the select input to pass to the ValueContainer. We were not finding the input correctly with the react-scripts build (timing maybe) so the ref was lost, and therefore the focus/blur was broken.

This fix ended up breaking the `closeMenuOnSelect={false}` for IE11, so I manually changed the `onFocus` event handler to help IE11 out.

This has been deployed to staging for QA vs what you see in production.

## Type of change

## Related issues

> Closes #596

### Development

### Code review

### DIFF
--- a/src/components/controls/MultiSelect/ValueContainer.js
+++ b/src/components/controls/MultiSelect/ValueContainer.js
@@ -24,9 +24,9 @@ import { optionPropType } from "../../propTypes";
 const ValueContainer = ({ allOptions, summingOption, children, ...props }) => {
   const { selectProps, getValue } = props;
   const values = getValue();
-  const selectInput = React.Children.toArray(children).find((child) =>
-    child.type === components.Input ? child : null
-  );
+  const selectInput = React.Children.map(children, (child) => {
+    return child.type === components.Input ? child : null;
+  });
 
   const isAll =
     !selectProps.inputValue &&
--- a/src/components/controls/MultiSelect/ValueContainer.js
+++ b/src/components/controls/MultiSelect/ValueContainer.js
@@ -24,9 +24,8 @@ import { optionPropType } from "../../propTypes";
 const ValueContainer = ({ allOptions, summingOption, children, ...props }) => {
   const { selectProps, getValue } = props;
   const values = getValue();
-
-  const selectInput = React.Children.toArray(children).find((input) =>
-    ["DummyInput", "Input"].includes(input.type.name)
+  const selectInput = React.Children.toArray(children).find((child) =>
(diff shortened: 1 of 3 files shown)


src/components/controls/MultiSelect/ValueContainer.js
```diff
@@ -24,10 +24,9 @@ import { optionPropType } from "../../propTypes";
 const ValueContainer = ({ allOptions, summingOption, children, ...props }) => {
   const { selectProps, getValue } = props;
   const values = getValue();
-
-  const selectInput = React.Children.toArray(children).find((input) =>
-    ["DummyInput", "Input"].includes(input.type.name)
-  );
+  const selectInput = React.Children.map(children, (child) => {
+    return child.type === components.Input ? child : null;
```
macfarlandian, line 28: what was "DummyInput", just some legacy thing we don't need to support anymore?

macfarlandian (commented):
I also did a quick QA on IE 11 in browserstack and I am now truly and deeply feeling your pain

src/components/controls/MultiSelect/ValueContainer.js
jovergaag: I *think the "DummyInput" was being used to try to focus the select input at some point but ended up going a different route. I found a blog post about creating a custom `ValueContainer` and it looks like this section was copied directly from that blog. A while back I found a line of code that was commented out that was trying to use the "DummyInput" to focus the input, so it looks like someone was struggling with this focus problem a while ago, but ended up not using the "DummyInput" in the end.

jovergaag:
> I also did a quick QA on IE 11 in browserstack and I am now truly and deeply feeling your pain

Thanks. It is truly a test of patience. But already much better than it was 3 weeks ago!

macfarlandian (approved):

macfarlandian:
one of these days I will have to get a VM testing environment running for IE 11. today is not that day

PR #647, opened 2020-12-08
Implement MobX and Typescript for RootStore and PageStore

### LINKED ISSUE #646: Add MobX and Typescript
As we are moving towards a more dynamic interaction with the API, we will need to have a better idea of the filter values, what values exist in the current in. memory dataset from the API, and be able to trigger new API requests easily when needed.

MobX and Typescript will both help us do this more easily and free of errors. A central store and more strongly typed data models and filter options will ensure clean request/response/filter actions.

**Additional context**

jovergaag:
## Description of the change

As we are moving towards a more dynamic interaction with the API, we will need to have a better idea of the filter values, what subset of data is currently cached, and be able to trigger new API requests easily when needed. A central store and more strongly typed data models and filter options will ensure clean request/response/filter actions. Enter MobX and TypeScript

We are still nailing down the specifics of how the modified API will operate, but it is clear that no matter the implementation, we will need better metric data/filter state management. I have been wanting to start slowing converting the JavaScript in this repo to TypeScript, these new stores seemed like the perfect place to get started.

This PR:
* installs MobX and adds a `RootStore` as well as a `PageStore` which was transitioned from a context. The `PageStore` is a very simple store, great for the first implementation. 
* installs Typescript and related linters/lint config. I haven't worked much with TypeScript so I copied a lot of the `.eslintrc.json` and `.tsconfig.json` from the public-dashboard repo. If anyone has any suggestions on what might not be useful here please let me know and I can make those adjustments! Changes a few files that were failing with these new linting config changes to adhere
* changes the `yarn test` script from using jest to using react-scripts. Also changes one test which started failing after this change, but it now covers some new functionality in that file anyway.
* removes the pre-push test hook because I have found it to be too cumbersome and no longer the `yarn test` script watching the files
* adds the `src/react-app-env.d.ts` - This is an artifact of create-react-app with TypeScript but I don't see it in public-dashboard so I'm not sure why it is being generated here and not there...

## Type of change

## Related issues

> Closes #646

### Development

### Code review


src/hooks/usePageLayout.js
```diff
@@ -16,27 +16,21 @@
...
-            type: "update",
-            payload: { isTopBarShrinking: true },
-          });
-        } else if (isTopBarShrinking && window.pageYOffset < 5) {
-          pageDispatch({
-            type: "update",
-            payload: { isTopBarShrinking: false },
-          });
+        if (!pageStore.isTopBarShrinking && window.pageYOffset > 90) {
+          runInAction(() => set(pageStore, "isTopBarShrinking", true));
```
jovergaag, line 31: I want to do some more reading to learn more about best practices for actions, but it seems like `runInAction` is recommended for async processes.

src/react-app-env.d.ts
```diff
@@ -0,0 +1 @@
+/// <reference types="react-scripts" />
```
macfarlandian, line 1: FWIW this is in public-dashboard, but it's in `/spotlight-client/src` because it's only for that package

src/hooks/usePageLayout.js
macfarlandian: this seems correct to me, but I think it's needed here not because this is async but because you are mutating the store from outside. (the mutation itself is synchronous, it's not like you changing one thing, waiting on an async function, then changing another thing)

I am thinking it might be clearer to move the update logic into a method on PageStore, so that it's closer to the data model? `makeAutoObservable` should make that an action and then you won't need to manually call all this MobX stuff (and if, e.g., another part of the application also needed to update this value, it would not also need to manually call all this MobX stuff).

src/hooks/usePageLayout.js
```diff
@@ -16,27 +16,21 @@
...
-          pageDispatch({
-            type: "update",
-            payload: { isTopBarShrinking: true },
-          });
-        } else if (isTopBarShrinking && window.pageYOffset < 5) {
-          pageDispatch({
-            type: "update",
-            payload: { isTopBarShrinking: false },
-          });
+        if (!pageStore.isTopBarShrinking && window.pageYOffset > 90) {
```
macfarlandian, line 30: does this hook need to create a Mobx "observer" somehow? or is it just that all components that are using this hook need to be observers? I am fuzzy on this but also not sure if it makes any practical difference

jessex (commented):
Excited to see these improvements beginning! This app is going to play an ever-increasing role throughout 2021 so getting it onto a more solid footing is good for us all.

Just want to make a comment that you're probably already keeping in mind: this marks the beginning of _two_ longitudinal introductions of new tech and is happening simultaneous to a bunch of other ongoing feature development, so let's be extra careful in testing to make sure that the app still runs and functions as expected with each new commit related to these changes, so we don't end up in a situation where the app is not releaseable but there's a feature or bug fix or security patch we need to get to production.

macfarlandian (changes requested):
whoops I meant to put all my comments in a review .... anyway this mostly looks good to me, just some questions and small stuff

src/hooks/usePageLayout.js
jovergaag: I originally did not have these lines inside of an action, and even with `makeAutoObservable` I was getting a console warning about changing an observable attribute outside of an action.

I also read found this snippet in the [docs](https://mobx.js.org/actions.html): 
```To leverage the transactional nature of MobX as much as possible, actions should be passed as far outward as possible. It is good to mark a class method as an action if it modifies the state. It is even better to mark event handlers as actions, as it is the outer-most transaction that counts.

That is why I tried pulling the action out of the `PageStore`.  I'm not quit sure in what situations its best to pull that action outward or not. More reading for me would be useful I think...

Comment (macfarlandian) [reply src/hooks/usePageLayout.js:31]: hmmm interesting ... I guess that makes sense. I wonder if that means you could, for example, wrap the entire event handler function in a call to `action` instead of using `runInAction`? 

Also I didn't really catch this before but I'm not sure whether you need `set` here either ... `pageStore` is already an observable class instance, I think you ought to be able to just use a normal assignment operator

File: src/hooks/usePageLayout.js (hunk 3 above)
Comment (jovergaag) [reply src/hooks/usePageLayout.js:29]: You know, I'm not entirely sure how the "observer" functionality is being injected here. I wrote a test in [this commit](https://github.com/Recidiviz/pulse-dashboard/pull/647/commits/20fa4abc3fb51a1efc08b2e6f5ba1957dd45918f) that shows that a change to `pageStore.isTopBarShrinking` triggers the `useLayoutEffect` hook to run, even without adding observer anywhere in this hook or on its parent component.

I added `autorun` in the hook as [this issue](https://github.com/mobxjs/mobx-react-lite/issues/114) suggests. It did not seem to change the behavior at all. I guess I will leave the `autorun` in there but am not convinced that it is necessary, maybe more of a best practice in this particular scenario?

File: src/hooks/__tests__/usePageLayout.test.js [hunk 4]
```diff
@@ -0,0 +1,130 @@
...
+      .mockImplementation((cb) => cb());
+
+    const wrapper = ({ children }) => {
+      rootStore = new RootStore();
+      return (
+        <StoreContext.Provider value={rootStore}>
+          {children}
+        </StoreContext.Provider>
+      );
+    };
```
macfarlandian, line 44: I have been reading through the React integration docs for Mobx and in [this section](https://mobx.js.org/react-integration.html#using-external-state-in-observer-components) under "using global variables" they say you can just instantiate a store inside a module and consume it directly rather than sending it through a context? The docs claim this "might complicate unit testing" but honestly .... it seems like that would be simpler? Otherwise your datastore is locked up inside React and you can't get access to it without doing something like this that deviates from the way it is actually used in the application. What if both the hook module and this test module could just import the same `RootStore` instance and you didn't have to create this shadow provider?

src/hooks/__tests__/usePageLayout.test.js
```diff
@@ -0,0 +1,130 @@
...
+      const { pageStore } = rootStore;
+      setIsTopBarShrinkingSpy = jest.spyOn(pageStore, "setIsTopBarShrinking");
+    });
+
+    describe("when window.pageYOffset > 90", () => {
+      it("isTopBarShrinking should not be changed", () => {
+        window.pageYOffset = 100;
+        global.dispatchEvent(new Event("scroll"));
+
+        expect(setIsTopBarShrinkingSpy).toHaveBeenCalledTimes(0);
```
macfarlandian, line 68: I kind of think it would be better to just verify the value of `isTopBarShrinking` here rather than spying on the method. 

In general, in these tests, I don't think the spies are really adding much value? I worry they will be brittle ... I would just verify that the state is correct after calling the hook and not worry about how it got that way

src/hooks/usePageLayout.js
jovergaag: I agree that the better implementation is to add a setter to the PageStore - done. In my reading I've seen lots of examples of the store being manipulated directly in the components, but in general I think I would also lean towards keeping all of the logic and manipulation centrally, more like a reducer pattern. 

I haven't been able to find a clear answer about when an action is required when using `makeAutoObservable`.  The action [docs](https://mobx.js.org/actions.html) say that it makes much of the action functionality automatic, but I haven't found any specifics other than that.

src/hooks/usePageLayout.js
```diff
@@ -28,14 +27,13 @@ const usePageLayout = () => {
       cancelAnimationFrame(frame.current);
       frame.current = requestAnimationFrame(() => {
         if (!pageStore.isTopBarShrinking && window.pageYOffset > 90) {
-          runInAction(() => set(pageStore, "isTopBarShrinking", true));
+          pageStore.setIsTopBarShrinking(true);
         } else if (pageStore.isTopBarShrinking && window.pageYOffset < 5) {
-          runInAction(() => set(pageStore, "isTopBarShrinking", false));
+          pageStore.setIsTopBarShrinking(false);
```
macfarlandian: I think this is fine too ... in practice this function is only going to make one state change so it doesn't really matter where the transaction boundary is. Maybe the rule of thumb we should follow is that we should pull the transaction boundaries out into the UI components when there are multiple changes and we don't want to display any intermediate UI states between them? That is just kind of a guess at this point though. Also I don't know what happens if you, e.g., nest an action inside another action, it's not clear to me whether Mobx will figure that out or if it will result in some redundancy? maybe someday we'll find out

macfarlandian (commented):
love to see the tests 🏅

src/hooks/__tests__/usePageLayout.test.js
```diff
@@ -0,0 +1,130 @@
...
+
+    it("triggers the effect to run", () => {
+      window.pageYOffset = 100;
+      expect(setIsTopBarShrinkingSpy).toHaveBeenCalledTimes(0);
+
+      rootStore.pageStore.setIsTopBarShrinking(false);
+      rootStore.pageStore.setIsTopBarShrinking(true);
+      rootStore.pageStore.setIsTopBarShrinking(false);
+
+      expect(setIsTopBarShrinkingSpy).toHaveBeenCalledTimes(3);
```
jovergaag, line 127: Oh duh. I'll look at this again in the morning.

src/hooks/__tests__/usePageLayout.test.js
jovergaag: That sounds much simpler. As I was reading your PR last week, the stores definitely seemed so much simpler for both using and testing.

Also, after having re-read the docs last week, there is a lot I would like to change in this PR so I think I'm just going to give it an overhaul.

PR #650, opened 2020-12-15
Fixes bug in District Chart when applying more restrictive to less restrictive filtering

### LINKED ISSUE #649: District chart - createGenerateChartData toLowerCase error
**Describe the bug**
We are getting a `TypeError: Cannot read property 'toLowerCase' of undefined` when trying to generate the District Chart after certain filtering actions. It seems like the `currentDistrict is not being set.

Only an issue for MO, not PA.

This exists on master but it has not yet been deployed to Production.

**To Reproduce**
1. From the MO dashboard, choose the Case Type=SEX_OFFENSE filter
2. Let the charts load, then choose Case Type=ALL filter
3. Note that the District Chart fails to load.

**Expected behavior**
District Chart loads without error.

**Screenshots**

 - OS [e.g. Windows 10]: 
 - Version [e.g. 22]: 

 - Device: [e.g. iPhone6]: 
 - OS: [e.g. iOS8.1]: 
 - Version [e.g. 22]: 

**Additional context**

jovergaag:
## Description of the change

We had a bug on the District Chart where we were seeing a `TypeError: Cannot read property 'toLowerCase' of undefined` error when switching from (for example) Case Type=SEX_OFFENSE to Case Type=ALL filters. The `labels` object occasionally had not updated to include the larger dataset before the chart tried to access it at the higher `dataIndex` . 

I traced this back, it started in the to the node/packages upgrade commit so there must be something slightly different with how chartJs is looping through the dataset given the new node env. I tested this thoroughly to ensure that the guard is not making us miss coloring one of the `currentDistrict` labels correctly. Even if a `currentDistrict` is the first or last of the labels, either before or after the filter change, chartJs eventually makes its way back through and applies the `currentDistrict` color treatment.

-This also includes a refactor to the statistically significant shading application on a few charts
-This also includes cleaned up of a test that was giving us a false positive (I saw the errors in the log but the test was not failing).
-This also includes a small update to the README after some feedback from Daniela as she was working on getting her environment set up today

## Type of change

## Related issues

> Closes #649

### Development

### Code review

### DIFF
--- a/src/utils/charts/__tests__/significantStatistics.test.js
+++ b/src/utils/charts/__tests__/significantStatistics.test.js
@@ -15,10 +15,14 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 // =============================================================================
 
+import pattern from "patternomaly";
+
 import {
   isDenominatorStatisticallySignificant,
   isDenominatorsMatrixStatisticallySignificant,
   tooltipForFooterWithCounts,
+  applyStatisticallySignificantShading,
+  applyStatisticallySignificantShadingToDataset,
 } from "../significantStatistics";
 
 const statisticallySignificantDenominators = [0, 100, 1019];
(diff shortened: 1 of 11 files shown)


src/utils/charts/__tests__/significantStatistics.test.js
```diff
@@ -70,6 +74,93 @@ describe("isDenominatorsMatrixStatisticallySignificant function", () => {
...
+    const drawSpy = jest.spyOn(pattern, "draw");
+    drawSpy.mockReturnValue("#ffffff");
+
+    const given = statisticallyNotSignificantDenominators;
+    const color = "anyColor";
+
+    given.forEach((denominator) => {
+      expect(applyStatisticallySignificantShading(color, denominator)).toBe(
+        "#ffffff"
+      );
```
macfarlandian, line 99: what does a normal return value look like when a pattern is applied? i'm wondering whether this might be a good place to use a snapshot test instead of a mock

jovergaag: Great suggestion. 

I was originally mocking it because patternomaly's `pattern.draw` creates and manipulates a canvas which was was throwing an error in Jest. I ended up installing and setting up `jest-canvas-mock` which allows `pattern.draw` to return successfully. I think these tests are much more clear now by just checking the `shapeType`. LMK what you think.

macfarlandian (approved):
Great find! I agree these tests are much easier to understand now

PR #671, opened 2020-12-17
Implement a redis cache and a new endpoint to refresh cache

### LINKED ISSUE #652: [API] Fetch files from the bucket and save to redis cache using the new cacheKey naming convention
There are two primary tasks here:
1. Fetch files from bucket on demand
2. Establish a Redis cache and save the file, unaltered, to the cache
3. Point the `newRevocations` endpoint to retrieve metrics from this cache rather than the in memory cache
4. Set the cache to invalidate after 24 hours and fetch new files again at that point automatically

The cache keys should be:
`${stateCode}-${file}-original`

**Additional context**
Some things we will want to keep an eye on are that the Redis cache persists, even if the instance scales to 0.
Latency of the cache response

daschi:
## Description of the change

**Update:** This PR has been updated to include a `cron.yaml`[ to deploy 2 cron jobs for refreshing the US_MO and US_PA ](https://console.cloud.google.com/appengine/cronjobs?serviceId=default&project=recidiviz-dashboard-staging)external caches. They're both set up to run at 10 and 10:30am PST, an hour or so after we expect the files to be updated. We're not enabling TLS at this time, the Google Support person let us know the feature is not yet working for GAE -> Redis connections. I am still enabling an AUTH password for connections. 

For development, we can still run `yarn dev` to start the backend and frontend together, and I've added `yarn server:dev` to run the backend server with the redis server together. More instructions were added to the readme.

This is currently deployed to staging, [here's a link to the memorystore instance.](https://console.cloud.google.com/memorystore/redis/locations/us-east1/instances/public-dashboard-instance-1/details?project=recidiviz-dashboard-staging) (I accidentally called it public-dashboard-instance-1 instead of pulse-dashboard :sob: let me know if I should recreate it with a new instance ID, or if this is OK)

~I'm opening this is a DRAFT PR to get some feedback and thoughts before going too far with this! Please let me know if this approach doesn't align with the proposed solution / design. Here are some of my thoughts, happy to chat about everything in more detail too.~

1. The caching strategy is a mix of "write-through" and "lazy loading". The idea is to have a daily cron job that hits an endpoint (currently `/api/:stateCode/refreshCache`) after metrics are exported to update the cache files. The downside is that we may cache more files than we need to on a daily basis, but it does mean that all of the files are available in the cache at any point a user hits it. The endpoint for a specific file is using the "lazy loading" strategy, so if for some reason the cache misses, the metric file endpoint would create it. There is a TTL set on both for 24 hours, and a date (right now just `Date.new().getDate()`) for each cache key so that the refreshCache uses a new key every 24 hours. This might also cause issues if we need to refresh the cache multiple times in one day, so still thinking through the best approach for that...

2. Security - for this first pass I didn't focus on securing the endpoint, but I was also thinking it might make sense to move this work of processing & refreshing the cache to a cloud function. Would love to hear your thoughts on that, and if you think we should stick with an endpoint. I also have not set up the actual redis instance on GC, which can be setup using TLS and an auth password. 

3. Depending on the endpoint, the cache will either be for a single file, or for all the files in a metric type. The keys will look like:
- US_PA-newRevocation-revocations_matrix_distribution_by_risk_level-17
- US_PA-newRevocation-17

4. We are using both the redis cache and the app memory cache because we've decided not to change anything yet for the US_ND dashboard. This is why there are both `cacheInMemory` and `cacheInRedis` helpers. 

5. Another question I had for everyone, now that we have the redis service, I was wondering if we should add a Dockerfile to run the API server and redis server. Since there's only two, I thought it might be overkill to add that at this point, but wanted to hear your thoughts. 

TODO - What I'll continue working on 
- [x] Update the existing tests and add new tests 
- [x] Setup the redis instance on Google Cloud
- [ ] Add the connection credentials to the vault 
- [x] Add a cron.yaml to run refresh cache on a schedule
- [ ] Deploy a staging app to live test
- [x] Add redis-server command to npm scripts to start servers
- [x] Add instructions for running redis locally for development

Thanks everyone for your help!!

## Type of change

## Related issues

> Closes #652

### Development

### Code review


server/core/redisCache.js
```diff
@@ -0,0 +1,73 @@
...
+const redisStore = require("cache-manager-ioredis");
+const Redis = require("ioredis");
+
+const REDISHOST = process.env.REDISHOST || "localhost";
+const REDISPORT = process.env.REDISPORT || 6379;
+
+// Expire items in cache after 1 day
+const REDIS_CACHE_TTL_SECONDS = 60 * 60 * 24;
+
+const redisInstance = new Redis({
```
jessex, line 39: Just a question: I presume that in production we're planning to use [Memorystore](https://cloud.google.com/memorystore) instead of maintaining our own Redis cluster. If so, is this the way that we'd connect to Memorystore in our deployed environments, or are you not looking that far ahead yet?

server/core/redisCache.js
```diff
@@ -0,0 +1,73 @@
...
+  ttl: REDIS_CACHE_TTL_SECONDS,
+});
+
+const redisCache = cacheManager.caching({
+  store: redisStore,
+  redisInstance,
+});
+
+// Timestamp will change after 24 hours, at which point the cache will refresh
+const CACHE_KEY_TIMESTAMP = new Date().getDate();
```
jessex, line 52: I see, though, that this is defined as a constant at the top-level of a file. So I presume this exact line of code only gets invoked on server start. Is that right? If so, then with longer-lived instances, `CACHE_KEY_TIMESTAMP` won't actually rotate when we want it to.

server/core/redisCache.js
```diff
@@ -0,0 +1,73 @@
...
+  db: 0,
+  ttl: REDIS_CACHE_TTL_SECONDS,
+});
+
+const redisCache = cacheManager.caching({
+  store: redisStore,
+  redisInstance,
+});
+
+// Timestamp will change after 24 hours, at which point the cache will refresh
```
jessex, line 51: Today, calculation pipelines tend to execute in the mid-to-late morning, I believe. This would seem to mean that, on average, the cache key will rotate before the day's metrics are produced (though see my comment below about when execution will actually happen). If that's true, then users will often be looking at metrics that are a day out-of-date. Is that right?

jessex (commented):
A few fly-by remarks! I'll think more about the high-level questions in your PR description later on, but wanted to go ahead and throw these questions into the fray. Excited to see this happening!

server/core/redisCache.js
daschi: I did look into it a bit and it looks like you can create a [Redis instance in Memorystore](https://cloud.google.com/appengine/docs/standard/nodejs/using-memorystore#setup_redis_db), and use the provided IP address & port. The GAE does need to be connected to a VPC network, which I still need to look more into. I assume it's not currently in a VPC?

server/core/redisCache.js
daschi: Oh yeah, good catch...I can move this down into the function definition for `cacheInRedis`.

server/core/redisCache.js
```diff
@@ -0,0 +1,73 @@
...
+const CACHE_KEY_TIMESTAMP = new Date().getDate();
+const redisClient = redisCache.store.getClient();
+redisClient.on("error", (error) => console.error("ERR:REDIS:", error));
+
+function cacheInRedis(cacheKey, fetchValue, callback) {
+  redisCache.wrap(
+    `${cacheKey}-${CACHE_KEY_TIMESTAMP}`,
```
macfarlandian, line 58: I am wondering whether you even need a timestamp in the cache key at all. if there's a TTL on the value and an endpoint where we can flush it whenever we want, isn't that enough? 

maybe you still want the timestamp for something but it could just be part of the cached value ... but if the refresh function was a bit safer to run more frequently (e.g. it checked whether the files were new or the cache was expired before refreshing) then maybe it would be more flexible or give us more control to refresh on a cron schedule or in response to new export files landing in the bucket, etc

server/core/redisCache.js
daschi: Hm, yeah I think I understand what you mean. If a user hits the endpoint (instead of the triggered refreshCache endpoint) then it will create the cache key with that day's date and never get refreshed with the new day's metrics. 

I'm also thinking about this with @macfarlandian's comment, and it would be better if the `refreshCache` endpoint could check whether files were new. Maybe we can cache the last updated date of the file, and use that to refresh the cache, and remove the timestamp on the cache key.

server/core/redisCache.js
daschi: I was trying to think of a way to change the key whenever we ran the `refreshCache` endpoint, but since I ended up using this `cacheInRedis` function wrapped around all the endpoints, I don't think it will work as I thought it would. I like your idea of making the refresh function check whether it should refresh a file, and store the last updated timestamp for the file. I think that would mean I couldn't use this `wrap` helper from cache-manager though, and would need to implement the logic to know when to cache for this endpoint.

server/core/redisCache.js
jovergaag: I agree that we should adjust this timing such that it runs after the metrics hit the bucket. 

Eventually we would like to set up this instance to listen for an event and refresh the Redis cache when the event is received. The metric files being changed or added to the bucket would trigger that event. 

One question that I had was whether to keep this 24 hour cache refresh mechanism in addition to the event trigger. In the case that there is an error in the prod data pipeline and the metrics files are not updated, do we think that the files should be refreshed after 24 hours? Two benefits I can see for this, 1. Backup for the possibility that an event message was lost or malformed, and 2. if the cache gets corrupted for some reason. Thoughts on this?

server/core/refreshRedisCache.js
```diff
@@ -0,0 +1,74 @@
...
+          cacheInRedisErrors.push(err);
+        }
+      }
+    );
+  });
+  return { errors: cacheInRedisErrors.length > 0 ? cacheInRedisErrors : null };
+}
+
+function refreshRedisCache(fetchMetrics, stateCode, metricType, callback) {
+  const cacheKeyPrefix = `${stateCode}-${metricType}`;
```
jovergaag, line 54: I'm wondering if we need `metricType` in the cacheKey. For the Lantern dashboard, the `metricType` is always `newRevocation`.

jovergaag: On second thought, it may be useful if the API is ever consolidated such that multiple FEs use this same API. I suppose it doesn't hurt to leave it there in case there are multiple endpoints for multiple FEs eventually using the redis cache.

jovergaag (commented):
I think this is heading in the right direction! Great job.

> Security - for this first pass I didn't focus on securing the endpoint, but I was also thinking it might make sense to move this work of processing & refreshing the cache to a cloud function. Would love to hear your thoughts on that, and if you think we should stick with an endpoint. I also have not set up the actual redis instance on GC, which can be setup using TLS and an auth password.

I like the idea of a cloud function. @macfarlandian had mentioned in a comment in the Design Doc that he and Julia found a way to use cloud functions to react to new files being added to a storage bucket in the covid repo. It seems like leveraging that knowledge and code already written as an example in the covid dashboard would alleviate our need to stand up and secure this endpoint. I would love to hear @macfarlandian's thoughts on this. We could write the cloud function to just refresh the endpoint for now for manual refreshes, then eventually add on the logic to watch the bucket.

> Depending on the endpoint, the cache will either be for a single file, or for all the files in a metric type. The keys will look like:

I'm trying to think of the case when we would cache all files in a metric type. The FE always requests individual files so I'm not sure there will ever be a request for the full set of files at once.

> Another question I had for everyone, now that we have the redis service, I was wondering if we should add a Dockerfile to run the API server and redis server. Since there's only two, I thought it might be overkill to add that at this point, but wanted to hear your thoughts.

It seems to me that the next step would be to stand up a Memorystore with Redis in GAE. Through that work we will have a better idea of whether this Redis implementation/package is going to be the best approach. Once we know that, then we can probably make a better decision about what the local dev experience will look like, whether we will need a Docker container or not.

macfarlandian:
@jovergaag yeah daniela and I were just discussing this offline ... running it on a schedule would be easier to stand up because you can do it in the same GCP project as these other services. The bucket watcher would need to live on the backend project and probably have cross-project access permissions but ought to be able to hit the same endpoint so it's a good incremental improvement

server/core/redisCache.js
nbhargava: Even if they're not in the same VPC, it shouldn't be too much work to set up the networking to ensure that they can talk to one another.

nbhargava (commented):
> The bucket watcher would need to live on the backend project and probably have cross-project access permissions
Do the frontends not currently live with the rest of Recidiviz services? [Regardless, GCP actually has super reasonable cross-project permissions, so it shouldn't be too bad.]

> Another question I had for everyone, now that we have the redis service, I was wondering if we should add a Dockerfile to run the API server and redis server.
I've always found putting anything with `yarn` in a docker container to be more trouble than it's worth, but I think we should definitely have a `docker-compose.yaml` that at minimum spins up a local Redis instance.

server/core/redisCache.js
daschi: @macfarlandian I just updated this to cache the file's timestamp, which I think is a good move regardless of how we get/set from the `refreshRedisCache` endpoint. I'm curious to see what you think of the logic, which currently is:

1. Fetch all the files from the bucket
2. If the file is in the cache, check whether it's been updated
3. If it has, update the cache value
4. If the file does not exist in the cache, add it to the cache

I'm not sure I'm seeing the benefit of checking whether it's been updated or not, since either way when we hit this endpoint, we would want the value to be updated? Or does this save some resources in case we hit the endpoint and the files have not changed?

daschi:
> I'm trying to think of the case when we would cache all files in a metric type. The FE always requests individual files so I'm not sure there will ever be a request for the full set of files at once.

@jovergaag That's currently how the US_ND endpoints cache is working, so wanted to keep that as it. Do the US_ND pages not request all the files at once?

server/core/redisCache.js
macfarlandian: hm good point ... I guess it doesn't actually serve much purpose in the end. I was originally thinking that you'd want the keys to expire at some point but I think I was just mixing up the different implementation details in my disordered mind?

server/core/redisCache.js
daschi: Yeah I had the same question, if it was worth adding a 24 hour expiration or not. I think it is, for both of the reasons you mentioned. Also,[ I decided to add a refresh threshold to the client requests](https://github.com/BryanDonovan/node-cache-manager#refresh-cache-keys-in-background), since those are using the `wrap` function.

server/core/redisCache.js
daschi: It made sense to me too at the time 😄 since the client requests are using `wrap` and `refreshThreshold`, even if we don't hit the `refreshCache` endpoint directly, they will expire after 24 hours and refresh. I think i'll remove this extra logic then, and just always set the value when this endpoint is hit. I like storing the file updated value too, in case we ever need that for debugging.

daschi:
@jessex I did some reading this morning and it seems like the Memorystore for Redis is always created in a VPC network, and the GAE standard environment must connect using a VPC Access Connector (and it's all very new to me and I'm sure I've missed something). I wanted to check in with you about this before moving forward with creating a Redis instance, since we chatted yesterday about looking into a service account solution. I haven't seen anything in the documentation yet about connecting with a service account, and wanted to ask if you had an idea about where that documentation might be. 

Here's links to what i've been looking at: 
- https://cloud.google.com/appengine/docs/standard/nodejs/connecting-vpc
- https://cloud.google.com/memorystore/docs/redis/connect-redis-instance-standard

jessex:
I think this all (mostly). makes sense to me, @daschi!

* The Node server will be set up with VPC access to be able to talk to Redis--nothing that's _not_ under our control in the VPC should be able to speak to Redis
* The endpoints on that Node server that invoke execution paths which touch Redis will be protected via JWT just like all of the other endpoints, so only authenticated users will be able to make API calls that touch Redis
* Being set up with VPC access will not prevent our continuing to use a service account to talk to GCS to retrieve new metric files

The only challenge here is in local development:
1. We'll need to run a local Redis instance to run the app locally
2. When running the _built_ app locally (via yarn build-staging or yarn build) while attempting to talk to data in staging or production (e.g. as part of debugging), we would _not_ be able to access Redis in staging or production, so in that case we would have a built app locally that is talking to GCS files in staging or production but using a purely local Redis cache. I think that that is _probably_ sufficient for our needs, but I want to make sure it has been considered!

server/core/demo_data/revocations_matrix_filtered_caseload.metadata.json
```diff
@@ -0,0 +1,6 @@
+{
```
jovergaag, line 1: This file will never be used since we're still using the .json version of this file vs the optimized .txt version.

jovergaag:
> 2\. When running the _built_ app locally (via yarn build-staging or yarn build) while attempting to talk to data in staging or production (e.g. as part of debugging), we would _not_ be able to access Redis in staging or production, so in that case we would have a built app locally that is talking to GCS files in staging or production but using a purely local Redis cache. I think that that is _probably_ sufficient for our needs, but I want to make sure it has been considered!

@jessex  I'm not sure if I understand this. When we are running a _built_ app, the local FE is hitting the deployed staging/prod API. That deployed API instance will have the connection with the Redis instance, so why would we need a local Redis cache?

jessex:
@jovergaag oh you are right, I totally apologize! Actually, the first scenario I mentioned is the relevant one and the second one is non-existent. In the first scenario where you are running locally _but not in demo mode_, then you have a local Node server talking to a live GCS bucket in staging or production. In that scenario, you would need to be talking to a local Redis instance because you wouldn't be able to connect to a deployed Redis instance since it is on a VPC. I think you all have that in mind already though, right?

jovergaag:
> @jovergaag That's currently how the US_ND endpoints cache is working, so wanted to keep that as it. Do the US_ND pages not request all the files at once?

Thats true. I was thinking about the `newRevocation` endpoint. The ND endpoints do return multiple files.

jovergaag:
> The bucket watcher would need to live on the backend project and probably have cross-project access permissions
> Do the frontends not currently live with the rest of Recidiviz services? [Regardless, GCP actually has super reasonable cross-project permissions, so it shouldn't be too bad.]

@nbhargava They are in different projects but still under the 'Recidiviz.org' organization.

jovergaag:
> In that scenario, you would need to be talking to a local Redis instance because you wouldn't be able to connect to a deployed Redis instance since it is on a VPC. I think you all have that in mind already though, right?

@jessex Yes we will need a local Redis server for local development. Daniela has that all figured out :)  A local redis server can be started up and monitored as part of the `yarn dev` script. 🎉   I believe she is going to add details about installing/managing the local redis server in the README in a "coming soon" commit. 

We also thought through the other environments. CI tests will not need a Redis instance, `firebase serve` will point at the deployed AP (with already configured Redis instance), and for the BE deploys, the Redis instance details are included in the gae yamls. I think that covers all of the bases.

Woot we're getting closer!

server/core/demo_data/revocations_matrix_filtered_caseload.metadata.json
daschi: Oh ok! I'll delete this, must've snuck in when I merged master.

server/core/fetchMetrics.js
```diff
@@ -44,41 +12,32 @@ const memoryCache = createMemoryCache(
...
-            );
-          });
+const { default: processMetricFile } = require("./processMetricFile");
+const { default: fetchMetricsFromLocal } = require("./fetchMetricsFromLocal");
+const { default: fetchMetricsFromGCS } = require("./fetchMetricsFromGCS");
 
-          console.log(`Fetched all ${cacheKey} metrics from ${source}`);
-          cacheCb(null, results);
-        })
-        .catch((err) => {
```
jovergaag, line 75: We definitely have some work to do in this repo around error handling. I just renamed one of the metric files in the bucket to test, and the API response is 200, but the 404 error and error message are embedded within the body of the response. We already have some issues on the board to work on error handling, responding here with a proper 404 will be first up. Removing this catch block seems ok to me for now since the error is wrapped in a rejected Promise.

server/routes/__tests__/api.test.js
```diff
@@ -23,11 +23,24 @@ const {
...
+jest.mock("../../core/redisCache", () => {
+  return {
+    cacheInRedis: jest.fn(),
+  };
+});
+jest.mock("../../core/createMemoryCache", () => {
+  return {
+    cacheInMemory: jest.fn(),
+  };
+});
```
macfarlandian, line 43: not to harp on the test thing with another comment but .... I had to do something somewhat similar in the spotlight api recently and what I decided to do was mock the cloud storage API rather than the cache library. then i could verify that repeat calls to the api function did not result in repeat cloud storage fetches without going into the cache implementation details. maybe something like that would work here? you'd probably have to mock the redis client also but at least it would integrate more application code

README.md
```diff
@@ -22,14 +22,18 @@ Bringing criminal justice analysis to decision makers to help reduce incarcerati
...
 
 1. Install Yarn package manager:
 
    `brew install yarn`
 
    For alternative Yarn installation options, see [Yarn Installation](https://yarnpkg.com/en/docs/install).
 
+1. Install Redis via Homebrew:
+
+   `brew install redis`
```
macfarlandian, line 35: is a particular version required or recommended? and do we also want to include non-mac instructions or at least make clear that these are platform-specific and link out to Redis install docs or something

server/core/redisCache.js
```diff
@@ -0,0 +1,75 @@
...
+});
+
+const redisCache = cacheManager.caching({
+  store: redisStore,
+  refreshThreshold: REDIS_CACHE_REFRESH_THRESHOLD,
+  redisInstance,
+});
+
+const redisClient = redisCache.store.getClient();
+redisClient.on("error", (error) => console.error("ERR:REDIS:", error));
```
jovergaag, line 56: Do we need to raise an error here in addition to logging it? What types of things constitute an error?

server/core/redisCache.js
```diff
@@ -0,0 +1,75 @@
...
+
+// Expire items in cache after 1 day
+const REDIS_CACHE_TTL_SECONDS = 60 * 60 * 24;
+// Set refresh threshold to 1 hour
+const REDIS_CACHE_REFRESH_THRESHOLD = 60 * 60;
+
+const redisInstance = new Redis({
+  host: REDISHOST,
+  port: REDISPORT,
+  password: REDISAUTH,
```
jovergaag, line 44: Did you add these secrets to the password vault?

README.md
```diff
@@ -118,17 +120,25 @@ To run eslint manually:
 
 > **Note**: we are gradually adding linting enforcement to our existing code, so not everything may be subject to linting yet. Refer to the [ignore file](https://github.com/Recidiviz/pulse-dashboard/.eslintignore) for details. While this transition is underway, if you are adding new files or substantially rewriting existing ones, you are encouraged to whitelist them for linting by updating the configuration file.
 
-### Running the application locally
+### Running the BackEnd application locally
+
+To run the backend locally, you will need to run it in demo mode. This can be done with the following yarn script:
```
macfarlandian, line 125: confused by this ... you don't _have_ to run it in demo mode, do you? i was able to run it just fine pointed at the staging bucket

macfarlandian (commented):
was able to run it just fine locally by following the readme. overall looks good, just a bunch of questions and mostly minor suggestions

server/core/refreshRedisCache.js
```diff
@@ -0,0 +1,65 @@
...
+    const cacheKey = `${cacheKeyPrefix}-${fileKey}`;
+    const metricFile = files[fileKey];
+    console.log(`Setting cache for: ${cacheKey}...`);
+    cachePromises.push(redisCache.set(cacheKey, metricFile));
+  });
+
+  return Promise.all(cachePromises);
+}
+
+function refreshRedisCache(fetchMetrics, stateCode, metricType, callback) {
```
jovergaag, line 43: Have you found a good way to trigger this when running it locally outside of tests?

jovergaag (changes requested):
This is amazing. It looks good and seems to be working great! I just have a few things.

1. the Redis instance name in GCP is currently `public-dashboard-instance-1`. We probably want to rename it or create another that is `pulse-dashboard-instance-1` :)
update: I just saw your comment above about the name. IMO its not necessary to change right now. Maybe we can keep it for now and once we get the staging implementation a little more built out - create a new instance to migrate over to to test that process out before deploying to production.
2. I would be curious to verify that the cache persists through a deploy. Would the best way to see this be to check the logs? Or via the Compute Engine VM you had started to play with? It would be really nice to have some visibility into the cache.
3. I'm seeing lots of `ERR:REDIS: { Error: connect ETIMEDOUT` errors in the logs from the redisClient.
4. Thinking through long term maintenance. If we ever wanted to make a change to the Redis instance, is this instance able to be edited to do things like increase the memory? Or would we need to create a new instance with more memory then deploy the BE pointing to the new instance?

Thats it for now. Seems like there are a few small kinks to work out but I'm so excited about this!

README.md
jovergaag: This is a good point. Vlad uses Windows. He is more than capable of figuring out the Windows install but it might be nice to include.

server/core/redisCache.js
daschi: Not yet, I was thinking of adding them after this merges, but if you think it makes sense to add now I can do that too!

server/core/redisCache.js
daschi: I've been thinking about this, I'm not sure if we should or not. Here's some things that constitute errors:

- TCP error (ECONNREFUSED) - The redis server is down, or some other reason a TCP connection was unsuccessful
- Redis ReplyError - Incorrect redis command, AUTH password is incorrect

In the first case for a TCP error, I don't know if we would want to raise an error, or have it default to use a memory cache if redis is not available?

server/core/refreshRedisCache.js
daschi: Yes you should be able to hit this just fine in the DEMO mode, is that what you mean? Otherwise, you can use the cron job to trigger it directly.

daschi:
> I would be curious to verify that the cache persists through a deploy. Would the best way to see this be to check the logs? Or via the Compute Engine VM you had started to play with? It would be really nice to have some visibility into the cache.

As far as I can tell from reading there's not a great way to inspect the cache, other than to export it and then import it to your local redis server. You can see the number of cache keys [available from the instance details ](https://console.cloud.google.com/memorystore/redis/locations/us-east1/instances/public-dashboard-instance-1/details?project=recidiviz-dashboard-staging) by clicked on the dropdown above the chart. 

You can also look at the redis logs using this query in the logs explorer, but I think this is only instance level logging (create/delete/check auth on instance)
```
resource.type="audited_resource"
protoPayload.serviceName="redis.googleapis.com"
```

I didn't go forward with creating a VM instance to test this, but I think it could be a good idea, at least while we're setting this up. I'll try to do that tomorrow. 

> I'm seeing lots of ERR:REDIS: { Error: connect ETIMEDOUT errors in the logs from the redisClient.
Are you seeing these locally, or do you mean in the GCP logs?

>Thinking through long term maintenance. If we ever wanted to make a change to the Redis instance, is this instance able to be edited to do things like increase the memory? Or would we need to create a new instance with more memory then deploy the BE pointing to the new instance?

We can increase the capacity, turn no/off AUTH requirement, and configure the `maxmemory-policy` configurations (currently set to kick out the least frequently used keys). We can't change the instance name (🤦‍♀️ ), edit TLS settings, or update the "tier" (currently on Basic).

server/core/redisCache.js
jovergaag: Doing it when you merge it sounds great. We just need to make sure that whoever deploys prod has downloaded them beforehand.

server/core/redisCache.js
jovergaag: I think we should raise an error, at least for the short term in staging just to see what errors we are receiving.

I found that you can opt in to being notified via email for the project here: https://console.cloud.google.com/user-preferences/communication

jovergaag: If it gets to be too verbose we can always change it.

jovergaag:
> Are you seeing these locally, or do you mean in the GCP logs?

I saw them in the GCP logs on Wednesday.

> We can increase the capacity, turn no/off AUTH requirement, and configure the maxmemory-policy configurations (currently set to kick out the least frequently used keys). We can't change the instance name (🤦‍♀️ ), edit TLS settings, or update the "tier" (currently on Basic).

I think its fine. We can always migrate to a new instance with a different name later if it becomes a problem. Good to know we can increase the capacity on the same instance!

jovergaag:
@daschi I've been running some performance analysis on this. IE11 has increased memory use on initial load quite a bit. Looking at the network responses in Chrome, the file sizes seem to be about double. What??!? I'm doing some digging to see if I can find why those files are getting bigger...

RTT for the requests is also longer, probably due mostly to the larger file sizes. Once we figure out the file sizes, we can do some analysis to figure out how much time the external cache access is adding.

jovergaag:
@daschi I changed my settings in GCP to receive notifications from recidiviz-dashboard-staging, but I haven't received any email notifications of the errors that we saw in the logs today.

> I've been running some performance analysis on this. IE11 has increased memory use on initial load quite a bit. Looking at the network responses in Chrome, the file sizes seem to be about double. What??!? I'm doing some digging to see if I can find why those responses are getting bigger...
> 
> RTT for the requests is also longer, probably due mostly to the larger file sizes. Once we figure out the file sizes, we can do some analysis to figure out how much time the external cache access is adding.

For anyone following along - we confirmed today that the increase in response size was due to differences in data between prod and staging. After copying the prod metric files to staging the response data size is the same 🎉 

The response time, however, is ~25% longer due to the external cache access. However, I don't think that's a big deal because all of the files are requested in parallel. The blocking time on the initial load is still about the same. I'm comfortable with this. LMK if anyone has a concern here!

server/routes/__tests__/api.test.js
daschi: @macfarlandian We decided to move this work into a new branch / PR!

server/core/redisCache.js
jovergaag: After discussion with @daschi today, we have decided only to log the error and not to raise it at this point. I have created this issue #683  to notify Sentry directly from the BE when errors occur.

server/core/redisCache.js
```diff
@@ -0,0 +1,75 @@
...
+        .catch((err) => {
+          console.error(err);
+          cacheCb(err, null);
+        });
+    },
+    callback
+  );
+}
+
+module.exports = { cacheInRedis, redisCache };
```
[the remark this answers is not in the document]
daschi: I'm using `redisCache` in the `refreshRedisCache` function to access the `set` method, since we always want to override what's in the cache for that endpoint.

jovergaag (changes requested):
Just a couple of small updates then its all set!

jovergaag (approved):
SWEET. Nice work. Very exciting to have this in and working so well! 

🔴 🥏 

Before you move off of Lantern, let's set up the production instance together so I can learn the steps.

Don't forget to add the secrets for the staging instance to the vault :)

jovergaag:
This still looks good after those last commits!

PR #672, opened 2020-12-18
Supports new level_1_supervision_location and level_2_supervision_location dimensions

### LINKED ISSUE #667: Modify charts and filters to support level_1_supervision_location and level_2_supervision_location dimensions
There is a BE change coming down the road that will add two dimensions to each metric file: `level_1_supervision_level` and `level_2_supervision_level`. We need to adjust the district chart, and filtering logic to be able to support BOTH the current metric files, and the new metric files such that when that BE change lands it does not break the FE. 

We will follow up with another PR to point the FE towards these new dimensions and away from the district dimension as cleanup. The BE can then follow and remove the district dimension.

So the process for this will be:
*FE PR (this issue) is merged and deployed to prod
*This BE PR is merged and deployed
*FE PR to remove use on district dimension entirely.
*BE PR to remove the district dimension

**Additional context**

jovergaag:
## Description of the change

We are moving toward deprecating the district dimension in all metric files. This PR is the first step towards supporting that. We decided to do this in two FE steps to make it a non-breaking change, allowing the BE and FE to merge their respective PRs at their own leisure.

The `levelTwoSupervisionLocation` filter and `levelTwoSupervisionLocation` are not actually enabled yet, at this point only the district filter is currently enabled. I wrote that filter code to make sure that this approach would easily support moving to those dimension filters when we needed to move, and it does. That means that the following PR will be quite a bit smaller.

Also includes fixture data including these new dimensions.

## Type of change

## Related issues

> Closes #667

### Development

### Code review


src/components/charts/new_revocations/__tests__/helpers.test.js
```diff
@@ -296,6 +304,134 @@ describe("applyTopLevelFilters", () => {
...
+          skippedFilters: [],
+          treatCategoryAllAsAbsent,
+        })(data);
+        filteredSupervisionLocations = filtered.map(
+          (f) => f.level_1_supervision_location
+        );
+      });
+
+      it("returns all of the rows", () => {
+        const expected = ["ALL", "04B", "ALL", "05X"];
```
macfarlandian, line 366: i am having trouble wrapping my brain around this ... why is "ALL" simultaneously "absent" and present?

jovergaag: You're right, that's not a realistic scenario. I was just trying to show that when `treatCategoryAllAsAbsent` is true, it passes through all of the values no matter how the filter is set. I updated the test to with more realistic data.

macfarlandian: ok makes sense! thanks

macfarlandian (approved):

Issue #657, opened 2020-12-16
[API-FE] Build a query string based on the current filter state, 


jovergaag:
**What needs to be done? Why does it need to be done?**
Build a query string based on the current filter state. Add the query string to the API request URI.

Match the params that were established in #653

**Additional context**
Add any other context about the task here.

PR #679, opened 2020-12-22
[API] Add path and query param validations for new revocations endpoint 

### LINKED ISSUE #653: [API] Verify query params on the `newRevocations` endpoint and respond with 400 if invalid
Accept a query string on the `newRevocations` endpoint and verify the query params (detail below). If the params or param values are invalid, respond with a 400 error.

The params and values are below. The `district` dimension has dynamic values so do not verify that value.
chargeCategory: ["alcohol_drug", "all", "domestic_violence", "general", "serious_mental_illness", "sex_offense"]
district: <any>, 
metricPeriodMonths: ["12", "3", "36", "6"]
reportedViolations: ["1", "2", "3", "4", "5", "6", "7", "8", "all"]
supervisionLevel: ["all", "electronic_monitoring_only", "enhanced", "high", "incarcerated", "maximum", "medium", "minimum", "special"]
supervisionType: ["all", "dual", "parole", "probation"]
violationType: ["absconded", "all", "elec_monitoring", "escaped", "felony" "high_tech", "law", "low_tech", "med_tech", "misdemeanor", "municipal", "no_violations", "substance_abuse", "technical"]

**Additional context**

daschi:
## Description of the change

This PR adds path and query param validations for the `/api/:stateCode/newRevocations/:filename` endpoint. This is the first step towards requesting files by filter params from the frontend, and the param validations are set to optional here until we add the query params to the frontend as part of #657. 

I chose to use the package `express-validator` because I thought the documentation was pretty good and it's straightforward, but let me know if there is a different validation package you prefer!

## Type of change

## Related issues

> Closes #653 

Relates to #657 

### Development

### Code review


server/routes/api.js
```diff
@@ -20,22 +20,25 @@
...
 const { default: isDemoMode } = require("../utils/isDemoMode");
-
 /**
  * A callback which returns either either an error payload or a data payload.
  */
 function responder(res) {
   return (err, data) => {
     if (err) {
-      res.send(err);
+      const status = err.status || 500;
```
jovergaag, line 32: hooray. Thank you.

public-dashboard is doing this: `res.status(500).json({ error: err.message });` which i think does a nice job of tidying up the message. Currently the body of the response is pretty hard to read so I think we should adopt this.

So, when I renamed a file in the bucket, the error comes through here with `{ code: 404 }`, so the status here is set to 500 to send to the FE. That's a huge improvement from the status 200 we were receiving :). I wonder if we should also check for that `code: 404` here and set the status to 404 as well?

server/routes/api.js
```diff
@@ -20,22 +20,25 @@
...
+      res.status(status).send(err);
     } else {
       res.send(data);
     }
   };
 }
 
+const BAD_REQUEST = 400;
```
jovergaag, line 40: Should we add a 500 here as well, then include it on line 32?

server/routes/api.js
jovergaag: Or just check the `code` in general and set the status accordingly. Why is it coming through as `code`? 🤔

jovergaag (changes requested):
The param validation all looks good. Since you started digging into it...a couple of small requests on the error statuses :)

server/routes/api.js
daschi: Yeah that's a good idea, I'll add the check for `code`.[ Here's the documentation for the json API responses for google cloud storage.](https://cloud.google.com/storage/docs/json_api/v1/status-codes#401-unauthorized)

jovergaag (approved):
Supa! 🚀

PR #680, opened 2020-12-24
Move filters to FiltersStore

### LINKED ISSUE #644: [API-FE] Move filters to MobX store
Install MobX and move the filters to a FiltersStore such that any updates to the filters are run in an action to update the FiltersStore

**Additional context**

jovergaag:
## Description of the change

Moves the filters and filterOptions to a central Store. This will make it easier to trigger a fetch if the currently cached dimension values for any given file do not fulfill the new filter state.

Eventually, there will be a `DataStore` action which will be called from FiltersStore.setFilters() to evaluate the local data/trigger the fetch. We could have put that logic directly in the Revocations component where the filters currently live, but this seems much more clean.

TODO:
- [x] find a better way to set the default tenant Id in `TenantStore`
- [ ] ~~make a StoreProvider wrapper to be used in tests to make testing components that access the store less cumbersome~~
- [x] final QA round with all filters and charts enabled
- [x] final IE11 performance analysis

## Type of change

## Related issues

> Closes #644

### Development

### Code review


src/App.js
```diff
@@ -59,15 +60,18 @@ const App = () => (
...
-            <LanternLayout stateCode={lanternTenant.MO}>
-              <Switch>
-                <Route path="/community/revocations" component={Revocations} />
-                <Route path="/profile" component={Profile} />
-                <Redirect exact from="/" to="/community/revocations" />
-                <Redirect from="/revocations" to="/community/revocations" />
-                <NotFound />
-              </Switch>
-            </LanternLayout>
+            <StoreProvider stateCode={lanternTenant.MO}>
```
jovergaag, line 63: I don't love wrapping both of these `LanternLayouts` individually but we need the stateCode. Until we get around to adding a `UserStore` where we can pull the stateCode from, this seemed like the best approach.

src/App.test.js
```diff
@@ -53,7 +53,7 @@ describe("App tests", () => {
 
   LanternLayout.mockImplementation(({ children }) => children);
   CoreLayout.mockImplementation(({ children }) => children);
-  Revocations.mockReturnValue(mockWithTestId(mockRevocationsId));
+  Revocations.type.mockReturnValue(mockWithTestId(mockRevocationsId));
```
jovergaag, line 56: This was necessary because Revocations is now a wrapped function (observer)

src/App.js
macfarlandian: a few thoughts; not sure if any of them will prove helpful: 

- if `stateCode` was an observable property rather than a constructor argument, would that help? (i.e. do you ever really need to render both of these layouts at the same time? I don't really know how this part of the app works so sorry if that's a useless question)
- alternatively, what if you had a single root store (and thus a single context/provider) that had multiple "tenant stores" or whatever it would make sense to call them? within each one you could use the state code to look up the matching tenant. I think in general it will probably make our lives easier to just have a single root store with everything in it, so it might be worth it to figure out how to have that now even if it's a little weird with pieces missing

macfarlandian (commented):
I ... think I follow? I can already see how much of an improvement this will be. Overall I think it makes sense beyond any specific comments below. 

One additional thing I would strongly recommend is that you enable the MobX runtime linting in both your test and application environments and see what warnings it throws. They are kind of annoying (in large part because they don't have stack traces, which drove me nuts) but I found they helped me understand the flow of data better.

jovergaag:
@macfarlandian Thank you!

> One additional thing I would strongly recommend is that you enable the MobX runtime linting in both your test and application environments and see what warnings it throws. They are kind of annoying (in large part because they don't have stack traces, which drove me nuts) but I found they helped me understand the flow of data better.

I have them in my index locally but there are SO MANY ERRORS right now that it crashes the browser, so I have them commented out for the moment. Once I can access the filters in the charts and filter components directly, and add an `observer` then it should quiet down and I will enable the linters.

> I think in general it will probably make our lives easier to just have a single root store with everything in it, so it might be worth it to figure out how to have that now even if it's a little weird with pieces missing

Good point. I'll spend some more time on that before going any further.

src/tenants.js
```diff
@@ -23,14 +23,6 @@ export default {
   [lantern.US_MO]: {
     name: "Missouri",
     availableStateCodes: [lantern.US_MO],
-    regions: {
```
jovergaag, line 26: This was added for the district restricted access feature that was prematurely coded quite a while ago - that feature is coming up this week but we have decided on a different approach than these regions.

src/utils/authentication/user.js
```diff
@@ -74,23 +74,3 @@ export function getAvailableStateCodes(user) {
 export function doesUserHaveAccess(user, stateCode) {
   return getAvailableStateCodes(user).includes(stateCode);
 }
-
-/**
- * Returns the district or districts that a user should be limited
- * to viewing data for, if there is user metadata in Auth0 specifying
- * a district or region (group of districts) for the user.
- */
-export function getUserDistricts(user) {
```
jovergaag, line 83: Same here - the implementation of the district restricted access has changed.

src/components/Revocations.js
```diff
@@ -115,83 +80,45 @@ const Revocations = () => {
           <div className="top-level-filters d-f">
             <ToggleBarFilter
               label="Time Period"
-              value={filters[METRIC_PERIOD_MONTHS]}
-              options={filterOptions[METRIC_PERIOD_MONTHS].options}
-              defaultOption={filterOptions[METRIC_PERIOD_MONTHS].defaultOption}
-              onChange={createOnFilterChange(METRIC_PERIOD_MONTHS)}
```
jovergaag, line 121: I would love feedback on this approach of accessing the values and options directly in the component via the store vs via the props. It definitely has changed how the filters are tested. Is this a step backward composition wise?

src/components/Revocations.js
```diff
@@ -43,64 +42,30 @@ import RevocationsByRace from "./charts/new_revocations/RevocationsByRace/Revoca
...
-  };
-
-  const createOnFilterChange = useCallback(
-    (field) => (value) => {
-      setFilters({ ...filters, [field]: value });
-    },
-    [filters]
-  );
-
-  const transformedFilters = limitFiltersToUserDistricts(
```
jovergaag, line 100: This is removed because the implementation of the District Restricted Access feature will be different than what was initially planned months ago when it was originally written (but not yet ready).

jovergaag:
@macfarlandian This is nearly ready for a final review. I am working on a few more tests and finding a better way to wrap components in a `StoreProvider` in the tests that is not so cumbersome, as well as a better way to set the default Tenant in the `TenantStore`.

I wanted to get some feedback while I'm finishing up these remaining items. Thoughts on the overall implementation?

src/RootStore/FiltersStore.js
```diff
@@ -0,0 +1,88 @@
...
+  get filterOptions() {
+    return filterOptionsMap[this.rootStore.currentTenantId];
+  }
+
+  setFilters(updatedFilters) {
+    this.filters = { ...this.filters, ...updatedFilters };
+  }
+
+  setRestrictedDistrict(restrictedDistrict) {
+    this.restrictedDistrict = restrictedDistrict;
```
daschi, line 85: For the `restrictedDistrict` feature -- does this mean that there are some users that can only see data for 1 district?

src/RootStore/TenantStore.js
```diff
@@ -38,21 +39,30 @@ function getTenantIdFromUser(user) {
...
-    makeAutoObservable(this);
+    makeAutoObservable(this, {
+      setCurrentTenantId: action,
+    });
+
+    this.rootStore = rootStore;
 
+    // TODO create a UserStore and setCurrentTenantId on line 63
+    // as a reaction to the user being updated, rather than setting
+    // default currentTenantId on line 49
```
daschi, line 62: Could this similarly also apply to setting the `restrictedDistrict` from the user in the `FiltersStore`?

src/components/charts/new_revocations/RevocationsOverTime.js
```diff
@@ -192,19 +187,20 @@ const RevocationsOverTime = ({
...
       timeDescription={getTrailingLabelFromMetricPeriodMonthsToggle(
-        metricPeriodMonths
+        filters.metricPeriodMonths
       )}
       labels={chartLabels}
       chartId={`${translate("revocations")}OverTime`}
       datasets={datasets}
       metricTitle={translate("revocationsOverTimeXAxis")}
-      filterStates={filterStates}
+      filterStates={filters}
```
daschi, line 203: I think `filterStates` was removed as a prop for this component?

src/components/Revocations.js
daschi: I think I like this change for the filter components. `ToggleBarFilter` now has a very clear/consistent onChange, and the other filters contain the code specific to their use cases. This pattern would differ from how we talked about passing in the `apiData` from the upcoming `DataStore`, but I think we have a good reason to pass in the data as a prop, and I'm not sure I know the drawbacks yet of accessing the filters from the store per component. Happy to keep chatting about it if you want to think through the idea more too..

src/RootStore/FiltersStore.js
jovergaag: Yes exactly. There will be a file mapping email addresses to district number, and the district filter will be locked to that district.

src/RootStore/TenantStore.js
jovergaag: Yes!

src/components/charts/community/RevocationCountOverTime.js
```diff
@@ -253,4 +254,4 @@ RevocationCountOverTime.propTypes = {
   disableGoal: PropTypes.bool,
 };
 
-export default RevocationCountOverTime;
+export default observer(RevocationCountOverTime);
```
macfarlandian, line 257: what observable data is being used here?

src/components/charts/new_revocations/CaseTable/CaseTable.js
```diff
@@ -101,7 +100,7 @@ const CaseTable = ({ dataFilter, filterStates, metricPeriodMonths }) => {
 
   return (
     <CaseTableComponent
-      filterStates={filterStates}
+      filters={filters}
```
macfarlandian, line 103: `filters` is an observable data structure, so `CaseTableComponent` needs to be an `observer` as well, right? I would even go further and say it ought to just consume the filters directly from the context rather than having it passed down like this. I think this pattern can cause unnecessary re-renders of this parent component when all it's doing is passing new data to its children

src/components/charts/new_revocations/Matrix/Matrix.js
```diff
@@ -159,7 +157,7 @@ const Matrix = ({
           labels={VIOLATION_COUNTS.map(violationCountLabel)}
           metricTitle={TITLE}
           timeWindowDescription={timeDescription}
-          filters={filterStates}
+          filters={filters}
```
macfarlandian, line 160: I don't think you need to pass this anymore? isn't `ExportMenu` getting this from context now?

src/components/charts/new_revocations/ToggleBar/__tests__/ToggleBarFilter.test.js
```diff
@@ -13,54 +42,74 @@ jest.mock("../FilterField", () => ({
...
+        </StoreProvider>
+      );
+
+      act(() => {
+        Select.mock.calls[0][0].onChange(
+          rootStore.filtersStore.filterOptions[props.dimension].options[3]
+        );
+      });
+
+      reaction(
```
macfarlandian, line 104: similar to the other tests I think maybe `when` would be a better option here?

macfarlandian (changes requested):
looking really good overall! love how much prop drilling you've been able to eliminate from these components.

src/RootStore/FiltersStore.js
```diff
@@ -0,0 +1,88 @@
...
+    makeAutoObservable(this, {
+      defaultFilters: computed,
+      filterOptions: computed,
+      setFilters: action,
+    });
+
+    this.rootStore = rootStore;
+
+    autorun(() => {
+      this.setFilters(this.defaultFilters);
```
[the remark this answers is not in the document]
jovergaag: Yes exactly.

src/RootStore/TenantStore.js
```diff
@@ -38,21 +39,30 @@ function getTenantIdFromUser(user) {
...
-  currentTenantId =
-    sessionStorage.getItem(CURRENT_TENANT_IN_SESSION) || "US_MO";
+  currentTenantId = LANTERN_TENANTS[0];
+
+  user;
 
   constructor({ rootStore }) {
-    makeAutoObservable(this);
+    makeAutoObservable(this, {
+      setCurrentTenantId: action,
```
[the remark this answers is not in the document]
jovergaag: Yep! Good catch.

src/components/__tests__/Revocations.test.js
```diff
@@ -1,5 +1,23 @@
+// Recidiviz - a data platform for criminal justice reform
+// Copyright (C) 2020 Recidiviz, Inc.
```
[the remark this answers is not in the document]
jovergaag: uff. Indeed it is.

src/components/charts/community/RevocationCountOverTime.js
jovergaag: Good catch. We have lots of components that are named nearly identically unfortunately. I must have added this thinking it was `RevocationsOverTime` by mistake.

src/components/charts/new_revocations/CaseTable/CaseTable.js
jovergaag: This is a mistake. CaseTableComponent does not even need filters!

src/RootStore/FiltersStore.js
```diff
@@ -0,0 +1,88 @@
...
+  SUPERVISION_LEVEL,
+  SUPERVISION_TYPE,
+  VIOLATION_TYPE,
+} from "../constants/filterTypes";
+import filterOptionsMap from "../views/tenants/constants/filterOptions";
+
+export default class FiltersStore {
+  rootStore;
+
+  filters = observable.map();
```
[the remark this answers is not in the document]
jovergaag: I struggled with this a lot. I started with it initialized as an object but get this linting warning: `[mobx] Observable FiltersStore@5.filters.isMobXObservableMap? being read outside a reactive context`. It seems like Mobx sees it as an observable.map. 

Screen Shot 2021-01-12 at 12 03 41 PM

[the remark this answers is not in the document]
jovergaag: I think because different components are observing specific attributes of the filters object, it has to be an observable.map?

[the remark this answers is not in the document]
jovergaag: I will play around with making filters a Map rather than an object.

src/components/__tests__/Revocations.test.js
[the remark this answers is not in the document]
jovergaag: I'm going to change these in another PR so that it doesn't clutter this PR.

src/RootStore/FiltersStore.js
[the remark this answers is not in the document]
daschi: I was wondering this too and I read this in the [API docs](https://mobx.js.org/api.html#observablemap):

`Creating observable Maps is the recommended approach for creating dynamically keyed collections if you don't have enabled Proxies.`

So I thought we were using it because IE11 doesn't support enabled proxies.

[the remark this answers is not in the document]
macfarlandian: weird ... I think there's probably a way to leave it as an object if that makes things easier? maybe it could at least be an `observable.object` to make it less confusing? I am still a bit fuzzy about all of this honestly so if it works and is reasonably readable then I don't think we need to be too precious about it

src/components/__tests__/Revocations.test.js
```diff
@@ -130,103 +139,27 @@ describe("Revocations component tests", () => {
...
-      });
-    });
-
-    expect(ViolationFilter.mock.calls[1][0].violationType).toBe(
-      mockNewViolationTypeValue
-    );
-    expect(ViolationFilter.mock.calls[1][0].reportedViolations).toBe(
-      mockNewReportedViolationsValue
+    expect(getByTestId(districtFilterId)).toBeInTheDocument();
+    reaction(
```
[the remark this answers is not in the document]
jovergaag: Not overkill! In fact it turns out that the expect in this test as well as the other test with this pattern were never running at all. I ended up approaching these tests completely differently. I still don't love them but its at least now doing what i think its doing :)

jovergaag:
@macfarlandian and @daschi  This is ready for another look

> If you are concerned about the tight coupling between UI and data source, maybe you could preserve the original filter components but wrap each one in a container component that interacts with the data store and passes it as props? I don't think this is all that critical, considering that all they're doing is reading the data rather than applying any logic to it, but if portability is a big concern then maybe that would help?

@daschi and I have been talking about the container concept over the last few days. It would remove our callback scope issue  which is causing charts to hold on to memory after being unmounted. Ultimately we decided to move ahead with a DataStore performing the fetch instead of the chart which also solves the scope problem. We decided that the container is just going to be a lot of extra code but provides no benefit over what the DataStore provides. I'm inclined to go this route and see how it feels after the DataStore is in and we can make adjustments after that if it still seems too tightly coupled.


I think I'm going to leave the StoreProviderWrapper for the tests for another day. There is a lot of work blocked behind this so the sooner we can get it in the better. It will also allow us more experience working with the observer components in the tests to get a feel for how best to implement the wrapper.

src/components/__tests__/Revocations.test.js
```diff
@@ -130,103 +138,21 @@ describe("Revocations component tests", () => {
...
-    );
-    expect(ViolationFilter.mock.calls[0][0].violationType).toBe(
-      filterOptions[VIOLATION_TYPE].defaultValue
-    );
-  });
-
   it("should set user district as default filter value if it is defined", () => {
-    getUserAppMetadata.mockReturnValue({ district: mockDistrict });
+    const setRestrictedDistrictSpy = jest.fn();
+    FiltersStore.prototype.setRestrictedDistrict = setRestrictedDistrictSpy;
```
macfarlandian, line 143: would it be better to mock FiltersStore here instead? I am worried this will pollute other test suites. 

(same applies to the other test where a similar pattern was implemented)

jovergaag: I had originally tried to set up a `__mocks__/FiltersStore` but no matter how I set that up I wasn't able to make it work for both test files. This updated mock feels a little clunky but I agree that its safer - in fact I think that spy was picking up another call in the ToggleBarFilter test.

macfarlandian (approved):
🥇

jovergaag:
@macfarlandian @daschi Based on your comments above about the `observable.map` - I dug into this a little more. As it turns out, the filters was an observable map, but the filter values that were being set weren't individually observable.

Here is the `filters` object before my last two commits - note that the filter keys are at the top level of the object (not inside `data`, and are not observable:
Screen Shot 2021-01-13 at 1 12 35 PM

Here is the `filters` object after the most recent commits - note that the filter keys are inside of `data` and are each observable
Screen Shot 2021-01-13 at 1 11 53 PM


Notice that the values in the filters object are now each observable - which I think is what we want. What we were doing before was working because the filters would run when the `filters` object would change, but now each of the filter values themselves is observable - useful for things like the `timeDescription` in each chart.

I also found that the `ExportMenu` exports were broken so I fixed that in the final commit.

A final review of those last two commits is appreciated!

src/components/charts/new_revocations/RevocationsOverTime.js
```diff
@@ -45,11 +46,12 @@ import { generateTrendlineDataset } from "../../../utils/charts/trendline";
 import { translate } from "../../../views/tenants/utils/i18nSettings";
 import RevocationsByDimensionComponent from "./RevocationsByDimension/RevocationsByDimensionComponent";
 import { useRootStore } from "../../../StoreProvider";
+import { METRIC_PERIOD_MONTHS } from "../../../constants/filterTypes";
 
 const RevocationsOverTime = ({ dataFilter }) => {
   const { filters, currentTenantId } = useRootStore();
 
-  const chartId = `revocationsOverTime`;
+  const chartId = `${translate("revocations")}OverTime`;
```
jovergaag, line 54: While working on this I found a bug - so I fixed it here. Meant to put it in its own commit but...such is life. Issue #689

src/RootStore/FiltersStore.js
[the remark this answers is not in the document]
jovergaag: I tried to make this an `observable.object` but nothing I did made Mobx happy. I'm ok with it being an `observable.map` if you all are... at least until we better understand the nuances of these different observables.

macfarlandian (approved):
I think the nuance of this approach may have gone slightly over my head, but the changes seem reasonable to me. I say let's give it a shot!

PR #681, opened 2020-12-29
TenantStore instead of StateCodeContext

jovergaag:
## Description of the change

As we move toward moving the filters to a MobX store, an interim step is to add a TenantStore which manages the `currentTenantId`. This allows both of the layouts to use the same store. There will be another PR to follow which adds the filters to the FilterStore (see WIP PR #680). It was feedback in that PR that led me to first add the TenantStore here.

This PR also changes all of the tenant IDs to be "US_MO", "US_PA", or "US__PA" instead of lower case or missing the "US_" prefix.

This PR is fairly large, but following it commit by commit should make it easier to digest.

## Type of change

## Related issues

> Part of #644

### Development

### Code review


src/components/TenantRoutes.js
```diff
@@ -42,24 +42,20 @@ const TenantRoutes = ({ children }) => {
...
-  if (!isAuthenticated || !user || !currentStateCode) {
+  if (!isAuthenticated || !user || !currentTenantId) {
     return null;
   }
 
   return Children.toArray(children).reduce((node, child) => {
-    const { stateCode } = child.props;
-
-    if (doesUserHaveAccess(user, stateCode) && stateCode === currentStateCode) {
+    const { tenantIds } = child.props;
```
ohaibbq, line 54: Where does this prop come from? I couldn't find its definition in Github search.

src/components/charts/new_revocations/ToggleBar/DistrictFilter.js
```diff
@@ -33,13 +34,15 @@ import {
...
 
 const allDistrictsOption = { label: "All", value: "All" };
 
-const DistrictFilter = ({ value, stateCode, onChange }) => {
+const DistrictFilter = ({ value, onChange }) => {
+  const { currentTenantId } = useRootStore();
   const { user } = useAuth0();
   const { isLoading, apiData } = useChartData(
-    `${stateCode}/newRevocations`,
+    `${currentTenantId}/newRevocations`,
```
ohaibbq, line 45: Do these URLs support the new uppercased names?

ohaibbq (approved):
Nice work removing `stateCode` from all the component props! 💯

A couple questions from me, but otherwise it looks great!

src/components/TenantRoutes.js
jovergaag: This is coming from the layouts. If you look in App.js, both the LanternLayout and CoreLayout have a tenantId prop - which is basically a list of tenants that should use that layout. You prob didn't find it in a Github search because its new in this branch.

src/components/charts/new_revocations/ToggleBar/DistrictFilter.js
jovergaag: yes - In fact @daschi 's param validation PR is up-casing the state_code param when it receives it and is caching it using the upper case state-code.

https://github.com/Recidiviz/pulse-dashboard/pull/679/files

jovergaag:
Note: This has been tested thoroughly on IE11, as well as all auth routes for different users since it touched the tenant routing.

Issue #660, opened 2020-12-16
[API-FE] Clear the cache and memory for any given data when the metric file is requested


jovergaag:
**What needs to be done? Why does it need to be done?**
Ensure that the in memory cache and chart references for any given data is cleared when that metric file is requested. 

**Additional context**
Add any other context about the task here.

jovergaag:
Closed by #694

PR #693, opened 2021-01-15
Memory optimization changes for FE charts

daschi:
## Description of the change

These changes were originally implemented in the branch of PR #694 as we were trying to figure out why the charts were not releasing references to the old data as new data from a filter change triggered fetch was passed to the chart. Between this work and the data store, the charts are now completely releasing and reloading new data, yay!

This work can be merged separately from the DataStore work in PR #694 so it was cherry-picked into a separate branch. But it is all closely related in terms of motivation.

Included changes:

- Return `apiData` from `useChartData` as either the optimized format or unoptimized format, and pass it to `filterOptimizedDataFormat` function. `filterOptimizedDataFormat` now has logic to figure out what format the data is in and filter appropriately.
- Move `currentDistricts` param for `generateChartData` to `createGenerateChartData`  for `RevocationsByDistrict`
- Remove the ticks and title callbacks from `RevocationsOverTime` and set maxValue for the chart to 7 if less than 7

## Type of change

## Related issues

> Related to #660 

### Development

### Code review


src/components/charts/new_revocations/Matrix/Matrix.js
```diff
@@ -69,12 +69,11 @@ const Matrix = ({ dataFilter, timeDescription }) => {
 
   const violationTypes = translate("violationTypes");
 
-  const { apiData, isLoading, isError, unflattenedValues } = useChartData(
```
jovergaag, line 72: We were storing two copies of the data in the charts, one as the `apiData.flattenedValueMatrix` which is a string representation of the data, and the `unflattenedValues` which is the 2d array of data. Now we are only keeping the data and the metadata. The data is passed as  `apiData` despite whether it is the optimized version of the data or whether it is the already expanded version.

src/components/charts/new_revocations/RevocationsOverTime.js
```diff
@@ -152,9 +138,6 @@ const RevocationsOverTime = ({ dataFilter }) => {
     tooltips: {
       backgroundColor: COLORS["grey-800-light"],
       mode: "x",
-      callbacks: {
```
jovergaag, line 155: These custom tooltips were not releasing memory when new data was passed into the chart. It turns out that the Chart.js standard tooltip is exactly the same as the custom tooltip we were creating here so we just removed the custom tooltip altogether to avoid memory buildup when the charts receive new data.

src/components/charts/new_revocations/RevocationsOverTime.js
```diff
@@ -138,12 +130,6 @@ const RevocationsOverTime = ({ dataFilter }) => {
           },
           ticks: {
             min: 0,
-            callback(value) {
```
jovergaag, line 141: Similar to the custom tooltip below, this custom tick was not releasing memory when new data was passed into the chart after filter change (see PR #694). The custom tick is ensuring that the y-axis value is always a whole number when the scale is smaller. But making the `maxValue` change on line 109, we eliminate the need for this custom tick and allow the charts to release memory when given a new set of data.

src/hooks/useChartData.js
```diff
@@ -119,16 +118,13 @@ function useChartData(url, file, eagerExpand = true) {
       })
       .finally(() => {
         setAwaitingApi(false);
+        cancel();
       });
-
-    return () => {
-      cancel();
-    };
```
jovergaag, line 126: We were seeing logs of un-cancelled promises keeping a reference to the old data after new data was fetched and passed to the charts (per filter fetching implemented in PR #694). Canceling the promise in the `finally` is one more change which allows that memory to be fully released on filter change.

macfarlandian: don't you still also want to return the cleanup function to run if the component unmounts while a promise is pending?

jovergaag: Yes!

macfarlandian (approved):

jovergaag:
@daschi Do you want to try to get your DataStore in before this? A lot of this is no longer applicable with the DataStore...I'll let you decide which should go first.

daschi:
@jovergaag My DataStore PR is actually branched off of this, so I think it would be easier to merge this first and then I should have very few conflicts.

PR #694, opened 2021-01-15
Add DataStore for new revocation charts

### LINKED ISSUE #657: [API-FE] Build a query string based on the current filter state,

### LINKED ISSUE #686: [API-FE] Implement that datastore for the API data
We need to implement a datastore for the API data so that every chart component receives its data as a prop. This is so we can improve the memory, since all anonymous functions defined in the same scope as the `useChartData` hook (which defines apiData) ends up holding onto a reference of the large data object. This solution will separate those definitions and should lower the retained memory size. 

**Additional context**

daschi:
## Description of the change

This PR adds two new features:

- Implement a DataStore for the MO and PA dashboard charts so that data fetching and state management is taken out of the components and out of useEffect hooks. 
- Request a new dataset on every filter change

There is a *lot* of duplication in this PR for each DataStore. I decided to spell everything out at this stage because I'm still learning mobx and wanted to get something working to test the IE11 memory performance before DRYing it up. 

This adds the following mobx data stores:

- DataStore - Parent store that initializes the other datastores and makes them available through the useRootStore context hook
- CaseTableStore
- MatrixStore
- RevocationsOverTimeStore
- RevocationsChartsStore - This is one store for the revocations charts since we only want to store one chart's data at a time in the app state

All of the chart datastores should be more or less the same with some small differences in the `filteredData` getter. The `RevocationsChartsStore` has a slightly different `fetchData` function since it fetches data based on the selected chart.

TODO: 
- [x] Use the DataStore for the DistrictFilter
- [x] IE11 performance testing with production build
- [x] Add tests for the new stores

## Type of change

## Related issues

> Closes #686 
> Closes #657 

### Development

### Code review


src/RootStore/DataStore/RevocationsOverTime.js
```diff
@@ -0,0 +1,141 @@
...
+
+    reaction(
+      () => getQueryStringFromFilters(this.queryFilters),
+      (queryString) => this.fetchData(queryString)
+    );
+  }
+
+  get queryFilters() {
+    return Object.fromEntries(toJS(this.filtersStore.filters));
+  }
```
macfarlandian, line 88: why doesn't this just call `getQueryStringFromFilters`?

src/RootStore/RootStore.js
```diff
@@ -15,24 +15,45 @@
...
 
+  auth0Context = observable.map({ loading: true });
+
   constructor() {
-    makeObservable(this, { filters: computed, currentTenantId: computed });
+    makeObservable(this, {
+      filters: computed,
+      currentTenantId: computed,
+      setAuth0Context: action,
+    });
```
macfarlandian, line 41: i think you should include `getTokenSilently` here too as a `computed`?

src/api/metrics/urlHelpers.js
```diff
@@ -0,0 +1,37 @@
...
+// =============================================================================
+import qs from "qs";
+
+/**
+ *
+ * @param {Object} filters - The filter values used to construct the query string to request metric data
+ * @param {string} filters.chargeCategory - A charge category or "All"
+ * @param {Array} filters.district - District IDs or "All"
+ * @param {string} filters.metricPeriodMonths - The number of months in the time period
+ * @param {string} filters.supervisionType - Supervision Type of "All"
```
macfarlandian, line 25: *or

src/components/charts/new_revocations/CaseTable/CaseTable.js
```diff
@@ -29,29 +28,23 @@ import {
...
 import { useRootStore } from "../../../../StoreProvider";
 import { METRIC_PERIOD_MONTHS } from "../../../../constants/filterTypes";
+import { dataStorePropTypes } from "../../propTypes";
 
 export const CASES_PER_PAGE = 15;
 
-const CaseTable = ({ dataFilter }) => {
-  const { currentTenantId, filtersStore } = useRootStore();
+const CaseTable = ({ dataStore }) => {
+  const { filtersStore } = useRootStore();
```
macfarlandian, line 40: I don't quite understand why one store comes in as a prop while the other is read from the context? (same question applies to all of the other components that do this)

if nothing else it is making this code hard to follow because it is not obvious which store is being used where

src/components/charts/new_revocations/CaseTable/CaseTable.js
```diff
@@ -29,29 +28,23 @@ import {
...
-  const { isLoading, isError, apiData } = useChartData(
-    `${currentTenantId}/newRevocations`,
-    "revocations_matrix_filtered_caseload"
-  );
-
   const sortedData = useMemo(() => {
-    return dataFilter(apiData || []).sort(comparator);
-  }, [dataFilter, apiData, comparator]);
+    return dataStore.filteredData.sort(comparator);
+  }, [dataStore.filteredData, comparator]);
```
macfarlandian, line 47: I wonder whether it would be cleaner for sorting to be done on the datastore instead of here? mutating the `computed` array in this component feels like it could be bug-prone or at the very least somewhat obscure

src/components/charts/new_revocations/CaseTable/CaseTable.js
daschi: @jovergaag After chatting on slack, we decided to move these into the components.

src/components/charts/new_revocations/CaseTable/CaseTable.js
daschi: For now I updated this to copy the data before calling sort on it. I think this PR is already pretty big so would like to separate out thinking through moving this to the datastore in a different branch.

macfarlandian:
@daschi are the TODOs in the description things you are planning to add to this PR? just wondering whether I should wait for more changes before reviewing again

src/RootStore/DataStore/CaseTable.js
```diff
@@ -0,0 +1,117 @@
...
+
+  isError = false;
+
+  apiData = [];
+
+  filteredData = [];
+
+  metadata = {};
+
+  auth0Context = observable.map({ loading: true });
```
jovergaag, line 47: This is coming from the `rootStore` so I don't think we need it as a a class attribute anymore.

src/RootStore/DataStore/Matrix.js
```diff
@@ -0,0 +1,115 @@
...
+      () => this.fetchData(this.queryFilters)
+    );
+
+    reaction(
+      () => this.queryFilters,
+      (queryString) => this.fetchData(queryString)
+    );
+  }
+
+  get queryFilters() {
```
jovergaag, line 76: This is never different per data store right? Should we put this on the parent `DataStore` class?

src/RootStore/DataStore/RevocationsCharts.js
```diff
@@ -0,0 +1,144 @@
...
+    );
+
+    reaction(
+      () => this.selectedChart,
+      () => this.fetchData(this.queryFilters)
+    );
+
+    reaction(
+      () => this.queryFilters,
+      (queryString) => this.fetchData(queryString, false)
```
jovergaag, line 89: I think we should set `isLoading` to true here. When I was playing around with this, I changed a filter, and during the time that the fetch was happening in the background the chart didn't change - then all of a sudden it changed. This might be confusing to a user because they may not understand that the chart changed in response to the filter change. I think it would be more clear (on all charts) to set `isLoading` to true when the fetch is initiated.

@sychang Thoughts on this? Should we show the loading indicator in the chart if a filter is changed such that it is requesting new data?

src/RootStore/DataStore/CaseTable.js
```diff
@@ -0,0 +1,117 @@
...
+    });
+
+    this.rootStore = rootStore;
+
+    when(
+      () => !get(this.rootStore.auth0Context, "loading"),
+      () => this.fetchData(this.queryFilters)
+    );
+
+    reaction(
```
jovergaag, line 69: I think this can be an `autorun` instead of `reaction`, for all data stores.

src/RootStore/DataStore/RevocationsOverTime.js
```diff
@@ -0,0 +1,143 @@
...
+    });
+
+    return filterOptimizedDataFormat({
+      apiData: data,
+      metadata,
+      filterFn: dataFilter,
+    });
+  }
+
+  get districts() {
```
jovergaag, line 132: I find this a little confusing to keep this in this store. Ultimately it doesn't matter which metric file we're pulling the list of districts from, but I think it would make more sense to store this list elsewhere.  

We have a feature coming down the pipe where we will receive a file ` supervision_location_ids_to_names.json` which will list all the district numbers and map them to their name to be displayed in the filter. When we implement that we will use that as the list of districts for the filter. Until then, parsing them from this metric file is fine, lets just let it live elsewhere.

What about in the `FiltersStore`?

src/RootStore/RootStore.js
```diff
@@ -15,24 +15,46 @@
...
     this.filtersStore = new FiltersStore({ rootStore: this });
+    this.dataStore = new DataStore({ rootStore: this });
+  }
+
+  setAuth0Context() {
+    const auth0Context = useAuth0();
+    this.auth0Context.merge(auth0Context);
+  }
+
+  get getTokenSilently() {
```
jovergaag, line 56: Why can't we just access this directly from `rootStore.auth0Context`?

src/api/metrics/urlHelpers.js
```diff
@@ -0,0 +1,37 @@
...
+// =============================================================================
+import qs from "qs";
+
+/**
+ *
+ * @param {Object} filters - The filter values used to construct the query string to request metric data
+ * @param {string} filters.chargeCategory - A charge category or "All"
+ * @param {Array} filters.district - District IDs or "All"
+ * @param {string} filters.metricPeriodMonths - The number of months in the time period
+ * @param {string} filters.supervisionType - Supervision Type or "All"
```
jovergaag, line 25: We're missing `supervisionLevel` in this list. MO and PA have different filters enabled 🎊 .

src/components/Revocations.js
```diff
@@ -108,92 +98,23 @@ const Revocations = () => {
...
-              dataFilter={matchesAllFilters({
-                filters,
-                skippedFilters: [DISTRICT],
-              })}
-              timeDescription={timeDescription}
-            />
-          </ErrorBoundary>
-        }
-      />
+      <RevocationCharts timeDescription={timeDescription} />
```
jovergaag, line 113: 😍

jovergaag: I think we lost our `ErrorBoundary` around these charts.

src/components/charts/new_revocations/CaseTable/CaseTable.js
```diff
@@ -29,43 +28,31 @@ import {
...
-
   const [page, setPage] = useState(0);
   const { sortOrder, toggleOrder, comparator } = useSort();
 
-  const { isLoading, isError, apiData, unflattenedValues } = useChartData(
-    `${currentTenantId}/newRevocations`,
-    "revocations_matrix_filtered_caseload",
-    false
-  );
+  const filteredData = store.filteredData.slice();
```
jovergaag, line 45: Why is this necessary?

src/components/charts/new_revocations/RevocationCharts/RevocationCharts.js
```diff
@@ -78,19 +113,14 @@ const RevocationCharts = ({
         ))}
       </div>
       <div className="RevocationCharts__chart">
-        {renderSelectedChartSingularLoad()}
+        <ErrorBoundary>{renderSelectedChartSingularLoad()}</ErrorBoundary>
```
jovergaag, line 116: Oh there it is. Nevermind!! :)

src/components/charts/new_revocations/RevocationCharts/RevocationCharts.js
```diff
@@ -32,31 +42,56 @@ const CHARTS = [
...
-  officerChart,
-  violationChart,
-  genderChart,
-  raceChart,
-  districtChart,
-}) => {
-  const [selectedChart, setSelectedChart] = useState(CHARTS[0]);
+const RevocationCharts = ({ timeDescription }) => {
+  const { dataStore } = useRootStore();
+  const store = dataStore.revocationsChartsStore;
```
jovergaag, line 47: Is there a benefit to accessing the store in this parent and passing it down to the chart rather than accessing it directly in the chart?

src/components/charts/new_revocations/RevocationsByOfficer/RevocationsByOfficer.js
```diff
@@ -80,7 +76,7 @@ const RevocationsByOfficer = ({ dataFilter, timeDescription }) => {
           />
         );
       }}
-      generateChartData={createGenerateChartData(dataFilter)}
+      generateChartData={createGenerateChartData(dataStore)}
```
jovergaag, line 79: Eventually we should generate this chart data in the data store and pass the transformed data to the chart. For another day...

jovergaag (changes requested):
Holy cow this is so nice! Those components are looking so fresh and clean!

I found a few little issues and have a few questions. But this is getting really close I think!

src/RootStore/DataStore/DataStore.js
```diff
@@ -0,0 +1,48 @@
...
+import CaseTableStore from "./CaseTable";
+
+export default class DataStore {
+  rootStore;
+
+  revocationsOverTimeStore;
+
+  matrixStore;
+
+  revocationsCharts;
```
jovergaag, line 31: I find the plural `revocationsCharts` to be a little misleading, thinking that it would have the data for all the charts. Maybe just `revocationsChart`?

src/RootStore/DataStore/RevocationsCharts.js
sychang: Yes, let's add the loading indicator!

daschi:
@macfarlandian I did the IE11 performance testing last Friday, and I'll spend this morning adding the tests for the new data stores and responding to Jen's comments. I'll re-request a review when that's ready! Thanks for checking

src/RootStore/DataStore/Matrix.js
daschi: Yeah I would like to DRY up the stores, but right now the `DataStore` is not a parent class, it's acting as a "root store" for all of the data stores. I think I could create another store as an abstract class like `BaseDataStore` that will be the parent to the other stores and have all of the shared attributes.

src/RootStore/DataStore/CaseTable.js
daschi: Ah okay, that makes sense. I would remove the `when` and the `reaction` and replace it with `autorun` that checks the auth0Context condition.

src/RootStore/RootStore.js
daschi: I think we can, I was probably overthinking it :)

src/components/charts/new_revocations/CaseTable/CaseTable.js
daschi: I added this to avoid directly mutating the observable data on line 46 with the sort. @macfarlandian mentioned we should move the sorting to the data store. I started going down that path but it seemed a little more complicated and figured this PR was already getting pretty big that we could tackle that in a separate issue.

src/components/charts/new_revocations/RevocationCharts/RevocationCharts.js
daschi: Yeah, but you make a good point that it's a little hidden. I'm going to move this back into the `Revocations` component so it matches all the other components' pattern.

src/components/charts/new_revocations/RevocationCharts/RevocationCharts.js
daschi: I kept it here since we need the `setSelectedChart` and `selectedChart` states in this parent component, but I'm open to also connecting each of the child component charts directly to the store too since mobx says it's better to access the data closer to the rendering component.

src/RootStore/DataStore/RevocationsOverTime.js
daschi: Oh okay, in that case what do you think about creating a `DistrictsStore`, that will eventually fetch the new json file?

src/RootStore/RootStore.js
jovergaag: Oh I see, ok. The `UserStore` removes the useAuth0 context so we will prob have to make adjustments depending on who merges first!

daschi:
This is ready for another review. I pulled out the shared properties and methods into a `BaseDataStore` superclass and added a test for that class. I think there's still some more tests that could be added for each DataStore's `filterData` function, which I could look into tomorrow (or move onto a new issue?). Let me know what you think..

src/components/charts/new_revocations/CaseTable/CaseTable.js
jovergaag: As we've discussed before, I think we should do all the data prep in the store and only send formatted/sorted/filtered data to the components. One day we will be there :)

src/components/charts/new_revocations/ToggleBar/DistrictFilter.js
```diff
@@ -93,9 +77,9 @@ const DistrictFilter = () => {
         isSearchable
       />
     );
-  }, [district, isLoading, apiData, filtersStore, filters]);
+  };
 
-  return <FilterField label="District">{select}</FilterField>;
+  return <FilterField label="District">{getSelectElement()}</FilterField>;
```
jovergaag, line 82: Oh nice. I was wondering what was wrong with the district filter. Thanks for fixing it!!

src/RootStore/DataStore/BaseDataStore.js
```diff
@@ -0,0 +1,111 @@
...
+      eagerExpand: true,
+    });
+
+    this.file = file;
+
+    this.rootStore = rootStore;
+
+    autorun(() => {
+      if (!get(this.rootStore.auth0Context, "loading")) {
+        this.fetchData(this.queryFilters);
```
jovergaag, line 71: One thing that is a little wonky with this (only in IE11 - Chrome is fine) - when I'm on the profile page and switch from one Lantern tenant to the other...the page becomes unresponsive for 20 seconds or so (depending on if switching to PA or MO) while the data is fetched and loaded. This means the user can't navigate back to the dashboard because the "Lantern" button is disabled for a while.

However, only Recidiviz and CSG users are the ones who are able to switch tenants...so maybe this isn't a big deal? I'm inclined to say not a big deal for now...CSG users are not on IE11 (at least to my knowledge). @sychang Does this seem like something we need to resolve? 

We can kick off the fetch when the user navigates back to the dashboard page to make IE11 happy...but it is kind of nice to have that fetching happening in the background for non-IE11 browsers so that when the user gets to the dashboard the data/charts are ready to go.

jovergaag (changes requested):
THIS IS USEABLE IN IE11!!! I actually used the dashboard in IE11- filtered and switched charts for at least 10 mins and it never crashed. Well done!!!

Just one issue with `eagerExpand`, but its oh so close!

src/RootStore/DataStore/BaseDataStore.js
sychang: This is absolutely not a big deal, since no one on Recidiviz nor CSG are using IE11

jessex: Just flying by with a quick comment: sounds like something to defer to some later point, so let's just make sure we file a task so we don't forget!

src/RootStore/DataStore/RevocationsOverTimeStore.js
```diff
@@ -0,0 +1,98 @@
...
+    super({ rootStore, file: `revocations_matrix_by_month` });
+    makeObservable(this, {
+      fetchData: flow,
+      expandedData: observable.shallow,
+      // TODO: Remove once we get a separate districts file
+      districts: computed,
+    });
+  }
+
+  *fetchData(queryString = "") {
```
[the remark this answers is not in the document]
daschi: This needs to also store the expanded data to create the `districts` data. I'm going to open a separate issue to create a `DistrictsStore` as a fast follow to this.

src/RootStore/__tests__/BaseDataStore.test.js
```diff
@@ -0,0 +1,155 @@
...
+        expect(baseStore.isError).toEqual(false);
+      });
+
+      it("sets apiData", () => {
+        expect(baseStore.apiData).toEqual([["0"], ["0"]]);
+      });
+    });
+  });
+
+  describe("when API responds with an error", () => {
```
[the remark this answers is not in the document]
daschi: I added a note to that issue that data fetching is coming from the data stores now for new revocations charts.

src/api/metrics/__tests__/urlHelpers.test.js
```diff
@@ -0,0 +1,43 @@
+// Recidiviz - a data platform for criminal justice reform
```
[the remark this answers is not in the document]
daschi: Do you mean `getQueryStringFromFilters`? Yeah I can move it into the DataStore helpers.

src/components/charts/new_revocations/CaseTable/CaseTable.js
daschi: I can open a issue for this

src/RootStore/DataStore/BaseDataStore.js
daschi: I can open an issue for this

src/api/metrics/__tests__/urlHelpers.test.js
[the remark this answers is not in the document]
jovergaag: Yes thats what I meant. Thanks for being psychic :)

jovergaag (approved):
You're a hero 🌈

macfarlandian (approved):
🎉

Issue #670, opened 2020-12-17
[MO District Restricted View] Update getUserDistricts to actually read proper districts list from file/Auth0


ageiduschek:
**What needs to be done? Why does it need to be done?**
Restricted district support was initially implemented in [this PR](https://github.com/Recidiviz/pulse-dashboard/pull/461/files), but will need to be adapted to support the final backend implementation, which will likely either a) send a `districts` list via the Auth0 user or b) send a mapping of email to districts list through an API file.

The getUserDistricts() function will have to be adapted to read the district information properly.

Note: there is extra code in the original PR built with the understanding that we would give regional supervisors access to data only in their region. But @sychang confirmed that this is not the case. There will be no region information passed to the frontend and we do not need to maintain the mapping of MO region -> district in `src/tenants.json`.

**Additional context**
Design task to figure out communication with pulse backend here: #669

PR #698, opened 2021-01-20
UserStore and AuthWall instead of TenantRoutes

jovergaag:
## Description of the change

For the district restricted access feature, we need to fetch the user to district mappings from the API. Rather than initiate this fetch from the charts, I decided to make the move to a `UserStore` and use that store to initiate the fetch (this PR does not have the district restricted access work yet). The `UserStore` and `AuthWall` is largely ported over from public-dashboard, with changes made as needed.

A few things to note:
- this introduces TypeScript, it didn't make sense to me to port the public-dashboard code to Javascript when we are eventually going to try to transition the whole repo to TypeScript (along with this addition of TS, I added tsconfig.json and added TS linting which required a few linting changes in files unrelated to the UserStore)
- This moves the functionality in `TenantRoutes` of checking whether the user has access to a given layout into the `AuthWall`. Eventually, we should keep the `currentTenantId` in the path and use that to determine which layout to render, but until then this will work.
- changes the `yarn test` script to run `react-scripts test` which I find much more efficient to work with
- - removes previous implementation of restricted district access because the new implementation makes more sense (coming soon...)

I have tested the auth flows thoroughly after this change, including:
- authentication from users from all state_codes available
- creating new users - redirect looks good
- auth/creating new user on IE11

## Type of change

## Related issues

> Related to #670 

### Development

### Code review


src/RootStore/RootStore.js
```diff
@@ -30,6 +52,11 @@ export default class RootStore {
   constructor() {
     makeObservable(this, { filters: computed, currentTenantId: computed });
 
+    this.userStore = new UserStore({
+      authSettings: getAuthSettings(),
```
daschi, line 56: I'm curious why you chose to put this in the RootStore instead of as part of the UserStore?

daschi (approved):
This looks great, I like the restructure of the AuthWall component into the App and how that simplifies the tests! I'm happy to get this merged first and then incorporate the UserStore in my DataStore PR.

.eslintrc.json
```diff
@@ -26,28 +53,27 @@
...
-    "import/prefer-default-export": "off"
-  },
-  "env": {
-    "browser": true,
-    "jest": true
+    "import/prefer-default-export": "off",
+    // the bare eslint rule breaks in typescript
+    // https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-use-before-define.md#how-to-use
+    "no-use-before-define": "off",
+    "@typescript-eslint/no-use-before-define": ["error"],
```
macfarlandian, line 60: I have found that I prefer putting all the `typescript-eslint` stuff to the typescript file overrides section ... sometimes the TS and JS eslint rules are not really compatible and they fight with each other.

that may include an "extends" config as well that includes the TS configs for eslint and prettier

src/AuthWall/AuthWall.tsx
```diff
@@ -0,0 +1,74 @@
...
+const AuthWall: React.FC = ({ children }) => {
+  const { userStore, currentTenantId } = useRootStore();
+
+  useEffect(
+    () =>
+      // return when's disposer so it is cleaned up if it never runs
+      when(
+        () => !userStore.isAuthorized,
+        () => userStore.authorize()
+      ),
```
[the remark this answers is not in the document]
macfarlandian: this should work as-is because it returns the disposer function directly, so it'll get called if necessary

src/AuthWall/AuthWall.tsx
```diff
@@ -0,0 +1,74 @@
...
+    throw userStore.authError;
+  }
+
+  if (userStore.isLoading) {
+    return <Loading />;
+  }
+
+  if (userStore.awaitingVerification) {
+    return <VerificationNeeded />;
+  }
```
macfarlandian, line 53: should this redirect to `/verify` instead since that path already exists?

src/AuthWall/AuthWall.tsx
```diff
@@ -0,0 +1,74 @@
...
+
+  if (userStore.isAuthorized) {
+    const authorizedChildren = React.Children.map(children, (child: any) => {
```
macfarlandian, line 56: instead of casting to `any` you could use `React.isValidElement` as a type guard

package.json
```diff
@@ -123,8 +132,7 @@
   },
   "husky": {
     "hooks": {
-      "pre-commit": "lint-staged",
-      "pre-push": "yarn test"
+      "pre-commit": "lint-staged"
```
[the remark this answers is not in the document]
jovergaag: I changed the test script to be `react-scripts test` which won't work with the test pre-push hook. I'm open to changing this back if you liked it!

src/RootStore/UserStore.ts
```diff
@@ -0,0 +1,156 @@
...
+      runInAction(() => {
+        this.isLoading = false;
+
+        if (user.email_verified) {
+          this.isAuthorized = true;
+          this.awaitingVerification = false;
+          this.getTokenSilently = (...p: any) => auth0.getTokenSilently(...p);
+          this.logout = (...p: any) => auth0.logout(...p);
+          this.user = user;
+          this.stateCode = getUserStateCode(user);
```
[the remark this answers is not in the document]
jovergaag: I had considered this. What I ended up landing on is that the user comes back from Auth0 with `state_code` in the metadata. Then when the TenantStore reads that stateCode it translates it into the `tenantId` which is the currently active tenant. For most users they will be the same, but for some users, the `state_code` and `tenantId` will be different and they do have different significance. For example, when I log in, my user metadata is `{ state_code: 'RECIDIVIZ' }` and then the tenantId is set to whatever tenant is currently active in the session out of the availableTenants to me (US_ND, US_MO, or US_PA).

If this is too confusing I can think through ways to make it more clear. But in my mind stateCode and tenantId are different.

src/__tests__/App.test.js
```diff
@@ -96,92 +92,82 @@ describe("App tests", () => {
...
     });
 
-    render(
-      <StoreProvider defaultTenantId="US_PA">
-        <App />
-      </StoreProvider>
-    );
+    // do not log the expected error - keep tests less verbose
+    jest.spyOn(console, "error").mockImplementation(() => {});
+    jest.spyOn(console, "log").mockImplementation(() => {});
```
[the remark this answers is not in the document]
jovergaag: Oh ya good catch. I had confused `resetAllMocks` with `restoreAllMocks`

src/views/Profile.js
```diff
@@ -17,21 +17,17 @@
...
 
 const Profile = () => {
-  const { loading, user } = useAuth0();
+  const { userStore } = useRootStore();
+  const { isLoading, user } = userStore;
 
-  const availableStateCodes = getAvailableStateCodes(user);
-
-  if (loading || !user) {
+  if (isLoading || !user) {
```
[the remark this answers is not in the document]
jovergaag: I wouldn't be!

src/RootStore/RootStore.js
jovergaag: I was just following the pattern in public-dashboard. As @macfarlandian mentioned earlier today, there are starting to be a few repos using this same `UserStore` and `AuthWall`. It just felt better to keep them consistent in case we ever try to consolidate.

Open to changing that though!

.eslintrc.json
jovergaag: Good catch. TS linting/prettier was working for me - but once I moved it up and turned on the plugins I got a few more errors so it clearly wasn't configured correctly. I think it's better now but let me know if you see anything else thats off.

src/AuthWall/AuthWall.tsx
jovergaag: I played around with this a bit. It turns out that our Auth0 rule is forcing the redirect. I ended up removing this `awaitingVerification` entirely and have tested the new user flow thoroughly and its working great!

macfarlandian (approved):

PR #705, opened 2021-01-21
Add the subset_manifest for violationType and generate cache keys based on filter params

### LINKED ISSUE #654: [API] Establish subset manifest write to cache with new cacheKey
Establish the subset manifest and list of files to apply the subsetting logic. For now, the subsetting (filtering) logic can just be a passthrough, but this can get that process established. 

Establish the subset manifest as the following:
 `subset_manifest: {
   violation_type: [["absconded", "all", "elec_monitoring", "escaped",  "high_tech",  "low_tech", "med_tech",  "municipal", "no_violations",
"substance_abuse", "technical"], ["felony", "law", "misdemeanor"]]
}`

The following files will have the subsetting filtering applied to them:
revocations_matrix_distribution_by_district
revocations_matrix_distribution_by_gender
revocations_matrix_distribution_by_officer
revocations_matrix_distribution_by_race
revocations_matrix_distribution_by_risk_level
revocations_matrix_distribution_by_violation

Write the subset file data (at this point, same as the original file) to the Redis cache using the new cacheKey. Establish the new cacheKey that incorporates details about the subset manifest. We want to apply the filtering logic to every file for every permutation of the subset manifest. So for example, if there are two keys in the subset manifest, and each of those keys has two subsets defined, we will need 2 x 2 = 4 subset files as a result. The cacheKey will need to capture which permutation is included in the file.

So the cacheKey could be:
${stateCode}-${file}-${subsetOneString}-${subsetTwoString}

Given this example subset manifest:
 `subset_manifest: {
   violation_type: [["absconded", "all", "elec_monitoring", "escaped",  "high_tech",  "low_tech", "med_tech",  "municipal", "no_violations", "substance_abuse", "technical"], ["felony", "law", "misdemeanor"]],
  charge_category: [ "all"], [ "domestic_violence", "alcohol_drug", "general", "serious_mental_illness", "sex_offense"]
}`

subsetOneString could be: `"violation_type=['felony', 'law', 'misdemeanor']"`
subsetTwoString could be: `"charge_category=['domestic_violence', 'alcohol_drug', 'general', 'serious_mental_illness', 'sex_offense']"`

OR maybe we just reference the index of the subset which might be cleaner but less descriptive:

subsetOneString could be: `"violation_type=0"`
subsetTwoString could be: `"violation_type=1"`

The result of this work will be storing the original file with the word 'original' in the cacheKey, as well as the full original file in each permutation of subsets cacheKeys. Thats ok because subset file caches should not be used yet. 

Note: the subset_manifest might make more sense as an array ob objects instead of array of arrays. That way we can make a note of which is the default subset

**Additional context**

daschi:
## Description of the change

This is the first step towards implementing the subsets of revocations files for the frontend. This PR creates all of the cache keys for the every combination of subset files based on the subset manifest. The `SUBSET_MANIFEST` is a nested array defined under the `server/constants` directory. 

This PR does NOT split up the files yet by subsets. Each cache key with the subset description will still point to the full file contents in the cache. I've also maintained the "original" cache keys so that the FE can continue to receive the response for the full file contents until it's ready to handle the split up files (and added TODOs where we need to remove this code later).

The cache keys with the subset keys look like this:

`US_PA-newRevocation-revocations_matrix_distribution_by_district-violationType=0`

We only have `violationType` subsets right now, but it will be easy to add in additional subsets. You can see how this would work in the `cacheKeys.test.js` and the `refreshRedisCache.test.js` files. 

## Type of change

## Related issues

> Closes #654 

### Development

### Code review


server/core/cacheManager.js
```diff
@@ -89,9 +89,20 @@ function clearMemoryCache() {
 }
 
 function cacheResponse(cacheKey, fetchValue, callback) {
-  const cache = getCache(cacheKey);
+  const cache = Array.isArray(cacheKey)
```
jovergaag, line 92: It might be nice to put a comment here explaining when it will be an array and when it won't. Or tests. It will help when we switch away from `cacheKeyOriginal`

server/core/cacheManager.js
```diff
@@ -89,9 +89,20 @@ function clearMemoryCache() {
...
+    ? getCache(cacheKey[0])
+    : getCache(cacheKey);
+  let cacheKeyOriginal = cacheKey;
 
-  return cache.wrap(cacheKey, fetchValue).then(
+  if (Array.isArray(cacheKey)) {
+    // Set the subset cache keys, but return the original cache value until FE is ready.
+    cache.wrap(cacheKey[1], fetchValue);
+    [cacheKeyOriginal] = cacheKey;
```
jovergaag, line 100: I'm not understanding this line - why do we want to set the `cacheKeyOriginal` to the array cacheKey?

server/core/refreshRedisCache.js
```diff
@@ -26,16 +26,43 @@
...
-    const cacheKey = `${cacheKeyPrefix}-${fileKey}`;
-    const metricFile = { [fileKey]: files[fileKey] };
-    console.log(`Setting cache for: ${cacheKey}...`);
-    cachePromises.push(cache.set(cacheKey, metricFile));
+  Object.keys(files).forEach((file) => {
+    const cacheKey = `${cacheKeyPrefix}-${file}`;
+    const metricFile = { [file]: files[file] };
+
+    if (FILES_WITH_SUBSETS.includes(file)) {
+      // TODO: Remove line 57 once the front end is ready to receive the subset files.
```
jovergaag, line 56: This should be line 59

server/constants/subsetManifest.js
```diff
@@ -0,0 +1,35 @@
+const SUBSET_MANIFEST = [
```
jovergaag, line 1: As I'm looking at this and thinking through edge cases, I think we need to guard against dimensions being added to the subset manifest that are multi-select filters. The only dimension right now that is multi-select is the district filter. We can't allow that to be subsetted because a user can easily select districts that are in more than one subset.

Can you add a guard and comment here to that effect pretty pls.

server/utils/cacheKeys.js
```diff
@@ -0,0 +1,113 @@
+const {
```
jovergaag, line 1: We're missing the copyright comment here.

server/utils/__tests__/cacheKeys.test.js
```diff
@@ -0,0 +1,90 @@
+import { getCacheKey, getSubsetCacheKeyCombinations } from "../cacheKeys";
```
jovergaag, line 1: Also missing copyright comment here :)

server/utils/cacheKeys.js
```diff
@@ -0,0 +1,113 @@
...
+  SUBSET_MANIFEST,
+  FILES_WITH_SUBSETS,
+} = require("../constants/subsetManifest");
+
+/**
+ *
+ * @param {string[][]} collection - A nested array of strings, [["dimensionKeyA=0"], ["dimensionKeyB=0"]]
+ * @param {number} max - The max number of items to include in each combination.
+ *
+ * @returns {string[][]} - A nested array of all the combinations. [["dimensionKeyA=0", "dimensionKeyB=0"]]
```
jovergaag, line 11: Can you show the [1] index of this return value in the comment. I think it would be helpful to see that it is `["dimensionKeyA=0", "dimensionKeyB=1"]`

jovergaag (changes requested):
Cool! It is a little confusing since we're handling both the subset keys and the original keys, but I think you did a good job of documenting those, and they won't last long before we swap over to the subset keys entirely so i say its ok!

Few comments...

server/core/cacheManager.js
daschi: This is actually de-structuring the array to get the value for `cacheKeyOriginal`. When `cacheKey` is an array it'll look like: `[cacheKeyOriginal, cacheKey]`. The de-structuring here is taking the first element and assigning it to the `cacheKeyOriginal` variable defined on line 95. I didn't originally write it as a de-structuring assignment, I blame eslint for that :P

server/constants/subsetManifest.js
daschi: Sure, I added a function to get the subset manifest that will validate the values, but for now it's all defined in the same place. Let me know what you think and if it's overkill?

server/core/cacheManager.js
jovergaag: OH. Of course it is :)

jovergaag (approved):
🌮

PR #707, opened 2021-01-23
Restricted district access

### LINKED ISSUE #632: Land FE support for district-restricted view in MO
PRD: https://docs.google.com/document/d/1bxSg8JgQHFFZZDSjBUKM-fOtU-8r3FVieQD5J81crkA/edit#

- [ ] Land any previous frontend PRs that support this feature
- [ ] Test integration when backend support lands

**Additional context**
Target EOY for feature support so district reports can be sent to supervisors in early Jan.

### LINKED ISSUE #670: [MO District Restricted View] Update getUserDistricts to actually read proper districts list from file/Auth0

jovergaag:
## Description of the change

We want to set the dashboard view to a specific district for certain users. This is not for all users of any given state, only for a subset of users. If a user's district is restricted, it should lock the district filter to that district, and filter all data such that the charts show only data for that district. The exception to that is for the district chart, we want to show all districts in the district chart, but highlight the restricted district.

If there is an error with the fetch when fetching the list of restricted users, or if the district ID for that particular user is invalid, display an error and do not show any data to that user. That error message looks like this:
Screen Shot 2021-01-22 at 3 58 55 PM

This PR:
- adds the restricted access email file to the API
- fetches the restricted access email file from the UserStore
- if the current user's email address is in the list of restricted access emails, set the restrictedDistrict after verifying that the restrictedDistrict is valid
- moves the list of districts for the district filter into the TenantStore since that list is dependent on the Tenant
- makes a few changes to the eslintconfig. I'm really struggling getting linting and prettier working for both .js and .ts files simultaneously. I think we will just need to duplicate rules in both the JS and TS "overrides" section. I'm definitely open to other suggestions here...
- Adds the new error message format

## Type of change

## Related issues

> Closes #670 and closes #632

### Development

### Code review


src/RootStore/DataStore/BaseDataStore.js
```diff
@@ -67,8 +67,11 @@ export default class BaseDataStore {
     autorun(() => {
       const { userStore } = this.rootStore;
 
-      if (userStore && !userStore.isLoading) {
-        this.fetchData(this.queryFilters);
+      if (userStore && !userStore.userIsLoading) {
```
daschi, line 70: I think we may need another condition here to wait on the `restrictedDistrictIsLoading`, this looks like it is currently fetching data when the user loads for all districts, and then after `restrictedDistrictIsLoading` is updated it loads the restricted district data.

daschi (changes requested):
This is exciting!! I have a some comments/questions for now, mostly about the mobx setup.

src/RootStore/DataStore/BaseDataStore.js
```diff
@@ -67,8 +67,11 @@ export default class BaseDataStore {
...
+        this.fetchData({
+          tenantId: this.rootStore.currentTenantId,
```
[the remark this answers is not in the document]
jovergaag: We need tenantId to be in the reaction so that the fetch will be triggered when the tenant changes. Without this, switching between MO and PA didn't fetch the new data.

src/RootStore/UserStore.ts
```diff
@@ -54,31 +65,39 @@ export default class UserStore {
...
 
     this.isAuthorized = false;
-    this.isLoading = true;
+    this.userIsLoading = true;
+    this.restrictedDistrictIsLoading = true;
 
-    // TODO 670 set restricted district based on user district
+    autorun(() => {
+      if (!this.userIsLoading) {
+        this.fetchRestrictedDistrictData(this.rootStore?.currentTenantId);
```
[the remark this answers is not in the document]
jovergaag: Same reason as previous comment, so that we fetch new data when the tenant changes.

src/RootStore/UserStore.ts
```diff
@@ -146,4 +172,58 @@ export default class UserStore {
...
+
+  /**
+   * Returns the state code of the authorized state for the given user.
+   * For Recidiviz users or users in demo mode, this will be 'recidiviz'.
+   */
+  get stateCode(): string {
+    return getUserStateCode(this.user);
+  }
+
+  async fetchRestrictedDistrictData(
```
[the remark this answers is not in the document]
jovergaag: I was struggling to get the flow generator to work with TypeScript, but I just got it to work so pushing that up!

src/utils/authentication/viewAuthentication.js
```diff
@@ -24,12 +24,12 @@ export function getDemoUser() {
     picture:
       "https://ui-avatars.com/api/?name=Demo+Jones&background=0D8ABC&color=fff&rounded=true",
     name: "Demo Jones",
-    email: "notarealemail@recidiviz.org",
+    // email: "notarealemail@recidiviz.org",
```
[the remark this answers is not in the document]
jovergaag: No I left those in there so that it is easy to switch demo mode to restricted access.

src/RootStore/__tests__/UserStore.test.ts
```diff
@@ -169,14 +185,139 @@ test.each(Object.keys(TENANTS))(
...
+
+    afterEach(() => {
+      jest.resetAllMocks();
+    });
+
+    it("makes a request to the correct endpoint for the data", async () => {
+      expect(callMetricsApi).toHaveBeenCalledTimes(1);
+      expect(callMetricsApi).toHaveBeenCalledWith(
+        endpoint,
+        expect.any(Function)
```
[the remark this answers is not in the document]
jovergaag: I tried for a long time but couldn't get that to work. The expectation was expecting a function typed: `mockConstructor` but it was receiving a function typed `anonymous`.  I couldn't figure out how to type case the mockConstructor. Any ideas @macfarlandian?

src/RootStore/__tests__/UserStore.test.ts
```diff
@@ -169,14 +185,139 @@ test.each(Object.keys(TENANTS))(
...
+      jest.resetAllMocks();
+    });
+
+    it("does not set the restrictedDistrict", () => {
+      expect(userStore.restrictedDistrict).toBeFalsy();
+    });
+
+    it("sets an authError and isLoading to false", () => {
+      expect(userStore.authError).toEqual(
+        new Error(ERROR_MESSAGES.unauthorized)
```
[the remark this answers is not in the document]
jovergaag: If there is an API error - we want to fail closed - meaning not show the dashboard at all and tell the user that they're not authorized to see the page. Serena wanted to show the same message as when there is an authorization error in this case.

src/RootStore/DataStore/BaseDataStore.js
jovergaag: Hmm, I don't think that it relies on `restrictedDistrictIsLoading`. We only need the user (specifically `getTokenSilently` to initiate this fetch. The restricted district fetch can happen simultaneously.

The only thing that relies on the `restrictedDistrictIsLoading` is the AuthWall, which gates whether or not the dashboard can be displayed.

src/RootStore/UserStore.ts
```diff
@@ -146,4 +172,58 @@ export default class UserStore {
...
+      this.restrictedDistrictIsLoading = false;
+    } catch (error) {
+      this.authError = new Error(ERROR_MESSAGES.unauthorized);
+      this.restrictedDistrictIsLoading = false;
+    }
+  }
+
+  setRestrictedDistrict(restrictedEmails: Array<RestrictedAccessEmails>): void {
+    const restrictedEmail =
+      restrictedEmails &&
```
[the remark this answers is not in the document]
jovergaag: Hmm, good point. We are encrypting the response but once it lands those addresses are exposed locally. @macfarlandian Thoughts on whether this is a security concern to send several email addresses in the network response?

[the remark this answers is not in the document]
jovergaag: I guess, it this does seem problematic, the solution is to add a queryString to the URI which has the email address, or make it a POST request and add the email address to the body and add the filtering to the BE such that the API responds with the restricted district ID string.

src/RootStore/DataStore/BaseDataStore.js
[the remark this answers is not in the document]
macfarlandian: FWIW, if this is only about making sure this reacts to changes in `currentTenantId`, it might be better to [use `reaction`](https://mobx.js.org/reactions.html#reaction) instead of `autorun` and reference `currentTenantId` in the data function. You could achieve the same result with less indirection

src/RootStore/TenantStore.js
```diff
@@ -44,15 +50,20 @@ export default class TenantStore {
...
-  user;
+  districts = [];
+
+  districtsIsLoading = true;
 
   constructor({ rootStore }) {
-    makeAutoObservable(this);
+    makeAutoObservable(this, {
+      setDistricts: action,
+      setCurrentTenantId: action,
```
macfarlandian, line 60: this shouldn't be necessary with `makeAutoObservable`?

src/RootStore/UserStore.ts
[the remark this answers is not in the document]
macfarlandian: it does seem ... questionable? In practical terms I am not sure if it's exposing information to anyone who doesn't already have it (i.e. because it's their colleague's email address) but I think if we can avoid it we probably should. This may just be a hard compliance issue also but that is a question for @jessex

src/RootStore/__tests__/UserStore.test.ts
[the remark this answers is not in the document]
macfarlandian: I think you can do something like `jest.Mock<RootStore>` when you typecast the mock object and it will pick up the type definition of the thing it's mocking?

src/RootStore/DataStore/BaseDataStore.js
daschi: Yeah, but if the restricted district is not loaded yet by the time the DataStore fetches data, then the first fetch on page load will use the default filters which includes all districts. That'll fetch all of the districts on page load, and then when the filter is updated with the restricted district, it will do a 2nd fetch for just the restricted data. I think we would only want the client to fetch the data with the restricted district filter on.

jovergaag: Weird. In demo mode it does fetch twice, but in dev mode it only fetches once. Investigating...

jovergaag: Fixed.

src/RootStore/UserStore.ts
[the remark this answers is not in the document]
jessex: Thanks for looping me in here, everyone.

After giving this some thought, I think this is close enough to the borderline that we should go with one of Jen's suggested alternatives instead of the current approach. It's not just that you could be seeing your colleagues' email addresses and that in certain hard-to-enumerate scenarios that might be problematic, it's also that if someone were able to somehow get access to this view when they shouldn't have it, this would further escalate their level of access by handing them a set of email addresses for other users. We should avoid building paths of rapid escalation like this, so let's take a different approach.

.travis.yml
```diff
@@ -6,7 +6,8 @@ before_install:
 install:
   - yarn install
 script:
-  - yarn test --coverage && cat ./coverage/lcov.info | ./node_modules/coveralls/bin/coveralls.js && rm -rf ./coverage
+  - yarn test-server
```
jovergaag, line 9: I realized that Travis was only running our client tests, so I added this to run the server tests separately. Since we are running with `--coverage` on the client tests, I did not include it here.

src/RootStore/UserStore.ts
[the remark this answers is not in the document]
jovergaag: I created a new endpoint called `restrictedAccess`. The `UserStore` makes a POST request and includes the user's email in the body. The response only includes the restricted access object that matches the user's email address, if any.

src/RootStore/UserStore.ts
```diff
@@ -146,4 +172,52 @@ export default class UserStore {
...
+  get stateCode(): string {
+    return getUserStateCode(this.user);
+  }
+
+  fetchRestrictedDistrictData = flow(function* (
+    this: UserStore,
+    tenantId: string | undefined
+  ) {
+    const file = "supervision_location_restricted_access_emails";
+    const endpoint = `${tenantId}/restrictedAccess`;
```
macfarlandian, line 189: if `tenantId` can be undefined here, won't this path be invalid?

src/RootStore/__tests__/UserStore.test.ts
```diff
@@ -169,14 +186,131 @@ test.each(Object.keys(TENANTS))(
...
+      expect(callRestrictedAccessApi).toHaveBeenCalledWith(
+        endpoint,
+        userEmail,
+        mockGetTokenSilently
+      );
+    });
+
+    it("sets isLoading to false and isError to false", () => {
+      expect(userStore.restrictedDistrictIsLoading).toEqual(false);
+    });
```
macfarlandian, line 243: no error here?

src/RootStore/__tests__/UserStore.test.ts
```diff
@@ -169,14 +186,131 @@ test.each(Object.keys(TENANTS))(
...
+        email: userEmail,
+      });
+
+      userStore = new UserStore({
+        authSettings: testAuthSettings,
+        rootStore: new RootStore(),
+      });
+
+      await userStore.authorize();
+      endpoint = `${tenantId}/restrictedAccess`;
```
macfarlandian, line 271: this describe block and the one  below it do not seem to actually be using this endpoint for anything. should we be verifying that it is not called, or something like that?

src/RootStore/UserStore.ts
jovergaag: Yes you're right, it should not be undefined. I wrote this before adding the `this.rootStore?.currentTenantId` check in the autorun. Good catch, thank you!

macfarlandian (approved):

server/routes/api.js
```diff
@@ -47,10 +47,41 @@ function responder(res) {
 
 // TODO: Generalize this API to take in the metric type and file as request parameters in all calls
 
+function restrictedAccess(req, res) {
+  const { stateCode } = req.params;
+  const { userEmail } = req.body;
+  const metricType = "newRevocation";
+  const file = "supervision_location_restricted_access_emails";
+  if (!userEmail) {
+    res.status(400).json({ error: "request is missing userEmail parameter" });
```
daschi, line 56: NBD for now, b this could also be defined in the `app.js` endpoint as a validation (similar to how we're doing the param validations). Also `400` is  defined in this file as a constant `BAD_REQUEST`.

server/routes/api.js
```diff
@@ -47,10 +47,41 @@ function responder(res) {
...
+  } else {
+    const cacheKey = `${stateCode.toUpperCase()}-restrictedAccess`;
+    cacheResponse(
+      cacheKey,
+      () => fetchMetrics(stateCode, metricType, file, isDemoMode),
+      responder(res),
+      (result) => {
```
daschi, line 63: I wonder if instead of making the `cacheResponse` more complicated, whether it would be better to define another fetch function that wraps `fetchMetrics` for this particular case. Something like `fetchRestrictedDistricts` that would call `fetchMetrics` and then modify the results before passing it to the cache.

daschi (commented):
I wouldn't want to hold this up much longer, but had a couple other comments that I'll leave up to you to decide! Happy to go with this as it is too.

server/routes/api.js
jovergaag: Great suggestion. I like it much better this new way!

daschi (approved):
⭐ ⭐ ⭐

PR #728, opened 2021-01-28
[API] Move filters to a shared directory and create filtered subset files & metadata dimensions

daschi:
## Description of the change

This adds a shared directory `shared-filters` (name TBD -- I think we should probably name this something else, but wanted to get something up for review!) which is added via npm as a local path dependency. This is so the frontend can require this shared code since CRA does not allow imports from outside the `src/` directory. 

The shared functionality that's been moved out of the frontend includes:

- `filterOptimizedDataFormat` - Takes an unflattened matrix (2d array) and returns an array of filtered objects
- `dataFilters` - The boolean checks for filters, exports `matchesAllFilters` and `matchesTopLevelFilters`
- `optimizedFormatHelpers` - Exports the helpers for working with the unflattened and flattened matrix

There's still a lot of TODOs I would like to get to:

- [x] Add JSDOCs for all the shared functions and the new server helpers
- [x] Add tests for the shared-filters functions
- [x] Add tests for the server side filtering helpers (`transformFilters`, `applyFilters`)
- [x] Add the same functionality to the `refreshCache` endpoint. This PR only adds applying the filters for the `newRevocationFile` endpoint. 

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [x] New feature (non-breaking change that adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [ ] Configuration change (adjusts configuration to achieve some end related to functionality, development, performance, or security)

## Related issues

> Closes #655 
> Closes https://github.com/Recidiviz/pulse-dashboard/issues/656

> Dependent on https://github.com/Recidiviz/recidiviz-data/pull/5529
> Dependent on  #641 
### Development

These boxes should be checked by the submitter prior to merging:

- [x] Manual testing against realistic data has been performed locally

### Code review

These boxes should be checked by reviewers prior to merging:

- [x] This pull request has a descriptive title and information useful to a reviewer
- [x] This pull request has been moved out of a Draft state, has no "Work In Progress" label, and has assigned reviewers
- [x] Potential security implications or infrastructural changes have been considered, if relevant

PR #736, opened 2021-02-01
[FE] Only request metric file if filter value is missing from the dimension manifest

### LINKED ISSUE #658: [API-FE] Add each metric file dimension manifest into a store and initiate fetch if filter value is missing
Add a store that includes the dimension manifest for each metric file when it is received from the API.

On a filter change, compare the new filter value to the dimension manifest of each file, and initiate a fetch request for that file if it is not an included dimension value.

**Additional context**

daschi:
## Description of the change

Currently on staging we are making a request to the API on every filter change. This PR changes that flow so that we will only request a new metric file if the value from the filter is not in the dimension manifest found in the file's metadata.

When we start using the subset files, the dimension manifest sent from the API will be updated to include only the values found in that subset. That way when a filter changes, it will only request a new subset file when it needs to.  

Some notable changes:

- I changed the observable value for `apiData` to be an `observable.ref` and removed the separate metadata property. The new apiData value looks like this: `this.apiData = { data, metadata }`. This is because I wanted to make `metadata` observable, and I did not want the `filteredData` to be computed twice (once for the metadata value change and again for the apiData value change). This way, when `this.apiData` changes, the data is only filtered once.  
- I changed `filteredData` to be a computed value. This is so it can be computed when either apiData or filters change. This does however remove the `isLoading` state when filtering...I was trying to think of a way to add that back and I'm not sure the best place to put that yet... 

## Type of change

## Related issues

> Closes #658

NOTE: This PR should merge before #728 

### Development

### Code review

### DIFF
--- a/src/RootStore/DataStore/BaseDataStore.js
+++ b/src/RootStore/DataStore/BaseDataStore.js
@@ -22,10 +22,16 @@ import {
   computed,
   toJS,
   autorun,
+  reaction,
 } from "mobx";
 
 import { callMetricsApi } from "../../api/metrics/metricsClient";
-import { processResponseData, getQueryStringFromFilters } from "./helpers";
+import {
+  processResponseData,
+  getQueryStringFromFilters,
+  dimensionManifestIncludesFilterValues,
+} from "./helpers";
+import { FILTER_TYPE_MAP, DISTRICT } from "../../constants/filterTypes";
 
 /**
  * BaseDataStore is an abstract class that should never be directly instantiated.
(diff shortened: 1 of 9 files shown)


src/RootStore/DataStore/BaseDataStore.js
```diff
@@ -38,74 +44,117 @@ export default class BaseDataStore {
...
+      const [name, values] = dimension;
+      if (FILTER_TYPE_MAP[name]) {
+        acc[FILTER_TYPE_MAP[name]] = values;
+      }
+      return acc;
+    }, {});
+  }
+
+  get filteredData() {
+    console.error(`filteredData should be defined in the subclass.`);
```
macfarlandian, line 141: if this should really never be used on its own why not throw this error rather than just logging it?

src/RootStore/DataStore/CaseTableStore.js
```diff
@@ -20,17 +20,24 @@ import { filterOptimizedDataFormat } from "../../utils/charts/dataFilters";
...
-    super({ rootStore, file: `revocations_matrix_filtered_caseload` });
+    super({
+      rootStore,
+      file: `revocations_matrix_filtered_caseload`,
+      treatCategoryAllAsAbsent: true,
+    });
   }
 
-  filterData({ data, metadata }) {
+  get filteredData() {
```
macfarlandian, line 30: I am [pretty sure](https://mobx.js.org/actions.html#actions-are-untracked) that if you wanted to set the `isLoading` flag while this function is running, you could just do it as long as you wrap those changes in `action`s?

src/RootStore/DataStore/BaseDataStore.js
daschi: Yeah that makes sense, will update that and add a test for it.

src/RootStore/DataStore/CaseTableStore.js
daschi: I played around with this a bit (using the actions in the `filteredData` getter) and I think we actually don't need it anymore. The loading delay we were seeing before was because we were fetching on every filter change, and the loading state is still being displayed before & after the fetch completes. When I added the `isLoading` changes before and after the filtering, it *seemed* to complete before the loading spinner could even render. I'm going to do a quick check in IE11 too to make sure that's the case there as well.

macfarlandian: that makes sense ... i was surprised that the filter function alone would be that expensive but what do i know

src/RootStore/DataStore/BaseDataStore.js
```diff
@@ -38,74 +44,116 @@ export default class BaseDataStore {
...
       );
-      const processedData = processResponseData(
+      this.apiData = processResponseData(
         responseData,
         this.file,
         this.eagerExpand
       );
-      this.apiData = processedData.data;
-      this.metadata = processedData.metadata;
-      this.filteredData = this.filterData(processedData);
```
macfarlandian, line 108: just want to make sure I understand the data structure change. I think it makes sense for the data and metadata to be grouped together in a single object - they should always change at the same time, right? if there were some use case for changing one but not the other then there might be a better alternative to prevent redundant computations (e.g. making sure changes are grouped in a single transaction when they occur together), but I think there probably isn't one? in which case this is good

daschi: Yes that's right, they'll always change at the same time since they're part of the same fetch response!

macfarlandian (approved):

jovergaag (approved):
🏑

Issue #655, opened 2020-12-16
[API] Filter original metric file based on query params and store subset in cache


jovergaag:
**What needs to be done? Why does it need to be done?**
Port the filtering logic from the FE to the server and make necessary modifications to filter the original dataset based on the subset_manifest. The naming convention and caching should already be established, so this is a matter of inserting the filtering logic into that process. At the end of this, the subset files in the Redis cache should truly be subsets of the file stored in the "original" cache for the same file.

Given the subset manifest, set the filter values to be equal to the subsets. Currently the filter logic only compares single value per dimension, but this will need to allow for more than one value. 

Given this subset manifest
`subset_manifest: { violation_type: [["absconded", "all", "elec_monitoring", "escaped", "high_tech", "low_tech", "med_tech", "municipal", "no_violations", "substance_abuse", "technical"], ["felony", "law", "misdemeanor"]] }`

if creating the file for the violation_type=["felony", "law", "misdemeanor"] subset, the filters will need to allow values where violation_type="felony" OR violation_type="law" OR violation_type="misdemeanor".

If a dimension is not listed in the subset_manifest, do not filter the data using that dimension and pass all data through.

Once the data has been filtered, update the metadata dimension_manifest to reflect the dimension values that exist in this file.

Save the data and metadata to the cache

**Additional context**
Add any other context about the task here.

PR #740, opened 2021-02-03
[API] Replace functions that are unsupported in node 10

daschi:
## Description of the change

This PR tries to fix the failing cron job we were seeing this past week. There are two issues it's trying to fix:

- Fetching all the data at once before filtering and caching every file's subset
- Errors on functions that are not supported in node 10

The next step is to see if this fixes the cron job. If it does, we want to undo the fix for fetching all the data at once and try to see if that really is causing the memory error we were seeing, or if it was just the unsupported functions causing a timeout. 

Side note -- I think we should update the `.nvmrc` to match the version we are deploying, thoughts @jovergaag ? Let me know if we should include that in this PR.

## Type of change

## Related issues

> Related to #655 

### Development

### Code review

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
     "build": "react-scripts build",
     "build-staging": "env-cmd -f .env.development react-scripts build",
     "test-react": "TZ=UTC react-scripts test",
-    "test-server": "TZ=UTC jest ./server",
+    "test-server": "TZ=UTC jest ./server --forceExit",
     "test-filters": "TZ=UTC jest ./shared-filters",
     "test": "yarn test-server && yarn test-filters && yarn test-react",
     "spa": "react-scripts start",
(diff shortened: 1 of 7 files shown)


macfarlandian (commented):
a) what were the unsupported functions? 
b) if this is so version sensitive then I agree you should update the `nvmrc` to match the GCP environment, I would just include that here

daschi:
@macfarlandian `Object.fromEntries` and `Array.prototype.flat` aren't supported in Node 10!

So we actually decided to cherry-pick the first commit here (that fixes those unsupported functions) into #728 because we realized the rest of the commits on the PR are unnecessary and weren't actually causing the memory error we were seeing. I'm going to close this PR and open a new one to see if we can update the `.nvmrc` version (it's a little complicated given that FE and BE are on different versions). 

Also currently discussing with Jen whether we should upgrade GAE to Node 12 now that it's recently fully supported.

Issue #711, opened 2021-01-23
Remove case type dropdown from PA Lantern dashboard


sychang:
**What needs to be done? Why does it need to be done?**
Remove the case type dropdown from the Pennsylvania Lantern dashboard.

If this happens before the calculate change to remove case types from the dashboard views, make sure case type = all 

**Additional context**
Do **not** remove the case type filter on the Missouri Lantern dashboard.

cc @jovergaag in case you want to add more details for this FE task.

jovergaag:
This will be changed in `filterOptions.js`. For PA, the case type filter will be disabled, and `defaultValue = "All"`, similar to the SUPERVISON_TYPE filter!

PR #743, opened 2021-02-04
Support for violation_type=ALL and reported_violations=ALL values

### LINKED ISSUE #641: Modify the filters to support the metric files with un-nestsed violation_type and reported_violations
Work is being done in #4533 to reduce the size of the metric files/views to reduce memory pressure on the client.

One of the results of this work is the un-nesting of violation_type and reported_violations in all of the views. The filters need to be changed to accommodate this change.

**Additional context**

jovergaag:
## Description of the change

This is in support of a breaking change in the views, and on its own is a breaking change. Must be deployed in production on Tuesday now that [#5529](https://github.com/Recidiviz/recidiviz-data/pull/5529) has been merged.

In support or reducing the size of the lantern views, we needed to unnest the `violation_type` and `reported_violations` dimensions, which means adding the 'ALL' values. This changes how the filters are run in the FE, as well as how the Matrix displays active cells. Updates fixtures to reflect this change.

#5529 (mentioned above) also removes all `charge_category` values except for 'ALL' for US_PA, so I disabled the corresponding filter, Case Type, in this PR as well. Issue #711 

## Type of change

## Related issues

> Closes #641 and #711

### Development

### Code review


server/core/demo_data/revocations_matrix_cells.metadata.json
```diff
@@ -1 +1 @@
...
..., "tcstl"]], ["level_1_supervision_location", ["01", "02", "03", "04", "04b", "04c", "04w", "05", "06", "07b", "07c", "07s", "08c", "08e", "08n", "08s", "09", "10", "10n", "10r", "11", "12", "13", "14", "15", "16", "17", "18", "19", "20", "21", "22", "23", "24", "25", "26", "27", "28", "29", "30", "31", "32", "33", "34", "35", "36", "37", "38", "39", "41", "42", "43", "44", "all", "ec", "ep", "era", "ppcmdctr", "slcrc", "tcstl"]], ["level_2_supervision_location", ["01", "02", "03", "04", "04b", "04c", "04w", "05", "06", "07b", "07c", "07s", "08c", "08e", "08n", "08s", "09", "10", "10n", "10r", "11", "12", "13", "14", "15", "16", "17", "18", "19", "20", "21", "22", "23", "24", "25", "26", "27", "28", "29", "30", "31", "32", "33", "34", "35", "36", "37", "38", "39", "41", "42", "43", "44", "all", "ec", "ep", "era", "ppcmdctr", "slcrc", "tcstl"]], ["metric_period_months", ["12", "3", "36", "6"]], ["reported_violations", ["1", "2", "3", "4", "5", "6", "7", "8"]], ["state_code", ["us_mo"]], ["supervision_level", ["all"]], ["supervision_type", ["all", "dual", "parole", "probation"]], ["violation_type", ["absconded", "felony", "misdemeanor", "municipal", "substance_abuse", "technical"]]]}
```
daschi, line 1: Should this have the `all` values for reported_violations and violation_type?

src/components/charts/new_revocations/__tests__/helpers.test.js
```diff
@@ -521,14 +521,38 @@ describe("applyMatrixFilters", () => {
...
+      },
     ];
   });
 
-  // For the applyMatrixFilters we do not need to worry about ALL values
-  // All values do not exist in violation_type and reported_violations fields
   describe("violationType filter", () => {
     let filteredViolationTypes = [];
 
+    describe("with violationType = 'ALL' filter applied", () => {
```
daschi, line 542: NOTE: we need to add the filtering logic and these tests in #728 as part of the `shared-filters` package after this merges

daschi (changes requested):
I have a couple suggestions/questions! I tested this using the demo_data fixtures and I noticed that the matrix is entering a `loading` state when you de-select one of the reported_violations filters, I think this is because we're missing the `all` values in the matrix cells metadata. Also, wanted to confirm that the fixture data is only a subset of the actual data exported from the query, as it seems the matrix chart is missing a lot of data.

src/RootStore/__tests__/BaseDataStore.test.js
```diff
@@ -144,14 +144,24 @@ describe("BaseDataStore", () => {
   });
 
   describe("when a filter value is not included in the dimension manifest", () => {
+    beforeEach(() => {
+      rootStore.filtersStore.setFilters({
```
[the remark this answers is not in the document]
jovergaag: You're right. it looks better now.

jovergaag (commented):
I don't know why i always put comments in a review...

server/core/demo_data/revocations_matrix_cells.metadata.json
daschi: Oooh okay, in that case I think we need to add `violation_type` and `reported_violations` to the `MatrixStore`'s list of `ignoredSubsetDimensions` (though maybe this should be renamed to just `ignoredDimensions` since this file will never have a subset)? That way we don't get a true value for `shouldFetchNewSubsetFile` because the "all" filter value won't be in the matrix store's dimension manifest.

daschi: Another thought I just had -- maybe we should add to the file's metadata whether or not it's a file that has subsets. That way in the frontend & backend we don't need to keep two lists of metrics that expect to have subsets...

jovergaag: Actually - I don't think that what I wrote earlier about the matrix never being able to subset due to the missing ALL values. I think weactually still could subset it - it would just never any all values that end up in that particular subset. Serena and I had decided when designing the new behavior that we never want the Matrix to reload because it would lead to a jarring user experience. I guess I will change the note to reflect that.

The caseload file also doesn't have ALL values - for any dimension. And we may want to subset it at some point.

> Another thought I just had -- maybe we should add to the file's metadata whether or not it's a file that has subsets. That way in the frontend & backend we don't need to keep two lists of metrics that expect to have subsets...
We don't keep that list on the FE do we? I only see it in `subsetManifest`. The FE only cares if the new filter value is in the current dimension manifest - it doesn't know which files are being subsetted and which aren't.

daschi: > We don't keep that list on the FE do we? I only see it in subsetManifest. The FE only cares if the new filter value is in the current dimension manifest - it doesn't know which files are being subsetted and which aren't.


No we don't need that list in the FE I guess...

 I still think we need to add `violation_type` and `reported_violations` as `ignoredSubsetDimensions` for the `MatrixStore` though, otherwise the `shouldFetchNewSubsetFile` will return true for the matrix whenever those values are "all".

jovergaag: Good point about adding to the `ignoredSubsetDimensions` for `MatrixStore`. Done

daschi (approved):
🍦  looks great! I think I would've passed in the `ignoredSubsetDimensions` as a constructor parameter in the MatrixStore and then assigned it in the BaseDataStore constructor since `ignoredSubsetDimensions` isn't an observable value, but this works too!

jovergaag:
> 🍦 looks great! I think I would've passed in the ignoredSubsetDimensions as a constructor parameter in the MatrixStore and then assigned it in the BaseDataStore constructor since ignoredSubsetDimensions isn't an observable value, but this works too!

That is what I had done initially, but didn't like it so I changed it to be the action, but I ended up adding it back in based on your comment!

src/RootStore/DataStore/MatrixStore.js
```diff
@@ -20,16 +20,20 @@ import { filterOptimizedDataFormat } from "../../utils/charts/dataFilters";
...
   VIOLATION_TYPE,
+  DISTRICT,
 } from "../../constants/filterTypes";
 
 export default class MatrixStore extends BaseDataStore {
   constructor({ rootStore }) {
     super({
       rootStore,
       file: `revocations_matrix_cells`,
+      additionalIgnoredSubsetDimensions: [
```
daschi, line 31: I think this should be updated to `ignoredSubsetDimensions`

PR #751, opened 2021-02-05
Set a min-height on chart containers when loading charts after a filter change

### LINKED ISSUE #724: Set min-height on charts to prevent the page from jumping on filter change
When we change a filter, and the FE needs to fetch new data, the charts show their loading indicator which has a smaller height than the charts themselves. This causes the page to jump around quite a bit. Set the min-height on the chart containers such that the height does not change while the new data is loading.

**Additional context**

daschi:
## Description of the change

Since we will soon be loading new data on certain filter changes, the varying heights of the loading states cause the page to jump around. This PR fixes that by adding a hook, `useContainerHeight`, which returns a ref for the chart's container, and the height of the chart container after it mounts. It also adds a new component, `LoadingChart`, which wraps the `Loading` component with a div that has the chart's container height.

Until #728 is merged, the charts will not re-load on every filter change. To be able to test this PR locally, you can add the following to the `BaseDataStore` constructor (or each individual data store) to test this interaction:

This reaction will toggle isLoading every time the filters change.

```
reaction(
      () => this.filters,
      () => {
        this.isLoading = true;
        setTimeout(() => {
          this.isLoading = false;
        }, 2000);
      }
    );
```  

## Type of change

## Related issues

> Closes #724

### Development

### Code review


src/components/charts/new_revocations/RevocationsOverTime.js
```diff
@@ -47,9 +48,10 @@ const RevocationsOverTime = () => {
   const { filters, dataStore } = useRootStore();
   const store = dataStore.revocationsOverTimeStore;
   const chartId = `${translate("revocations")}OverTime`;
+  const { containerHeight, containerRef } = useContainerHeight();
 
   if (store.isLoading) {
-    return <Loading />;
+    return <LoadingChart containerHeight={containerHeight} />;
```
jovergaag, line 54: I'm getting a console warning for this chart on initial load:

Screen Shot 2021-02-05 at 1 24 45 PM

src/components/charts/new_revocations/RevocationsByDimension/RevocationsByDimension.js
```diff
@@ -63,32 +65,34 @@ const RevocationsByDimension = ({
...
-      })}
-      modeSwitcher={
-        modes.length ? (
-          <ModeSwitcher mode={mode} setMode={setMode} buttons={modeButtons} />
-        ) : null
-      }
-      classModifier={chartId}
-      dataExportLabel={dataExportLabel}
-    />
+    <div ref={containerRef}>
```
jovergaag, line 68: Its nice that the by dimension charts maintain their size after a filter change, can we expand that to also include when changing from one chart to another? I know the chart size will change once the new chart is rendered because each chart is slightly different, but maybe the container can hold the size of the old chart, with the loading spinner, until the new chart is ready to be rendered?

jovergaag (commented):
Your thoughts on the `RevocationsByDimension` charts?

src/components/charts/new_revocations/RevocationsByDimension/RevocationsByDimension.js
daschi: Soo I looked into this a bit and the answer is yes -- but it will require a bit of restructuring and it involves using [a feature of `mobx-react-lite` that might become deprecated at some point ](https://github.com/mobxjs/mobx/issues?q=is%3Aissue+forwardref+) (though hopefully the new version should still support this, just in a different way).

I think we could get this UI effect by using `forwardRef` and forwarding the ref from the parent `RevocationCharts` component. Then the `containerHeight` would hold onto its value across the different chart renders. We would also need to move the `LoadingChart` and `Error` components into each of the dimension charts

Because of how `mobx-react-lite` and `forwardRef` work (or really don't work) together, this is how each of those dimension components will look:

```
const RevocationsByRiskLevel = observer(
  ({ containerHeight, timeDescription }, ref) => {
    const dataStore = useDataStore();
    const { revocationsChartStore } = dataStore;
    
    if (revocationsChartStore.isLoading) {
      return <LoadingChart containerHeight={containerHeight} />;
    }

    if (revocationsChartStore.isError) {
       return <Error />;
    }

    return (
      <div ref={ref}>
        <RevocationsByDimension {...all the props...} />
      </div>
    );
  },
  { forwardRef: true }
);
```

It's getting late on the east coast, so let me know what you think and I can work on implementing it next Monday!

daschi:
@jovergaag This is ready for another review! I updated it so that it maintains the height between filter changes & fetches!

jovergaag (approved):
:box

Issue #702, opened 2021-01-21
Move all revocation chart data filtering/sorting/generating into the DataStores


daschi:
**What needs to be done? Why does it need to be done?**
In [PR #694](https://github.com/Recidiviz/pulse-dashboard/pull/694) we began moving data fetching and filtering out of the Revocations chart components and into mobX Data Stores. There is still some data transformations that can be taken out of the components and moved to the respective chart data stores (i.e. `MatrixStore`, `CaseTableStore`, `RevocationsChartStore`, `RevocationsOverTimeStore`).

Moving all of the data transformations out of the components will centralize this logic and keep the components simple and "dumb". Each chart component should access the data it needs to render the chart, and should not have to do any further transformations to the data.  
 
**Additional context**
Notes by component:

`RevocationsByDimension` - this would include refactoring the `generateChartData` and `createGenerateChartData` so that this logic lives in the data store.   

`Matrix` - We want to move the violation types filtering and the generation of the dataMatrix into the store.

`CaseTable` - We want to move the sorting and formatting of the data into the store.

`RevocationsOverTime` - We should move the generation of `chartData` to the store, and possibly also the datasets/trendline dataset generation

Issue #590, opened 2020-10-30
Fetch error handling - write tests and pass error to handler rather than swallowing it


jovergaag:
**What needs to be done? Why does it need to be done?**
I don't think our fetch error handling is working the way that we think it is. Dig into the fetch responses and ensure that we are actually catching the errors, and passing them along to the handler correctly. Play around with the API responses to wee what all we are getting back.

**Additional context**
Add any other context about the task here.

PR #755, opened 2021-02-08
Add Sentry.ErrorBoundary to App and set some context / tags 

### LINKED ISSUE #589: Add intentional error reporting to Sentry for fetch errors and ErrorBoundary
Notify Sentry with fetch errors or when caught in an ErrorBoundary. We want to know about these. Check to see if there are any other errors that are being swallowed.

Also, ErrorBoundary does not catch errors that are inside of an event handler. Check to see if there are any other error prone operations happening inside of event handlers and add a try block.

**Additional context**
Data fetching is being done in the datastores for the new revocation charts

daschi:
## Description of the change

This PR adds the `@sentry/react` client package and configures a `Sentry.ErrorBoundary` that wraps everything at the `App.js` level and adds some additional context if available:

- currentTenantId
- filters
- user (currently sending entire user object, let me know if we should limit this information)

I want to make note of the types of errors that `ErrorBoundary` does not catch, and we may want to add some more error reporting context in other places (i.e. async requests, etc). 

**Error boundaries do not catch errors for:**
- Event handlers
- Asynchronous code (e.g. setTimeout or requestAnimationFrame callbacks)
- Errors thrown in the error boundary itself (rather than its children)

@jessex Just wanted to ask you security-wise, has Sentry been approved as a vendor and can we send user details in error reports?

This PR also comments out the Segment script for now while we test this PR, but for merging it we will follow this plan so we don't error out on the Segment/Sentry integration:

1. Deploy to staging for testing / QA
2. Uncomment out the Segment script
3. Merge PR
4. Remove the Sentry integration on Segment
5. Deploy to production

Steps 3-5 should happen the same day!

## Type of change

## Related issues

> Closes #589

### Development

### Code review


src/components/Revocations.js
```diff
@@ -62,9 +61,7 @@ const Revocations = () => {
               label="Time Period"
               dimension={METRIC_PERIOD_MONTHS}
             />
-            <ErrorBoundary>
-              <DistrictFilter />
-            </ErrorBoundary>
+            <DistrictFilter />
```
jovergaag, line 64: 😍

src/components/ErrorBoundary.js
```diff
@@ -17,36 +17,39 @@
...
+      scope.setUser(parsedUser);
     }
-
-    return children;
-  }
+    if (filters) {
+      const parsedFilters = Object.fromEntries(toJS(filters));
+      scope.setContext("filters", parsedFilters);
+    }
+  };
```
jovergaag, line 39: YAY DETAILS!

jovergaag: I don't actually see this coming through in the Sentry Error yet though...

src/index.js
```diff
@@ -37,6 +37,11 @@ configure({
   observableRequiresReaction: false,
 });
 
+Sentry.init({
+  environment: process.env.REACT_APP_SENTRY_ENV,
+  dsn: process.env.REACT_APP_SENTRY_DSN,
+});
+
```
jovergaag, line 44: Can this be done in `App.js`? We're initializing a bunch of other services there so it might make more sense to keep it in the same place.

.env.frontend.example
```diff
@@ -2,3 +2,5 @@ REACT_APP_API_URL={API_BASE_URL}
 REACT_APP_AUTH_ENV={AUTH_ENV}
 REACT_APP_FEEDBACK_URL={FEEDBACK_URL}
 REACT_APP_INTERCOM_APP_ID={APP_ID}
+REACT_APP_SENTRY_ENV={SENTRY_ENVIRONMENT}
```
jovergaag, line 5: So we will have a `staging` and a `production` environment such that we can keep track of errors separately. And just ignore errors that are neither of those two environments (like development) so we do away with the inbound/localhost filters. Is that what you had in mind?

Let's deploy this to staging tomorrow and set up the staging environment in Sentry to test out that env setup.

jovergaag (commented):
I'll have to get the env var from you and do more testing on this tomorrow.

One comment is that I forced a fetch error, and the error that was thrown was a `TypeError` instead of a 404 error. I had meant for this issue to have two requirements 1. Make sure that Sentry is notified in the ErrorBoundary (which this PR is fulfilling) and 2. check the response code after a fetch and throw an error if it is not 200. I'm sorry for not being more clear on that, I had written the issue thinking I would for sure be the one working on it and didn't include enough detail. Do you want to tackle that here or should we create another issue?

jessex:
Thanks for checking in, @daschi! In summary, Sentry received an assigned risk level of MEDIUM and was cleared to handle only anonymized user data -- nothing directly identifiable -- because Sentry is explicit that any customer data they receive _may_ be stored or processed outside of the United States. So we should not send any non-anonymized PII to Sentry.

What are you planning to send to Sentry with this PR? Let me know if you'd like to discuss this offline!

daschi:
@jovergaag Oooh, my mistake! I think the card is pretty clear actually, I just didn't read it closely enough yesterday. I can update this PR with the additional error reporting! I think I was getting that workflow mixed up with this issue https://github.com/Recidiviz/pulse-dashboard/issues/590 -- Is 590 supposed to be for the BE?

@jessex Okay good to know! I was thinking about whether it might be useful to send the user's email address to Sentry in case it can help debug issues, but maybe just sending the state code and filters state is enough. @jovergaag Do you have any thoughts about what additional context we might want to send to sentry for errors that I may have missed?

src/components/ErrorBoundary.js
daschi: I was testing this out a few different ways yesterday, but you can see the context & tags on this issue that came through:
https://sentry.io/share/issue/cad7ae0c7131417f897116a422002846/

The filters is getting set as context so that will show up under the `Details` tab, and the `currentTenantId` under the `Tags` tab.

jovergaag: Oh ok now I see them for that issue. Cool 😎

jovergaag:
> Do you have any thoughts about what additional context we might want to send to sentry for errors that I may have missed?

@daschi We will definitely want to be notified if there is an error fetching the restricted district or if `verifyRestrictedDistrict` fails because that results in the full page being blocked. So maybe send the restricted district if it exists - or in the case of verification not passing - send the malformed district id.

Other than that, the user information would be nice but not necessary given the security concerns.

jovergaag:
>  Is 590 supposed to be for the BE?

No you're right, I had meant #590 to be for the FE, so that we check `response.ok` in the metrics client. I think what you've done here is great, notifying Sentry in the catch blocks of the data stores. Then we can use #590 to check the response status and send sentry the response code.

src/components/Revocations.js
jovergaag: I got all excited about removing these `ErrorBoundary`s But I think we actually want to keep them. When I throw an Error in the Matrix component, we used to see the whole page load correctly except for the Matrix, which would show an error in place of the Matrix saying this chart can't load. Now, I get a full page error.

Screen Shot 2021-02-09 at 4 10 41 PM

jovergaag (changes requested):
Looking good! I'm so excited about the environments. 

I think there are two things remaining, remove the user information from the scope, and add back the ErrorBoundary around the chart components so that we only lose one chart if there is an error in that specific component.

jovergaag:
> Error boundaries do not catch errors for:
> Event handlers
> Asynchronous code (e.g. setTimeout or requestAnimationFrame callbacks)
> Errors thrown in the error boundary itself (rather than its children)

I think the last two are probably ok for now. The event handlers is one I've had on my mind for a while. All of our `RevocationsByDimension` chart data is being transformed in a `createGenerateChartData` callback. I just threw an error in the district `createGenerateChartData` and it was never caught - page rendered fine. All the more reason to move that logic into the data stores.

I think those are the only callbacks that look troublesome, at least on the Lantern side of things.

I think we should prioritize issue #702 to move all of the chart logic out of the components into the stores instead of writing code to handle errors specifically in the callbacks. I'll mark #702 for next sprint.

src/components/Revocations.js
daschi: ah yeah that makes sense, okay will add those back!

daschi:
> No you're right, I had meant #590 to be for the FE, so that we check response.ok in the metrics client. I think what you've done here is great, notifying Sentry in the catch blocks of the data stores. Then we can use #590 to check the response status and send sentry the response code.

@jovergaag Hmm, maybe we can chat about this more today, I didn't add the Sentry.captureException to `callMetricsApi` because I wasn't sure if we needed to capture it both at the fetch level and where we're actually calling it in the DataStore/useChartData try/catch blocks. I assumed it would report two sentry errors for the same exception. Maybe we could just set the tags in the scope at that level? I think we can see the API status code in the issue breadcrumbs already too.

jovergaag (approved):
👷‍♀️ 🚧

jovergaag:
When should we do the merge/deploys? I think it would make sense to do it sooner than later so that the remainder of the error reporting issues can leverage the new tooling. Thoughts?

daschi:
@jovergaag I could do it either right now or early tomorrow morning?

PR #766, opened 2021-02-12
Add dimension validations on the backend for new revocation files

### LINKED ISSUE #663: [API-fast-follow] Raise an alert if dimension values received in metric file do not match known values
Establish data models for each metric file, and when fetching the original metric file from the dashboard metric file bucket, compare the dimension manifest to the data models. If a value in the dimension manifest exists that is not in the data model, raise an alert.

**Additional context**
Part of this work should also include aligning the metric dimension values to the expected param values for the API in the params validations.

daschi:
## Description of the change

This PR adds a validation step when fetching files from GCS that will check whether the incoming new revocations files have any invalid or unexpected dimension values. This may happen if there was an issue with generating the data, and could point to having too many data points. The validation does not throw an error if there are values in the expected set that are not in the incoming source files. 

These validations will likely fail first in the `refreshCache` endpoint with the daily cron jobs. @jovergaag is planning to add the Sentry client to the backend to report these failures. 

Before deploying this branch, we will need to upgrade the node version in the gae-staging.yaml to the latest stable version for GAE which is 12. 

## Type of change

## Related issues

> Closes #663

### Development

### Code review


server/models/metrics/US_MO.js
```diff
@@ -0,0 +1,155 @@
...
+        violation_type: DEFAULT_VIOLATION_TYPES.concat(["all"]),
+      },
+    },
+    revocations_matrix_filtered_caseload: {
+      filename: "revocations_matrix_filtered_caseload.txt",
+      dimensions: {
+        charge_category: DEFAULT_CHARGE_CATEGORIES,
+        metric_period_months: DEFAULT_METRIC_PERIOD_MONTHS,
+        reported_violations: DEFAULT_REPORTED_VIOLATIONS,
+        supervision_type: DEFAULT_SUPERVISION_TYPES.concat(["all"]),
```
daschi, line 149: @jovergaag I think we shouldn't have "all" values in the case table, should I open an issue for this?

jovergaag: @daschi Oh yes, that is a bug. @jjdressel11 had noticed another bug that was causing duplicate rows in the caseload view yesterday afternoon. Maybe she can tackle the two of these together - or maybe they are even related.

jovergaag (changes requested):
YAY validations. We are close to being able to breathe easier!

Few comments, biggest one is around naming. Definitely open to chatting about it!

server/models/metrics/US_MO.js
daschi: This will be fixed by adding "EXTERNAL_UNKNOWN" instead of "ALL" for missing values with https://github.com/Recidiviz/recidiviz-data/pull/5887

server/models/BaseMetrics.js
```diff
@@ -0,0 +1,75 @@
+// Recidiviz - a data platform for criminal justice reform
```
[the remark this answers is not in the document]
daschi: I agree models is kind of a stretch here...we could try the collections/resources and see how that reads. The collections would then refer to `BaseMetrics` and `NewRevocationsMetrics`, where `BaseMetrics` is used for several of the collections. The "resources" are then descriptions of the data in the collections.

server/models/metrics/US_PA.js
```diff
@@ -0,0 +1,160 @@
...
+  "technical",
+];
+
+const DEFAULT_GENDERS = ["external_unknown", "female", "male"];
+
+const DEFAULT_SUPERVISION_TYPES = ["all"];
+
+const DEFAULT_CHARGE_CATEGORIES = ["all"];
+
+module.exports = {
```
jjdressel11, line 54: I tried to poke around your branch locally to understand how this is working, but I very much do not know javascript! However, I think there is too much duplication between these `.../metrics/US_PA.js` and `.../metrics/US_MO.js` files, and it will be unnecessarily tedious to maintain as we add new states and new dimensions. For each filename, the dimension _keys_ are the same between the two states. The only thing that changes between the states are the valid dimension _values_. Can you define all of these file -> dimensions mappings in `server/models/metrics/shared.js`, but have the instantiation of this object require a `state_code` to fetch the valid dimension values for the given state? (So then the only thing in the US_MO and US_PA files would be the definitions of each `DEFAULT_` list.)

jjdressel11 (changes requested):
Very excited about this!

src/components/Revocations.js
```diff
@@ -56,34 +56,39 @@ const Revocations = () => {
...
-              dimension={METRIC_PERIOD_MONTHS}
-            />
-            <ErrorBoundary>
-              <DistrictFilter />
-            </ErrorBoundary>
-            {filterOptions[CHARGE_CATEGORY].componentEnabled && (
-              <ToggleBarFilter label="Case Type" dimension={CHARGE_CATEGORY} />
-            )}
-            {filterOptions[SUPERVISION_LEVEL].componentEnabled && (
+          <>
```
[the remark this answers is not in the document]
daschi: It removes a console warning that `ErrorBoundary` is expecting 1 child element instead of an array, so this wraps the elements in a fragment.

server/models/metrics/US_PA.js
daschi: Yeah that's a good point, I'll try to reduce the duplication here!

src/components/Revocations.js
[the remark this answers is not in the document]
jovergaag: Oh bless you.

server/collections/resources/dimensions/index.js
```diff
@@ -0,0 +1,8 @@
+const { US_MO, US_PA } = require("../../../constants/stateCodes");
+const { US_MO_DIMENSIONS } = require("./us_mo");
```
jovergaag, line 2: We are validating the dimension values, not the dimensions. Should this be called `US_MO_DIMENSION_VALUES`?

jovergaag (commented):
Just two suggestions. I don't feel strongly about either of them so lmk if you like it the way that it is.

server/collections/resources/getCollections.js
```diff
@@ -0,0 +1,272 @@
...
+          reported_violations: dimensions.reported_violations,
+          supervision_type: dimensions.supervision_type,
+          supervision_level: dimensions.supervision_level,
+          violation_type: dimensions.violation_type,
+        },
+      },
+      revocations_matrix_distribution_by_officer: {
+        filename: "revocations_matrix_distribution_by_officer.txt",
+        dimensions: {
+          risk_level: dimensions.risk_level,
```
jovergaag, line 83: I missed this the previous review...risk_level doesn't exist on officer view.

server/collections/resources/dimensionValues/us_pa.js
```diff
@@ -0,0 +1,70 @@
...
+//
+// This program is distributed in the hope that it will be useful,
+// but WITHOUT ANY WARRANTY; without even the implied warranty of
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+// =============================================================================
+const {
```
jjdressel11, line 17: So much cleaner !!

server/collections/resources/index.js
```diff
@@ -0,0 +1,9 @@
+const { default: getCollections } = require("./getCollections");
+const { US_MO, US_PA, US_ND } = require("../../constants/stateCodes");
+
+exports.default = {
+  US_MO: getCollections(US_MO),
+  US_PA: getCollections(US_PA),
+  US_DEMO: getCollections(US_MO),
```
jjdressel11, line 7: confirming that demo should be `US_MO`?

jjdressel11 (approved):
GREAT work on this. Such an exciting step forward.

server/collections/resources/index.js
daschi: I chose it arbitrarily instead of creating a default metadata set for US_DEMO. It works with US_MO's values and if we ever need to change the metadata in the demo files to something different, then we could always create a default set for that.

jovergaag (approved):
✅ 🏁

PR #783, opened 2021-02-16
Set Redis cache expiration to 48 hours

### LINKED ISSUE #782: Increase redis TTL to 48 hours to allow room for data exports to fail

Today we realized a couple hours before the cache would refresh itself that the data export was going to fail and that we should prevent the cache from being refreshed. To avoid that last minute situation in the future, we're proposing to increase the default TTL on the redis cache from 24 hours to 48 hours. 

In case we need to stop the `refeshCache` cron job, we can delete the cron jobs and let the current cache persist an additional 24 hours. The `refreshCache` endpoint will continue to run daily at 1pm and 1:30pm EST for US_MO and US_PA, so as long as the cron jobs are re-deployed and scheduled, the data will still be refreshed. 

**Additional context**

daschi:
## Description of the change
Today we realized a couple hours before the cache would refresh itself that the data export was going to fail and that we should prevent the cache from being refreshed. To avoid that last minute situation in the future, we're proposing to increase the default TTL on the redis cache from 24 hours to 48 hours.

If this situation happens again, we would need to follow this process: 

1. Delete everything in the `cron.yaml` except for the `cron:` line
2. Deploy the cron jobs to staging or production 
3. When the data is fixed, re-deploy the cron jobs

The `refreshCache` endpoint will continue to run on a daily schedule at 1pm and 1:30pm EST for US_MO and US_PA. In case there's a mistake in the process and the cron jobs are not re-deployed after the data is fixed, the cache will expire after 48 hours and will depend on the first client request to refresh the cache. 

## Type of change

## Related issues

> Closes #782 

### Development

### Code review


jovergaag (approved):

jjdressel11 (changes requested):
I'm not sure that we want to do this. The export today would have failed the validation checks we have in place to avoid loading bad files into the bucket. When the export fails validations, the files in the bucket don't get overwritten and the cache re-fetches the same version of the file for a 2nd day in a row. Changing the cache expiration to 48 hours would mean that an export every other day never gets seen by the FE.

We should instead build up the backend validation to have more checks that ensure a bad export won't happen.

@ageiduschek, thoughts?

jovergaag:
> I'm not sure that we want to do this. The export today would have failed the validation checks we have in place to avoid loading bad files into the bucket. When the export fails validations, the files in the bucket don't get overwritten and the cache re-fetches the same version of the file for a 2nd day in a row. Changing the cache expiration to 48 hours would mean that an export every other day never gets seen by the FE.
> 
> We should instead build up the backend validation to have more checks that ensure a bad export won't happen.
> 
> @ageiduschek, thoughts?

Our thinking here was that when everything is running smoothly (including days when the validations fail and the export does not run) the cache will be refreshed using the data in the bucket. So that means it will refresh with either fresh data or the stale metric files in the bucket, and the data in the cache will never be more than 24 hours old.

The reason behind this change is to give us some breathing room on days where things are not running smoothly. For example, when the validations do not catch errors and the bad data ends up being exported. In that case, if we know about the bad export, we can disable the refresh cache cron job and allow the existing cache to give us more time while we troubleshoot or revert to known good versions of the files in the bucket. Hopefully this will NEVER be needed if the validations are tuned well.

I completely agree that we should build out the backend validations - this is just to give us more time to fix problems if the validations ever don't catch them.

ageiduschek (approved):
I feel ok about this as long as we still have a mechanism to refresh data every 24 hours

jjdressel11 (approved):
Ok I buy it! 🚢

PR #785, opened 2021-02-17
Load supervision locations into store for the SupervisionLocationFilter

### LINKED ISSUE #704: Fetch and save the districts for the DistrictFilter in a different store
We are currently fetching and filtering the data for the `DistrictsFilter` inside the `RevocationsOverTimeStore`. This is confusing and we should move the districts out of this store.   

We have a feature coming down the pipe where we will receive a file `supervision_location_ids_to_names.json` which will list all the district numbers and map them to their name to be displayed in the filter. When we implement that we will use that as the list of districts for the filter. Until then, parsing them from the `RevocationsOverTime` metric file is fine, but the logic for getting the districts in the store should live somewhere else.

We could either fetch and filter the districts data in its own `DistrictsStore`, or we could add it as part of the `FiltersStore`. Since it is a filter, it seems like it would fit with the `FiltersStore`, but we would need to see how that affects the UI interaction given that currently the `FiltersStore` is not fetching any data. 

We are already fetching and setting the districts data in one of the revocation chart's data stores, which is not ideal, but works for now. The alternative to using the FiltersStore would be to create a DistrictsStore, but I'm not sure if it really needs its own store since this data is only used to generate the filter options.

**Additional context**
See discussion: https://github.com/Recidiviz/pulse-dashboard/pull/694#discussion_r560511230

### LINKED ISSUE #759: Malformed restricted district ID in the response data does not trigger an auth error
**Describe the bug**

When the response data from `supevision_location_restricted_access_emails` includes an invalid district ID, it is not setting an authError and the user is able to load the dashboard and see some data.

**To Reproduce**
1. Add yourself to the `supevision_location_restricted_access_emails` staging file and give yourself a made up district name:
`{"state_code":"US_MO","restricted_user_email":"daniela+mo@recidiviz.org","allowed_level_1_supervision_location_ids":"not-real"}`
2. Spin up a local server with `yarn dev` (or view this on staging)
3. Login using the restricted user email 
4. See the dashboard!

**Expected behavior**
If a response data for `supevision_location_restricted_access_emails` include district IDs that are not in the tenant's list of districts, the user should see an Error page with an unauthorized error message. 

**Screenshots**
[Screen Shot 2021-02-09 at 5.52.04 PM.png](https://images.zenhubusercontent.com/5ee2e4197b73dcb817a96cfa/80d735be-e938-4245-a9e0-62e80c36cd57)

**Additional Context**
Related issue that can fix this bug: https://github.com/Recidiviz/pulse-dashboard/issues/704

Also related: https://github.com/Recidiviz/pulse-dashboard/issues/714

### LINKED ISSUE #786: Update filters to filter on level_1/level_2 supervision_location instead of district
The views will soon have additional `level_1_supervision_location` rows (other than 'ALL') for US_PA. To support this change, we need to start using `level_1_supervision_location` and `level_2_supervision_location` in the filters instead of district.

For US_PA - use `level_2_supervision_location=<districtFilterSelection>` and `level_1_supervision_location=ALL'
For US_MO use `level_2_supervision_location='ALL'` and `level_1_supervision_location=<districtFilterSelection>`

**Additional context**

daschi:
## Description of the change

This PR adds the `SupervisionLocationsStore` that will fetch supervision locations from the `supervision_location_ids_to_names` file in the buckets. It transforms the data into filter options for the SupervisionLocationFilter dropdown. 

I've also added these keys to the `*_translations` files which will make it easy for us to specify which values to use for filtering vs display:

`supervisionLocationValueKey` - the key that has the value to use for the filters
`supervisionLocationLabelKey` - the key that has the label for displaying in the filters
`supervisionLocationFilterKey` - the key we want to store the filter value as in the filters object (i.e. `levelOneSupervisionLocation`)

This also fixes a bug in the restricted access where malformed supervision location IDs were not throwing an auth error. 

**NOTE:** For now I've kept the logic for fetching the restricted supervision location in the `UserStore`. I could try to separate that out and add it to this store instead, but wanted to hear your thoughts about that first @jovergaag...we could also move it to a separate issue so that this is easier to QA and review.

## Type of change

## Related issues

> Closes #704 
> Closes #759 
> Closes #786

### Development

### Code review


jovergaag (approved):
🔹

src/RootStore/SupervisionLocationsStore.js
```diff
@@ -0,0 +1,105 @@
...
+    });
+  }
+
+  *fetchSupervisionLocations({ tenantId }) {
+    if (!this.rootStore?.tenantStore.isLanternTenant) {
+      this.isLoading = false;
+      this.isError = false;
+      return;
+    }
+    const endpoint = `${tenantId}/newRevocations/${this.file}`;
```
jovergaag, line 64: Just doing some QA and it looks like US_PA looks great but US_MO has a bunch of extra IDs that do not exist in any of the metrics rows. I'm checking with Anna about where these IDs came from and if we can get them filtered out in  the pipeline

jovergaag (changes requested):
Its working pretty good! Its the basis of a pretty big change so I think we should try to get the naming and organization to a place where we're really comfortable before merging. Let me know if you want to chat about it.

src/components/charts/new_revocations/ToggleBar/SelectDropdown.js
```diff
@@ -0,0 +1,79 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+// =============================================================================
+import React from "react";
+import PropTypes from "prop-types";
+import Select from "../../../controls/Select";
+import MultiSelect from "../../../controls/MultiSelect";
+
+const SelectDropdown = ({
```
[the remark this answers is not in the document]
daschi: Hmm yeah I wasn't sure if we really needed to, it helped me read the `SupervisionLocationFilter` and understand it more easily. I think it would make sense to move it to `src/controls` if you're okay with keeping it its own component. It could just become a generic component then for dropdowns that can have a (potentially optionally) disabled single value

src/components/__tests__/Revocations.test.js
```diff
@@ -59,15 +59,15 @@ describe("Revocations component tests", () => {
   const mockTenantId = "TEST_TENANT";
   const mockUser = { [metadataField]: { state_code: mockTenantId } };
   const toggleBarIdPrefix = "toggle-bar-";
-  const districtFilterId = "district-filter";
+  const SupervisionLocationFilterId = "district-filter";
```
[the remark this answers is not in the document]
daschi: I think I like number 2 as well as long as the UI label refers to it as `District`. If we end up changing the language on the UI to match Supervision Location, I could see us wanting to change back, but for now `District` is easier to parse and matches the UI so I think that makes sense!

src/RootStore/DistrictsStore.js
```diff
@@ -0,0 +1,96 @@
...
+        this.rootStore.currentTenantId
+      ) {
+        this.fetchDistricts({
+          tenantId: this.rootStore.currentTenantId,
+        });
+      }
+    });
+  }
+
+  *fetchDistricts({ tenantId }) {
```
jovergaag, line 57: This makes so much more sense to me as `districts` instead of `supervisionLocations`. I think this was the right choice.

src/RootStore/TenantStore/tenants.js
```diff
@@ -0,0 +1,41 @@
...
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+// =============================================================================
+
+import { US_MO, US_PA } from "../../views/tenants/utils/lanternTenants";
+
+const tenantMappings = {
+  districtValueKey: {
+    [US_MO]: "level_1_supervision_location_external_id",
+    [US_PA]: "level_2_supervision_location_external_id",
```
jovergaag, line 23: It will not be too far in the future that we will want to filter on both level_1 and level_2. Will this structure allow us that flexibility?

src/components/__tests__/Revocations.test.js
```diff
@@ -60,7 +60,7 @@ describe("Revocations component tests", () => {
   const mockTenantId = "TEST_TENANT";
   const mockUser = { [metadataField]: { state_code: mockTenantId } };
   const toggleBarIdPrefix = "toggle-bar-";
-  const districtFilterId = "district-filter";
+  const DistrictFilterId = "district-filter";
```
jovergaag, line 63: nit: I don't think this needed the capital D.

src/RootStore/FiltersStore.js
```diff
@@ -70,15 +73,51 @@ export default class FiltersStore {
...
+
+  get districts() {
+    const { apiData } = this.rootStore.districtsStore;
+    if (!apiData || !apiData.data) return [];
+    return uniqBy(apiData.data, this.districtKeys.valueKey)
+      .map((d) => ({
+        value: d[this.districtKeys.valueKey],
+        label: d[this.districtKeys.labelKey],
+      }))
+      .sort(compareStrings("value"));
```
jovergaag, line 130: I think that in order to get this out the door to unblock the data PR - if the state_code is US_MO filter it against the known set of districts that exist in prod right now. Then once we get that location ids file cleaned up we can remove it. Thoughts on that?

src/RootStore/__tests__/FiltersStore.test.js
```diff
@@ -74,6 +76,49 @@ describe("FiltersStore", () => {
         ]
       ).toStrictEqual([userDistrict]);
     });
+
+    it("clears the filters when switching tenants", () => {
```
jovergaag, line 80: Nice.

jovergaag (approved):

PR #795, opened 2021-02-19
Fix data export on iOS Chrome 

### LINKED ISSUE #622: Export data not working on iPad
**Describe the bug**
A PA user was unable to export data from any of the charts in the Lantern dashboard on iPad. They tried downloading each chart but saw nothing in their downloads folder. Not sure if this is WAI, given iPad. If it is WAI, we could consider adding documentation to 

This was the followup note from the PA user: "Good morning Serena. I was again unsuccessful in exporting from the dashboard while using Google Chrome version 86.0.4240.93 on an  I checked my download folders as advised and saw nothing." 

**To Reproduce**
1. On iPad, go to any chart on the Lantern PA dashboard
2. Click on the three dots and "export data"
3. Nothing happens

**Expected behavior**
Data should appear in a downloads destination on the ipad.

 - Device: [e.g. iPhone6]: iPad Pro (3rd generation)
 - OS: [e.g. iOS8.1]: unknown
 - Browser [e.g. stock browser, safari]: Chrome
 - Version [e.g. 22]: 86.0.4240.93

**Additional context**
Not sure if this is WAI, given iPad.

AlexK-ino:
## Description of the change

Fixed a bug in exporting data from the iPad.

The server/tmp directory is used for temporary files, which are deleted after downloading.

Previously tested possibility to do this without server, but all tried libraries and methods manually led to the same errors (WebKitBlobResource error 1 or Unknown file without extension)

That's why it was decided to use download in this way.

## Type of change

## Related issues

> Closes #622 

### Development

### Code review


jovergaag (changes requested):
Hi @AlexK-ino - the code looks pretty good - but I'm seeing bugs downloading files in Chrome. When trying to download the image file from the Matrix, I get an error. When downloading any other files, the .zip downloads successfully but OS X is not able to open the .zip.

Screen Shot 2021-02-19 at 5 35 15 PM

I have not yet tried this on iOS, let's get the traditional browsers working then i'll test on iOS!

Thank you,
Jen

AlexK-ino:
Hi Jen!
Corrected a bug in Matrix graphics 

On the broken zip files:
I am using OS X. I had this problem once before I created the PR. 
I have not been able to reproduce this problem at this time. If you get it, please send us the file, we will take a look at it. I, unfortunately, don't have it left.
Снимок экрана 2021-02-20 в 18 00 32

jovergaag:
@jessex I want to check in with you about the potential security concerns with this implementation. There is a [webkit bug](https://bugs.webkit.org/show_bug.cgi?id=167341) that [downloading files from the client in iOS](https://github.com/eligrey/FileSaver.js/#ios) is not possible until recent versions of iOS. Vlad, Alex, and I all did research into different packages and we came to the same conclusion, the only solution is to upload the file to the server in a tmp file and then download to the client.

This seems like a potential security risk, about the potential for malware to be injected into the payload before being uploaded to the server. Before we continue to work on the kinks in this PR, thoughts on whether it is a non-starter security wise?

jessex:
@jovergaag can you tell me more about that concern? Where does the data come _from_ before it goes to _our server_? Is it that the user presses the button on their client and then the data goes from their client, up to the server, and then back down to their client? If so, uploads-and-back-downloads for different users are not shared, right? So the concern would be limited to if someone hacked their client to upload malware targeting _our server_, as opposed to targeting a separate user of the app? Want to make sure I understand this clearly before proceeding.

jovergaag:
> @jovergaag can you tell me more about that concern? Where does the data come _from_ before it goes to _our server_? Is it that the user presses the button on their client and then the data goes from their client, up to the server, and then back down to their client? If so, uploads-and-back-downloads for different users are not shared, right? So the concern would be limited to if someone hacked their client to upload malware targeting _our server_, as opposed to targeting a separate user of the app? Want to make sure I understand this clearly before proceeding.

@jessex The data is coming from the javascript client. When the user presses the 'Export Data' button, the javascript bundles up the blob, then sends it to the server which then responds to the browser with a downloader url for the client to download from directly. iOS seems to not have a mechanism to download a file directly from a browser without trying to open it in a new tab. It can download it from a server via the url, but not from the web app. 

My concern is that the bundle that is created by the javascript could possibly be intercepted and injected with malware before landing on the server. Maybe? It is all handled within the Javascript and encrypted in transit so I believe it will be OK but am not positive.

There are not any [security advisories](https://github.com/expressjs/multer/security/advisories) for Multer which is the package we are using handle the file downloader POST request and responding with the downloader URL. 

This process is only used for iOS, other browsers will download directly from the javascript client.

jessex:
I think we're okay here, @jovergaag, because the client side code that gets triggered when you click the download button grabs data that had previously been downloaded into _memory_, so to corrupt the blob that's going to be sent back up to the server they would need to somehow corrupt memory/cache in the browser itself. As long as the actual upload back to the server is encrypted via TLS, then I think you can proceed here!

jovergaag:
@jessex Thank you!

@AlexK-ino I have done some more testing on this branch and the download is working as expected in demo mode, but the unsupported format error is only popping up in dev mode and when deployed to staging. I will do some investigation and push up a commit once I get that working. I'll circle around for your input when I'm finished with that. THANK YOU!

src/api/exportData/exportDataClient.js
```diff
@@ -0,0 +1,53 @@
...
+// MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+// GNU General Public License for more details.
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+// =============================================================================
+import JsFileDownloader from "js-file-downloader";
+import * as Sentry from "@sentry/react";
+
+export default async function exportDataClient(
```
macfarlandian, line 20: given its brittleness I kind of wish this function were ... not public, or named differently, or that it would actually throw an error or something for the browsers it is known not to support instead of just failing silently if it were actually to be used incorrectly? i am not totally sure though which, if any, of those options is best here

src/api/exportData/downloadData.js
```diff
@@ -0,0 +1,189 @@
...
+function downloadImage(filename, imageData) {
+  const jsFileDownload = new JsFileDownloader({
+    autoStart: false,
+    filename,
+    url: imageData,
+  });
+  jsFileDownload.start();
+}
+
+async function downloadZipFile({ files, filename, getTokenSilently }) {
```
macfarlandian, line 61: there is a truly miserable amount of prop drilling in this PR in service of this `getTokenSilently` function ... why can't this function retrieve it from the datastore directly?

src/api/exportData/transformCanvasToBase64.js
```diff
@@ -40,39 +51,4 @@ function transformCanvasToBase64(canvas, chartTitle, filters) {
...
-    const files = [methodologyFile, imageFile];
-
-    downloadZipFile(files, "export_image.zip");
-  } else {
-    downloadjs(imageData, filename, "image/png;base64");
-  }
-}
-
-export default downloadCanvasAsImage;
+export default transformCanvasToBase64;
```
jovergaag, line 54: I find it a little confusing how files are now split up between exportData and utils/downloads. The only thing actually using an API is exportDataClient. Should we leave that inside of api and move everything else back to downloads? Or is there another reason all of the other formatting/downloading is in api?

src/api/exportData/downloadData.js
```diff
@@ -0,0 +1,189 @@
...
+        forceDesktopMode: true,
+        autoStart: false,
+        filename,
+        url: `data:application/zip;base64,${content}`,
+      });
+      jsFileDownload.start();
+    });
+  } else if (isIE || isEdge) {
+    zip.generateAsync({ type: "blob" }).then(function (content) {
+      downloadjs(content, filename);
```
jovergaag, line 88: This is funny...IE11 requires the old download package? Wow what a run around.

src/api/exportData/downloadData.js
```diff
@@ -0,0 +1,189 @@
...
+    });
+  } else {
+    const content = await zip.generateAsync({ type: "blob" });
+    const formData = new FormData();
+    formData.append("zip", content, filename);
+    await exportDataClient(formData, filename, getTokenSilently);
```
jovergaag, line 94: What about desktop Chrome and Firefox and Safari? Do they need to hit the download api? I thought that we could get away with just downloading them directly and only hit the api for mobile?

jovergaag (changes requested):

server/routes/api.js
```diff
@@ -209,6 +212,53 @@ function programmingExplore(req, res) {
   );
 }
 
+function generateFileLink(req, res) {
+  const { file } = req;
+  const fileName = `${uuid.v4()}-${file.originalname}`;
+  const protocol = process.env.AUTH_ENV === "development" ? `http` : `https`;
```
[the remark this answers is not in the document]
daschi: Yeah it was part of it, this was causing the secure origin error, and then there was also the issue of trying to create a new `tmp/` folder on the GAE environment. You can't create a new directory but you can use the existing `/tmp` directory on the server.

server/routes/api.js
```diff
@@ -209,6 +212,53 @@ function programmingExplore(req, res) {
...
+  const fileName = req.params.name;
+
+  res.sendFile(fileName, options, (sendErr) => {
+    if (sendErr) {
+      throw new Error(
+        `Failed to send file for download: ${fileName}. ${sendErr.message}`
+      );
+    }
+
+    /* Checking on Chrome iOS */
```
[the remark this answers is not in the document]
daschi: No it's only not deleted if the header `upgrade-insecure-requests` is not in the request. I'm not sure why this was added here, so I wasn't sure if I should delete it. I **think** [maybe Chrome sometimes sends an additional request](https://web.dev/fixing-mixed-content/#upgrading-insecure-requests) when downloaded content to ensure security and maybe this is trying to ignore that extra request, but I couldn't reproduce that locally at all. I could remove it and re-deploy and see if there are any issues..

src/api/exportData/downloadData.js
daschi: Yes that's true...the datastores are only accessible in a react component because of how they're initialized in the `StoreProvider`....is there another way you found to share datastores ? It would be nice if this was just a download data component that we could render wherever we need it, but that felt like a big change to the US_ND dashboard, which I thought we wanted to just leave as is.

src/api/exportData/downloadData.js
daschi: I think so? I'll try one more time to make sure I didn't miss something when testing this with the other package..

src/api/exportData/downloadData.js
daschi: I didn't realize we only wanted to hit the API for mobile...I could try to rearrange this so that's how it's working

src/api/exportData/transformCanvasToBase64.js
daschi: I was trying to pull stuff out of `utils/downloads` because it was confusing me where the actual download was coming from. I really just wanted to try to put it all in one place, so I could put it back into utils/downloads..I guess it isn't really doing anything with an API so it doesn't make sense here.

src/api/exportData/downloadData.js
daschi: Ok confirmed that `JsFileDownloader` is not the right tool for downloading zip files on IE11.

server/routes/api.js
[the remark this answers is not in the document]
daschi: Hm okay I was able to recreate this - chrome sends this header on iOS when the user clicks on the download dialog, and if this check isn't here then the server will delete the file before the user has a chance to download it. I'm trying to think of how we can then delete this file after the user has downloaded it...

src/api/exportData/exportDataClient.js
daschi: I updated this to `exportZipDataOnMobileDevices` and added an error if it's not used for a mobile user - let me know what you think and if that helps.

src/api/exportData/downloadData.js
macfarlandian: oh yeah I see ... what I have been doing in Spotlight is just initializing the application's root store outside of the Provider and exporting it so I can access it outside of React. Then you'd have to make sure you were always calling these download functions from a reactive context. I don't know whether making that change as part of this would really be worth it?

src/api/exportData/downloadData.js
jovergaag: We could hit the API for desktop, but it seems to me that if the local download is working on desktop then we should avoid the additional RTT unless absolutely necessary.

jovergaag (commented):
This all looks great! I agree that the downloads directory is a big ball or yarn. Once we get into torch we can work on cleaning it up.

Code looks good, I'm going to hold off on approving until we can get this deployed to staging and tested on all the devices (hopefully this evening).

daschi:
@sychang and @jovergaag - The path of least resistance that I could find to make the chrome iOS download work for the US_ND dashboard was to also upload it as a zip file. Is that flow okay for mobile devices on the ND dashboard?

sychang:
Yes!

jovergaag (approved):
⬇️

PR #807, opened 2021-02-23
Fix nested filter component for admission type

### LINKED ISSUE #623: Update nested filter component
We currently have an Admission Type nested filter, and will soon be turning our District filter into a nested filter which to include district sub-offices.

The functionality is as follows: 
- The filter will have "All" selected as the default. 
- If the user selects the parent (e.g., 01 - ), all of the nested children will also be checked.
- If the user selects a nested child (e.g., Specialized), the nested child will be checked, and the parent will have a (-) in the checkbox.
- If the user then clicks the (-), the parent will be checked, and all of the nested children will be checked.
- If the user then clicks "All", "All" will be checked and everything else will be unchecked.

**Additional context**
See [here](https://public.tableau.com/views/HierarchicalSelections/2_BarsTreemap?:language=en&:display_count=y&:origin=viz_share_link) for similar nested filter functionality.

daschi:
## Description of the change

This PR updates the nested filter component to include the following fixes:

- "All" is the default value for the admission type filter 
- Fixed spacing for the grouped filter options
- When a group is partially selected, show a partially selected checkbox

Currently deployed to staging for QA on the PA dashboard.

## Type of change

## Related issues

> Closes #623

### Development

### Code review


src/components/Revocations.js
```diff
@@ -99,7 +99,7 @@ const Revocations = () => {
 
       <div className="bgc-white p-20 m-20">
         <ErrorBoundary>
-          <RevocationsOverTime />
+          <RevocationsOverTime timeDescription={timeDescription} />
```
jovergaag, line 102: Maybe we should just put the `timeDescription` in the `FiltersStore`

jovergaag (approved):
Beautiful

src/components/Revocations.js
daschi: Hmm yeah maybe..it is very presentational at the moment but that doesn't mean it couldn't live in a store. Maybe it would make sense to put it as part of the `BaseDataStore` since it is shared across all the charts.

PR #810, opened 2021-02-24
[API] Notify Sentry of Redis cache refresh errors

### LINKED ISSUE #683: [API - Fast Follow] Notify Sentry of errors in the API
When errors occur in the API, notify Sentry. Sentry can be integrated and notified directly from the BE. In the FE, the error reporting happens via Segment but this is not available in the BE so we need to integrate directly.

We want to be notified of Redis errors, fetch errors, etc.

https://docs.sentry.io/platforms/node/

**Additional context**

jovergaag:
## Description of the change

Notify Sentry when there is an error in `refreshRedisCache`.

I had considered notifying Sentry whenever the API response was an error (left the code in but commented - I can remove the comments if we agree to remove it)- but decided that the client will raise that error and we don't want to get notified twice from the same error. LMK if this doesn't seem like the right approach. The API response will cover all other logic in the API except the redis refresh I believe.

## Type of change

## Related issues

> Closes #683

### Development

### Code review


daschi:
If we do add the Sentry integration at the app/routes level, it might be helpful to be able to see the exception stack trace and could help us debug the backend more easily. I think it could be fine to see two errors for the same issue -- maybe we want to have an additional identifier or environment for the API environment vs the FE environment so we can filter on that?

jovergaag:
@daschi this is ready for another look.

daschi (approved):

PR #813, opened 2021-02-24
Single Select Redesign

### LINKED ISSUE #796: [CORE] Implement dropdown redesign
NOTE: All development and PRs for this task should be done against the branch `feature/core-header-refresh`.

The goal of this task is to update `Select.js` to match the design below

This is a single-selection dropdown.

The dropdown should collapse when clicking outside the menu, clicking on the header item, or pressing esc.

On hover, the background of the item under the cursor should darken.

The currently selected option should have a dark-green background with white text.

Mocks: https://www.figma.com/file/H08AQJSaAIRxfeMP4iCUPh/UP-MVP?node-id=1521%3A43

**pill**
border: 1px solid #D2D8D8;
height: 2.5rem

**item text**
weight: 500
size: 0.8rem
text-color: #00413E

**label text**
weight: 500
size: 0.8rem
text-color: #58706F

**selected**
background: #006C67;
text-color: #FFF;

**on hover**
background: #E3E6E6

MeloncholyShiro:
## Description of the change

I moved to .scss, but the library we use for drop-down (react-select) has problems working in css, can we still leave the CSS-IN-JS syntax? I can just rollback the commit

## Type of change

## Related issues

> Closes #796

### Development

### Code review


Catacola (commented):
Thanks for getting a start on this work. Which issue is this PR for?

src/components/controls/Select.scss
```diff
@@ -0,0 +1,59 @@
+.Select {
+  flex-grow: 1;
+  font-size: 0.8em;
+  font-weight: 500;
+  color: #00413e;
+
+  &__control {
+    border: 1px solid #d2d8d8 !important;
+    box-shadow: 0 0 0 1px #d2d8d8 !important;
+    border-radius: 15px !important;
```
Catacola, line 10: Are all of the `!important` labels necessary? If they are, that's no problem, but I like to only use this when necessary because it makes debugging future layouts harder.

Catacola (changes requested):
Looking good so far. I noticed a small issue with the top corners of the dropdown. I think there's still a border-radius applied to the top corners.

Screen Shot 2021-02-25 at 10 29 42 PM
Screen Shot 2021-02-25 at 10 38 28 PM

src/components/controls/Select.scss
MeloncholyShiro: > 
> 
> Are all of the `!important` labels necessary? If they are, that's no problem, but I like to only use this when necessary because it makes debugging future layouts harder.

Above in the description of the task, I wrote that react-select is not really friendly with .css / .scss files, and in any case applies its own styles, which can override either more specific (Will look strange), or !important, it is more friendly with css-in-js

Catacola: Got it! That makes much more sense. Thanks!

Catacola (changes requested):
From a design standpoint, this looks good. However, this change shouldn't clobber the existing dropdowns used in the Lantern views.
Screen Shot 2021-03-03 at 12 01 11 AM

I believe the best approach is to break this out into its own component instead of setting style flags. We can call it CoreSelect.js

jovergaag:
> I believe the best approach is to break this out into its own component instead of setting style flags. We can call it CoreSelect.js

I agree that this should become its own component. Maybe we need to start moving the repo to a top level "Core" and "Lantern" directory to house components under. We had tried to implement code splitting to reduce the bundle size (to keep IE happy) but there were so many shared components it didn't help much. If we do end up sharing fewer components, code splitting might end up making sense! We may want to wait for @daschi reorganization PR to decide what to do here.

daschi:
I agree we should create a separate select component for the core layout. The reorganization PR is up and ready for a review (https://github.com/Recidiviz/pulse-dashboard/pull/855), if we want to continue on this PR to add a new component, we should just make sure it is easy to move to the core/ directory when we need to resolve conflicts.

Catacola:
Looks good! I'll manage the merges tomorrow when I have time to deal with any conflicts that come up.

Catacola (commented):
This looks good to go. I'll start merging now. Thank you!

Catacola (approved):

Issue #812, opened 2021-02-24
Agent/officer chart QA


sychang:
**What needs to be done? Why does it need to be done?**
Notes from QA atop this doc: https://docs.google.com/document/d/1o1mJD0Cbb8B9M-Dt10uWkbPbhzbnSNN3XO9_gQ4bEmQ/edit

TLDR:
- A few potentially related data issues (e.g., districts don't appear to be mapping properly)
- A few minor, not launch-blocking visual tweaks

**Additional context**
Add any other context about the task here.

Issue #639, opened 2020-12-03
Enable Officer Chart


jovergaag:
**What needs to be done? Why does it need to be done?**
set enableOfficerChart to true to enable it once all features are complete!

**Additional context**
Add any other context about the task here.

PR #831, opened 2021-02-26
Officer chart finishing touches

### LINKED ISSUE #779: Use the new districtId-officerName label field for Agent Chart x-axis
Currently we are generating the x-axis label for the officer chart by concatenating the district ID with the officer name. When the `district='ALL'` filter is selected, this results in a label that does not make sense. [#4612 ](https://app.zenhub.com/workspaces/-lantern--5f2cdf04d4d9670013da86ed/issues/recidiviz/recidiviz-data/4612) is adding this new labels field. Use the new label field instead of generating it.

**Additional context**

jovergaag:
## Description of the change

Puts the finishing touches on the officer chart.

* Changes the x-axis labels to pull from `officer_label` - and formats the name correctly
* Only displays agent whose caseload is 10 or greater in the rates chart
* hides the rate line for now on the rates chart
* small styling tweaks
* updated methodology

QA doc here: https://docs.google.com/document/d/1o1mJD0Cbb8B9M-Dt10uWkbPbhzbnSNN3XO9_gQ4bEmQ/edit

## Type of change

## Related issues

> Closes #779 and #812 and #639

### Development

### Code review


src/utils/transforms/labels.js
```diff
@@ -104,6 +104,11 @@ const pluralize = (count, term) => {
   return count > 1 ? `${base}s` : base;
 };
 
+const formatOfficerLabel = (label) => {
```
daschi, line 107: This is throwing an error on both pages, I think that with the new data the officer name now has a colon instead of a dash.

daschi (changes requested):

src/utils/transforms/labels.js
jovergaag: Hmm, I'm not seeing this, the chart looks fine. Did you refresh your local cache? LMK if you are still seeing this error.

daschi: I am still seeing this on US_MO:

Screen Shot 2021-03-02 at 11 23 54 AM


and in the data from staging it looks like the officer names have a colon instead of a dash

daschi: Although I guess that might not be the issue since it's working fine in US_PA and the names have the same format!

daschi: It looks like the culprit is an officer that is missing the officer_label.

jovergaag: I still am not seeing this. What filters state are you using that is generating the error in MO? I also don't see any rows in the view where officer_label is null so I'm not quite sure where this is coming from. 🔍

jovergaag: @daschi ^^

daschi: Hm weird, seems to be working now - but I think we should put a guard here in case it happens again so that the chart doesn't crash... maybe this could just return an empty string or something if label is empty? Or filter the dataset for null officer_labels?

daschi: I guess the better solution would be to have this validation on officer_label be on the backend view.

daschi (approved):
This looks like it's working now! Just a suggestion to maybe guard against possible undefined officer labels so it doesn't crash the chart, and I also noticed a warning from the `PercentRevokedChart` propTypes, `datasets[].backgroundColor` is typed as a func instead of a string, not sure if you'd want to fix that in this PR.

PR #838, opened 2021-03-01
Fix district filter jump bug on IE11

### LINKED ISSUE #827: Selecting a district filter causes filter bar to jump and header to disappear
**Describe the bug**
In IE11 - ( have only seen in PA but might also exist in MO) - when selecting a district filter, the filters bar jumps down and the header disappears such that the chart appears behind.

**To Reproduce**
1. Go to 'the PA dashboard
2. Scroll down to view the district chart
3. Click on the district filter and select a filter
4. Now you can see the matrix above and behind the filters bar.

**Expected behavior**

**Screenshots**
Screenshots revealing sensitive information _must not_ be shared. If a screenshot is helpful in diagnosing and resolving the bug, please send this directly to a contact on the Recidiviz team.

 - OS [e.g. Windows 10]: 
 - Version [e.g. 22]: 

 - Device: [e.g. iPhone6]: 
 - OS: [e.g. iOS8.1]: 
 - Version [e.g. 22]: 

**Additional context**

MeloncholyShiro:
## Description of the change

The bug consisted of several things:
1) IE11 does not know what <main> is
2) React + IE11 doesn't know how to dynamically update inline styles

## Type of change

## Related issues

> Closes #827 

### Development

### Code review


Catacola:
Thank you for fixing my broken code. IE11 strikes again!

Catacola (approved):

src/components/Revocations.js
```diff
@@ -55,10 +56,13 @@ const Revocations = () => {
   );
 
   return (
-    <main className="Revocations">
+    <div className="Revocations">
       <Sticky
-        className="FilterBar"
-        style={{ zIndex: 700, top: hideTopBar ? 0 : 65 }}
+        className={cx("FilterBar", {
+          "Revocations__top-65": !hideTopBar,
```
jovergaag, line 62: Hm, these classnames are difficult to understand. Let's just add `Revocations--top-bar-hidden": hideTopBar`.

src/components/Revocations.scss
```diff
@@ -3,6 +3,14 @@
   padding-bottom: 5px;
   background-color: #f9fafb;
 
+  &__top-0 {
```
jovergaag, line 6: So here `Revocations--top-bar-hidden` we set `top: 0`, otherwise we set `top: 65` up in the `.Revocations` block above. Does that work?

jovergaag (changes requested):
Thank you @MeloncholyShiro ! Couple of classnames notes. Also I noticed that the padding above the profile dropdown disappeared on IE11.

Screen Shot 2021-03-03 at 4 01 07 PM

jovergaag (changes requested):
I think you can revert that last commit - it is doing very strange things to the animation. Sorry for leading you astray!

I reverted the commit locally but am still seeing a gap between the filters bar and the top of the window during the scroll in IE11. I think this still needs a bit of work.

MeloncholyShiro:
> I think you can revert that last commit - it is doing very strange things to the animation. Sorry for leading you astray!
> 
> I reverted the commit locally but am still seeing a gap between the filters bar and the top of the window during the scroll in IE11. I think this still needs a bit of work.
> 
> 

Okay, I'll work on that! I also wanted to ask. should I go through the whole project and remove inline styles and HTML Tags that IE doesn't support, like <main>?

jovergaag:
@MeloncholyShiro 

> should I go through the whole project and remove inline styles and HTML Tags that IE doesn't support, like < main >?

Not in this PR. I think let's just fix this header bug and I'll create another issue to fix the incompatible IE11 styles. Great idea!

Catacola:
> I reverted the commit locally but am still seeing a gap between the filters bar and the top of the window during the scroll in IE11. I think this still needs a bit of work.

I think that gap is due to the way I implemented hiding. The filter bar and nav bar are separate elements and they don't share a common parent. The top bar hides and the bottom bar changes it's snap position at a breakpoint or upon scrolling up. However, if we get to the top before the nav bar finished its animation, the filter bar will run into its snap position. The real fix to this will be to move the filter bar up into the nav bar and have that component slide on and off screen as a unit. This should be much easier thanks to Daniela's work splitting up this codebase.

Issue #828, opened 2021-02-26
When switching tenants rapidly - the districts filter does not populate with districts


jovergaag:
**Describe the bug**

**To Reproduce**
1. Go to MO dashboard and select one or two districts from the district filter
2. Switch to the PA dashboard and select a district from the district filter
3. Switch back to MO and click on the district filter and see that 'ALL' is the only option.

**Expected behavior**

**Screenshots**
Screenshots revealing sensitive information _must not_ be shared. If a screenshot is helpful in diagnosing and resolving the bug, please send this directly to a contact on the Recidiviz team.

**Desktop (please complete the following information):**
 - OS [e.g. Windows 10]: 
 - Browser [e.g. chrome, edge]: 
 - Version [e.g. 22]: 

**Smartphone (please complete the following information):**
 - Device: [e.g. iPhone6]: 
 - OS: [e.g. iOS8.1]: 
 - Browser [e.g. stock browser, safari]: 
 - Version [e.g. 22]: 

**Additional context**

PR #890, opened 2021-03-09
Use revocations_matrix_supervision_location_ids_to_names for districts filter list

### LINKED ISSUE #798: Use `revocations_matrix_supervision_location_ids_to_names` to populate list of districts for the District Filter
In [#5941](https://app.zenhub.com/workspaces/-lantern--5f2cdf04d4d9670013da86ed/board?labels=team%3A%20lantern&repos=138423816,108306686) the supervision_location ids will be filtered down to match the list of ids that exist in the metric files. Once this is done, we can use these IDs to populate the district filter instead of the filtered FE list.

**Additional context**

jovergaag:
## Description of the change

The `supervision_location_ids_to_names` file had an exhaustive list of supervision location ids, such that we had to use the ids from one of the metric files to filter the list down for the district filter to match the ids in the charts. There is a new mapping file, `revocations_matrix_supervision_location_ids_to_names` which is filtered to match the ids in the metrics files. Use that list of ids instead of the manually filtered list to populate the District Filter.

This has an added bonus of fixing the switching tenants bug #828. There was a timing between fetching the mapping file, and fetching/filtering the matrix which was used to filter the list of districts. 

Note: the `revocations_matrix_supervision_location_ids_to_names` is not yet in the production bucket, it should land today so do not deploy until verifying it has landed.

## Type of change

## Related issues

> Closes #798 and #828

### Development

### Code review


src/RootStore/DistrictsStore.js
```diff
@@ -77,6 +79,7 @@ export default class DistrictsStore {
         this.file,
         this.eagerExpand
       );
+      this.districts = this.apiData.data;
```
daschi, line 82: I'm not sure if we need to keep both references of `apiData.data` and `districts`, I think we can do with just one of these.

daschi (approved):
Yayyy so happy this fixed that other issue too! I just had one suggestion about the store but looks good to me.

Issue #777, opened 2021-02-15
Update UI for top-level navigation bar


sychang:
**What needs to be done? Why does it need to be done?**
This ticket involves the redesign of the current ND top-level navigation. This task involves only new UI and visual updates of the existing ND views - no changes to the actual page hierarchy.

Mocks: https://www.figma.com/file/H08AQJSaAIRxfeMP4iCUPh/UP-MVP?node-id=1521%3A43

Existing nav:
![image.png](https://images.zenhubusercontent.com/5ec1f17e7e863d6fd9e7e91c/5206dda9-a6ac-44f9-907e-ade815972414)

New nav: 
![image.png](https://images.zenhubusercontent.com/5ec1f17e7e863d6fd9e7e91c/04d03b2d-ab49-49b0-85fc-4711a6fcacb8)

On hover:
![image.png](https://images.zenhubusercontent.com/5ec1f17e7e863d6fd9e7e91c/593a3a96-9af9-4202-a7f6-166fd2ce94fd)

Dropdown (the options here should be "Facilities", "Community", and "Programming" to match what currently exists in ND instead of what is shown here):
![image.png](https://images.zenhubusercontent.com/5ec1f17e7e863d6fd9e7e91c/07ab2a55-d5d7-4da3-8930-845958b965d8)

The default page order should be: 
1. Facilities
2. Community 
3. Programming

When the dropdown is showing, the current page should be at the top of the list, followed by the other two pages in the default order. For example, if the user is currently on the "Community" page, the dropdown should be in this order: Community, Facilities, Programming. 

A new navigation bar enables:
- Extensibility to match UP's future structure
- Hierarchies the page more based on the user
- The navigation to be less obtrusive and take up less room
- Focus on accessibility

@soniachokshi to add more details

**Additional context**
CC @Catacola this could be a good FE starter task since it's UI polish building atop our current ND dash, unless it conflicts with how you are thinking through the overall infrastructure for UP.

jovergaag:
@sychang Is this assigned to @Catacola so that he can keep an eye on it or is he going to implement it? I just want to make sure this is still intended to be implemented by Inobright since it is in their queue. Thanks!

sychang:
To keep an eye on!

PR #891, opened 2021-03-09
Core design cleanup

Catacola:
## Description of the change

This diff incorporates all of the new components built for the leadership dash over the past few weeks and adds the fit and polish needed for the filter bar.

The design in this diff should match Mocks: https://www.figma.com/file/H08AQJSaAIRxfeMP4iCUPh/UP-MVP?node-id=1521%3A43

Things addressed in this diff:
  - Proper placement of filters on the nav bar
  - Proper width of filters
  - Coloring of dropdown filters
  - Size and placement of text within filters
  - Rename `FiltersBar` to `FilterBar`
  - Convert `FilterBar` and `CoreSelect` to `tsx`
  - Update tests
  - Add a new library to support enzyme in react 17
  - Refactor and flatten the div structure of `FilterBar` to make styling and reasoning more straightforward

Things not addressed:
 - Logo on the top bar
 - Placement of section selector on top bar
 - Media queries for responsive design

Since this is such a hodgepodge of little bits, I'm happy to jump on a call to talk over any parts.

## Type of change

## Related issues

Works towards https://github.com/Recidiviz/pulse-dashboard/issues/777

### Development

### Code review


src/core/__tests__/FilterBar.test.js
```diff
@@ -16,26 +16,30 @@
...
 import React from "react";
-import { queryAllByAttribute } from "@testing-library/dom";
-import { render } from "@testing-library/react";
-import FiltersBar from "../FiltersBar";
+import { configure, mount } from "enzyme";
+import Adapter from "@wojtekmaj/enzyme-adapter-react-17";
+import FilterBar from "../FilterBar";
 
-const getAllById = queryAllByAttribute.bind(null, "id");
+configure({ adapter: new Adapter() });
```
daschi, line 23: Should this go into a setupTests file so that it runs before all tests?

daschi (approved):
Woo - exciting seeing the design come together! I had one question about the enzyme configuration, and some design things I noticed below 👁️ 

**For the things addressed in this diff:**
1. Should we also add `cursor: pointer` to the dropdown options?
`.Core-Select__option:active, .Core-Select__option:hover`

**For things not addressed in this diff but I wanted to keep track of:**

1. I noticed on this class, there's a box shadow on the navigation dropdown that is defined with `rem` and I think we may actually want `px`, because the computed size is actually bigger than the design (1.6px):
`.CoreNavigation ul.nav-left .dropdown:hover`

2. The dropdown arrow looks a little cutoff on my screen:
Screen Shot 2021-03-09 at 4 45 16 PM

3. I think for the CorePageSelector options, we need to add a `color` property for this class so that it doesn't use the standard blue color when it's selected:
`.CorePageSelector--Option-Selected  { color: #25b894; }`

src/core/__tests__/FilterBar.test.js
Catacola: Brilliant! Done

Catacola:
> **For the things addressed in this diff:**
> 
> 1. Should we also add `cursor: pointer` to the dropdown options?
>    `.Core-Select__option:active, .Core-Select__option:hover`

I didn't even notice that before. Nice catch!

> **For things not addressed in this diff but I wanted to keep track of:**
> 
> 1. I noticed on this class, there's a box shadow on the navigation dropdown that is defined with `rem` and I think we may actually want `px`, because the computed size is actually bigger than the design (1.6px):
>    `.CoreNavigation ul.nav-left .dropdown:hover`

I'll address that in the next diff which will be all about the top bar. More generally, what's our rule around when to use px and when to use rem? I've been going with rem to 1 or 2 digits as much as possible and if something is off by 0.2px, so be it.

> 2. The dropdown arrow looks a little cutoff on my screen:
>    Screen Shot 2021-03-09 at 4 45 16 PM

Noted. Thanks!

> 3. I think for the CorePageSelector options, we need to add a `color` property for this class so that it doesn't use the standard blue color when it's selected:
>    `.CorePageSelector--Option-Selected  { color: #25b894; }`

Good catch! Will do in next diff.

daschi:
> More generally, what's our rule around when to use px and when to use rem? I've been going with rem to 1 or 2 digits as much as possible and if something is off by 0.2px, so be it.

I've been thinking about it in terms of the property it's used for, like in this example, I think we would want box-shadow to be more static, and not respond as much to screen changes. Properties that are used for spacing, font-sizes and element sizes I think should be much more responsive and we should use `rem` for those. Does that makes sense?

Catacola:
I see what you're saying, thanks! I'll keep that in mind as I go forward.

daschi:
@Catacola I forgot to mention in my review that I did have to run this for my server to compile, and just noticed that this is also causing a lint failure on travis:

`yarn add -D @types/react-select`

I'm not sure if that could be lending to the test failures too

Catacola:
Thanks for that tip! I had added it, but I think it got lost in a rebase.

For your fulfillment, it turns out that git was just ignoring one of my files. When I went to git mv it, git said it was untracked. When I went to add it, git said ok, and then said there were no new changes. I wanted to rename the file anyway since it is a core-only component, so I just copied the file to its new location, edited it there, and then deleted the old one. Funnily enough, git recorded it as a move from the old location and now the tests pass. This may be the weirdest git situation I've gotten myself into.

PR #925, opened 2021-03-17
Add filters for time period and gender to the projections dashboard

### LINKED ISSUE #933: Change filter options based on page view
The filter options on the projections dashboard will change depending on the page view (Community vs Facilities). We need to store the current page view in a shared store so that the filter options can be selected dynamically.

Use the tenant id to select filter options for the population filters

The alternative would be to use the `useLocation` hook from `react-router-dom` whenever we need to check the current page. The issue with this approach is that the hook is unavailable in the mobx filters stores, where we are doing the logic of selecting the filter options.

**Additional context**
Related to https://github.com/Recidiviz/recidiviz-data/issues/6094

daschi:
## Description of the change

This is the first step towards adding the filters bar to the projections dashboard. I decided to create a separate `CoreFiltersStore` to manage the filters state instead of following the current core pattern of `useState` in the components. I think this will help us when we transition to using mobx for all of core. 

I only added the filters for `timePeriod` and `gender` for now because those are what are available in the demo data. The only data that will show up in the filters is for timePeriod = 6 months and gender = Male for now. 

I posted a question about the remaining filters that we need to add in the linked issue:

- Legal Status
- Facility
- Custody Level
- Supervision Type
- Supervision Level
- District

## Type of change

## Related issues

> Closes https://github.com/Recidiviz/recidiviz-data/issues/6094
> Closes #933 

### Development

### Code review


src/RootStore/CoreFiltersStore.ts
```diff
@@ -0,0 +1,53 @@
...
+export type Filters = {
+  [k: string]: any;
+  timePeriod: string;
+  gender: string;
+};
+
+export type FilterKeys = keyof Filters;
+
+const defaultFilters: Filters = {
+  timePeriod: timePeriodOptions[4].value,
```
jovergaag, line 30: It might be nice to put a `defaultValue` getter on the `timePeriodOptions` object like we do in Lantern to avoid this index value here. Same for the next line.

src/RootStore/CoreFiltersStore.ts
```diff
@@ -0,0 +1,53 @@
...
+  gender: genderOptions[0].value,
+} as const;
+
+export default class CoreFiltersStore {
+  rootStore;
+
+  filters: Filters = defaultFilters;
```
jovergaag, line 37: Its not an `observable.map`. Yay we're learning :)

src/core/CoreFilterBar.scss
```diff
@@ -23,7 +17,7 @@
 
       &__title {
         white-space: nowrap;
-        color: #72777a;
+        color: #5d747e;
```
jovergaag, line 20: Should we add this to `colors.js`?

src/core/CoreFilterBar.tsx
```diff
@@ -137,14 +145,28 @@ const CoreFilterBar: React.FC<CoreFilterBarProps> = ({
...
+                  metricPeriodFilterOptions
+                )}
+                options={metricPeriodFilterOptions}
                 onChange={createOnFilterChange(setChartMetricPeriodMonths)}
                 defaultValue={defaultMetricPeriodOption}
               />
             </Filter>
           )}
 
+          {setGender && gender && (
```
jovergaag, line 159: I wonder if we should be guarding for `gender` and `setGender` in the component instead of on this line? It makes more sense to me to handle an undefined value or an undefined callback inside of the component than in the display logic.

jovergaag: same for all of these filters in this component

src/core/CoreFilterBar.tsx
```diff
@@ -70,12 +71,16 @@ type CoreFilterBarProps = {
...
+  setGender?: (value: string) => void;
+  gender?: string;
   availableDistricts?: string[];
   districtOffices?: {
     district: number;
     // eslint-disable-next-line camelcase
     site_name: string;
   }[];
+  genderOptions?: FilterOption[];
+  metricPeriodFilterOptions?: FilterOption[];
```
jovergaag, line 83: `replaceLa` lol. Maybe we should make an issue to do this in the data platform.

src/core/PageProjections/index.tsx
```diff
@@ -15,15 +15,34 @@
...
+import { observer } from "mobx-react-lite";
 import PageTemplate from "../PageTemplate";
 import PopulationSummaryMetrics from "../PopulationSummaryMetrics";
+import CoreFilterBar from "../CoreFilterBar";
+import { useCoreFiltersStore } from "../../components/StoreProvider/StoreProvider";
+import { timePeriodOptions, genderOptions } from "../utils/filterOptions";
 
 const PageProjections: React.FC = () => {
+  const filtersStore = useCoreFiltersStore();
+  const setFilters = (filterKey: string) => (value: string) => {
```
jovergaag, line 27: Why are you doing it this way instead of accessing the `filtersStore` in the `CoreFilterBar` directly?

src/core/PopulationSummaryMetrics/HistoricalSummaryMetrics.tsx
```diff
@@ -15,17 +15,22 @@
...
+import { useCoreFiltersStore } from "../../components/StoreProvider";
+import { formatTimePeriodLabel } from "../utils/timePeriod";
 import type { HistoricalSummaryRecord } from "../models/types";
 
 const HistoricalSummaryMetrics: React.FC<{
   data?: HistoricalSummaryRecord;
   isLoading: boolean;
 }> = ({ data, isLoading }) => {
+  const { filters } = useCoreFiltersStore();
+  const timePeriodLabel = formatTimePeriodLabel(get(filters, "timePeriod"));
```
jovergaag, line 30: Should this go into a store somewhere? I see the exact same line in `ProjectedSummaryMetrics`

jovergaag: Once we have the Metrics models - we can put the whole heading on the metric 🎉

jovergaag (changes requested):
Just a few things. Making progress!

src/RootStore/CoreFiltersStore.ts
```diff
@@ -0,0 +1,53 @@
...
+};
+
+export type FilterKeys = keyof Filters;
+
+const defaultFilters: Filters = {
+  timePeriod: timePeriodOptions[4].value,
+  gender: genderOptions[0].value,
+} as const;
+
+export default class CoreFiltersStore {
```
Catacola, line 34: From the naming here, is the idea that we will store all filters related to core features here? We already have filters for the existing core views and I can see more filters being adding with new features in the future. I think that's perfectly fine, but if this store is just for population-related things, can we name it as such?

src/core/CoreFilterBar.scss
```diff
@@ -2,14 +2,8 @@
...
 
 .CoreFilterBar {
-  padding-top: 1rem;
-  padding-bottom: 1rem;
   background: $core-background;
-  padding-left: 13.5rem;
-
-  @media screen and (max-width: $core-breakpoint-lg) {
-    padding-left: 4rem;
-  }
```
Catacola, line 12: Thank you for cleaning up my CSS! It looks like you're a few commits behind master. I don't think my changes will conflict, but I did make some small tweaks to this code to make everything play nice on IE11.

src/core/CoreFilterBar.tsx
Catacola: I believe the idea here is that we have a generic FilterBar component which displays any filters we need across the core views in the app. The change to TS required the double guard since we have no way currently to assure that gender is passed if setGender is passed. Now we could do some better type guarding by requiring an object with both `value` and `setter` keys for each filter so TS can guard against only getting just one or the other.

That said, I think we may be trying to cram too many options into this filter bar now. Would it make sense to have a separate filer bar for each view or set of views. E.g. `PopulationFilterBar`, `GoalsFilterBar`, and `CommunityFilterBar`, that all render using a more generic component like this under the hood? If the set of filters are close enough, we could reuse a bar between views, but I think this one may have gotten out of hand.

This change is almost certainly out of score for this diff, but it feel like a nice clean project for April clean up.

src/core/PageTemplate.js
```diff
@@ -23,23 +23,21 @@ import "../assets/styles/index.scss";
...
-    <main className="main-content PageTemplate">
-      <div id="mainContent">
-        <div
-          className="row gap-20 pos-r"
-          style={{ overflow: "initial !important" }}
-        >
-          {importantNotes.length !== 0 && (
-            <ImportantNotesBar importantNotes={importantNotes} />
-          )}
+    <div id="mainContent">
```
Catacola, line 28: This is such a better layout. Thank you!

That said, IE11 doesn't support \<main\> and other semantic tags because it hates fun: https://caniuse.com/?search=main

Catacola (commented):
This is coming along nicely! I suggest rebasing on master. There are a few styling changes around wrapping the filter bar on narrow screen I think you should pick up and may affect some of your styling.

src/RootStore/CoreFiltersStore.ts
daschi: Hm yeah my original thinking was that this could be expanded to manage the filters on all the views, but until we are ready for that I think it makes sense to name this to be specifically for PopulationProjections

src/core/CoreFilterBar.tsx
daschi: Agree it would be nice to separate it out so that we can get a clearer picture of what each view needs. I think for this PR I will create a more generic wrapper around the filter bar and separate it out, and also use the filters store directly in the component.

src/core/PageProjections/index.tsx
daschi: The `CoreFilterBar` just has a lot going on in there, but I think the solution here is to create a new filter bar for the population page until we can clean up the other views.

src/core/CoreFilterBar.scss
daschi: possibly? I'm not sure - I know soon we want to switch to using the component library for all typography and colors, so for now I am just adding them like this until we can get them from the library.

src/core/CoreFilterBar.tsx
daschi: Created a general filters refactor issue that we can expand on - https://github.com/Recidiviz/pulse-dashboard/issues/930

daschi: Oops, realized after I posted this that's not the issue you meant - making another one :)

src/core/PopulationSummaryMetrics/HistoricalSummaryMetrics.tsx
daschi: Yeah that would be good...I think it makes sense to move this to the filters store for now..

src/core/CoreFilterBar.tsx
jovergaag: Oh. ok. The normalizing logic is a little hairy. Maybe the issue is just to map the district name to the normalized district name. Or even add a metric file that has this has this mapping that can be requested from the API.

src/core/PopulationSummaryMetrics/ProjectedSummaryMetrics.tsx
```diff
@@ -18,14 +18,17 @@ import React from "react";
...
 
 const ProjectedSummaryMetrics: React.FC<{
   data?: ProjectedSummaryRecord;
   isLoading: boolean;
 }> = ({ data, isLoading }) => {
+  const { timePeriodLabel } = usePopulationFiltersStore();
+
   return (
-    <MetricsCard heading="Next 6 months" subheading="Projected">
+    <MetricsCard heading={`Next ${timePeriodLabel}`} subheading="Projected">
```
jovergaag, line 31: I'm getting a bunch of `Computed value CorePopulationFiltersStore@12.timePeriodLabel is being read outside a reactive context. Doing a full recompute` warnings on the page. This component (and others that use `timePeriodLabel) should be observers.

src/core/PopulationSummaryMetrics/index.tsx
```diff
@@ -61,17 +78,14 @@ const PopulationSummaryMetrics: React.FC = () => {
...
-  );
+  // Filter records
+  const historicalData = projectionSummaries
+    .filter(recordMatchesSimulationTag("HISTORICAL"))
+    .find(dataFilter) as HistoricalSummaryRecord;
 
-  const historicalData = historicalPopulationSummaries[0] as HistoricalSummaryRecord;
-  const projectedData = projectedPopulationSummaries[0] as ProjectedSummaryRecord;
+  const projectedData = projectionSummaries
+    .filter(recordMatchesSimulationTag("POLICY_A"))
```
jovergaag, line 87: Should `POLICY_A` this be a constant?

src/core/CoreConstants.scss
```diff
@@ -5,5 +5,10 @@ $core-breakpoint-md: 1440px;
 $core-breakpoint-sm: 1024px;
 $core-breakpoint-xs: 600px;
 $core-max-view-width: 93rem;
-
 $core-font-color: #00413e;
+$slate-80: rgba(53, 83, 98, 0.8);
```
daschi, line 9: This came from a chat with @soniachokshi, I wanted to update the filter title to a color in the design system, so she decided on Slate 80. Let me know what you think about adding the color names here to match the design system, I think that's the direction @ohaibbq is taking in the component library so I was hoping it would make things easier when we switch over to that!

jovergaag (changes requested):

src/core/PopulationSummaryMetrics/index.tsx
Catacola: So this is a bit of prep for the future. For now, we have one projection based on current policy. In not too long, we will start shipping a set of projections based upon different policy assumptions. At this point, this field will become relevant. Maybe for now, we can change it to "DEFAULT" or "NO_CHANGE" or anything else. I want to make it as easy to understand as possible.

src/core/CoreFilterBar.tsx
daschi: I tried to capture it here: https://github.com/Recidiviz/recidiviz-data/issues/6518

src/core/PopulationSummaryMetrics/index.tsx
daschi: Will the different policy projections be selected from a filter dropdown?

daschi:
After chatting with @Catacola on slack, I'm going to update this PR to include this feature: #933

jovergaag (approved):
🚴‍♀️

src/core/routes.ts
```diff
@@ -0,0 +1,53 @@
...
+  facilitiesProjections: "/facilities/projections",
+  methodology: "/methodology",
+  profile: "/profile",
+} as const;
+
+export function getViewFromPathname(pathname: string): keyof typeof CORE_VIEWS {
+  switch (pathname) {
+    case CORE_PATHS.communityExplore:
+    case CORE_PATHS.communityProjections:
+      return CORE_VIEWS.community;
```
jovergaag, line 40: This feels very cumbersome to me.  Maintaining both constants with the permutations of paths to views might get to be difficult to maintain as the number of tenants and views available to them grows. I don't know what the answer is, it just doesn't sit very well.

My mind thought a `PageStore` that had `view` getter on it would be easier to understand and use, instead of routes. Routes also feels misleading since this just accessing some constants and not doing any routing. But thats just my two cents...i dont feel super strongly about these!

src/core/utils/filterOptions.ts
```diff
@@ -0,0 +1,158 @@
...
+  { label: "Counts", value: METRIC_TYPES.COUNTS },
+  { label: "Rates", value: METRIC_TYPES.RATES },
+];
+export const defaultMetricType = metricTypeOptions[0].value;
+
+export const metricPeriodOptions = [
+  { label: "3 years", value: "36" },
+  { label: "1 year", value: "12" },
+  { label: "6 months", value: "6" },
+  { label: "3 months", value: "3" },
```
Catacola, line 37: 3 months shouldn't be an option for population projections, but 2 years should be.

src/core/PopulationSummaryMetrics/index.tsx
Catacola: > Will the different policy projections be selected from a filter dropdown?

The interface is TBD. This feature is is for V2. We will not allow users to select different policies when we launch. I could just change this tag to `"PROJECTED"` for now for all projected data.

src/core/utils/filterOptions.ts
```diff
@@ -0,0 +1,158 @@
...
+    },
+    get defaultValue(): string {
+      return this.defaultOption.value;
+    },
+    enabledViews: [CORE_VIEWS.community, CORE_VIEWS.facilities],
+  },
+  [FILTER_TYPES.GENDER]: {
+    type: FILTER_TYPES.GENDER,
+    title: "Gender",
+    width: "7rem",
```
Catacola, line 91: This one is completely my fault. I made a hack for the width because it was only on some options. Now that it's propagating, should we look into reworking the filter bar so this can be defined via CSS instead of in js? I don't want to slow you down, so just add this issue as a TODO and I'll cover it: https://github.com/Recidiviz/pulse-dashboard/issues/940

src/core/utils/timePeriod.ts
```diff
@@ -0,0 +1,42 @@
...
+  monthNumbers: string[],
+  abbreviated: boolean
+): string[] {
+  return monthNamesWithYears(monthNumbers, abbreviated, false);
+};
+
+export function formatTimePeriodLabel(months: string): string {
+  const numMonths = Number(months);
+  if (Number.isNaN(numMonths) || !months) return "";
+  if (numMonths === 1) return `${months} month`;
```
Catacola, line 34: Nit, no need for string creation on the fly:
```js
if (numMonths === 1) return "1 month";
```

src/core/utils/timePeriod.ts
```diff
@@ -0,0 +1,42 @@
...
+  if (numMonths < 12) return `${months} months`;
+  if (numMonths === 12) return `${numMonths / 12} year`;
```
Catacola, line 36: Ditto
```js
if (numMonths === 12) return "1 year";
```

src/core/routes.ts
Catacola: I agree that this can spiral out of control soon and feels a bit cumbersome. However, the mapping so far seems like it can be summarized as `"/$view/something" -> $view`. Would it be worth just reducing to that and then special-casing anything the the future.

I'm also not extremely set on one solution now. I'm often happy building something out with the knowledge that it won't scale, but that it works for now. The process of growing the app in other places will sometimes expose the "right" answer in a way we can't see now.

src/core/PopulationSummaryMetrics/index.tsx
```diff
@@ -30,18 +31,34 @@ import {
...
   isError: boolean;
   apiData: RawApiData;
 };
 
+function applyDataFilters(filters: PopulationFilterValues) {
+  return (record: PopulationProjectionSummaryRecords[number]) => {
+    return (
+      record.timePeriod === filters.timePeriod &&
+      record.gender === filters.gender &&
+      // TODO: Remove the check for "all" once fixture data is updated
```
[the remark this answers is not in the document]
daschi: Yes good idea, captured it here: #941

src/core/routes.ts
daschi: Yeah I see how it can become cumbersome. I can try to simplify this for now and rename the file to `views`, does that seem better?

src/core/utils/filterOptions.ts
daschi: This is for the Explore view filters, it just shows as being newly added because I changed this file to be TypeScript. The filter options used on the projections page is in the `PopulationFilterOptions` object.

jovergaag (approved):
💹

Catacola (approved):

PR #926, opened 2021-03-17
Fix issue with cache promises in refresh cache script, add more logs

daschi:
## Description of the change

This PR came from an investigation into the `refreshRedisCache` endpoint to try to find whether we were leaking memory or missing something. I found that using `concat` to create the `cachePromises` array was not modifying the existing array as I thought, but instead was creating a new array. This meant that any errors that occurred while caching the file were swallowed and we were not waiting for those promises before sending an http response. 

This fixes that issue and also adds some more logging so that we can continue to investigate that GAE instance issues.

## Type of change

## Related issues

> Closes [#XXXX]

### Development

### Code review


server/core/refreshRedisCache.js
```diff
@@ -50,31 +54,39 @@ function cacheEachSubsetFile(cache, cacheKey, fileKey, metricFile) {
...
       );
+      cachePromises.push(...subsetCachePromises);
     } else {
       console.log(`Setting cache for: ${cacheKey}...`);
       cachePromises.push(cache.set(cacheKey, metricFile));
     }
   });
-
+  console.log(
+    `Waiting for ${cachePromises.length} cache promises to resolve for ${cacheKeyPrefix}...`
```
jovergaag, line 88: Maybe we should add a log to the `finally` on line 111 so that this log isn't left hangin... It was a bit alarming to me when i first read that it was waiting without a resolution.

jovergaag (approved):
One one suggestion for an additional log. Other than that - I played around with the `refreshRedisCache` a while and really like the new logging.

server/core/refreshRedisCache.js
daschi: It's not that the `finally` wasn't being called because the `Promise.all` was resolving with the array of promises it was given, the array of promises it was given just did not include the promises for the subset caching. I'll add a log to line 111, I was going back and forth about it and I think it'll be helpful!

server/core/refreshRedisCache.js
```diff
@@ -109,6 +114,9 @@ function refreshRedisCache(fetchMetrics, stateCode, metricType, callback) {
       Sentry.captureException(message, responseError);
     })
     .finally(() => {
+      console.log(
+        `Finally responding to request and finished resolving ${cachePromises.length} cache promises.`
```
jovergaag, line 118: Yes this is exactly what I was thinking would be helpful. Thank you!

PR #931, opened 2021-03-18
Allow user to access views based on tenant id

### LINKED ISSUE #909: Restrict navigation to core dashboards based on tenant id
The core dashboard will have different views depending on the tenant ID. We need to update the routing in the front end app to restrict navigation to certain views based on the tenant id. 

**US_ND** should continue to route to the following views:
- `/community/explore`
- `/facilities/explore`
- `/goals`

- coming soon `/community/vitals` and `/facilities/vitals`

**US_ID** should route to:
- `/community/projections` 
- `/facilities/projections` 

**Additional context**

daschi:
## Description of the change

This PR adds a `ProtectedRoute` component for the CoreLayout that will only make the paths available to the tenant if they're in the `allowedPaths` list on the `tenants` constant. It also updates the CoreNavigation to use the tenant-specific navigation layouts.

## Type of change

## Related issues

> Closes #909 

### Development

### Code review

### DIFF
--- a/src/RedirectHome.tsx
+++ b/src/RedirectHome.tsx
@@ -15,35 +15,19 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 // =============================================================================
 import React from "react";
-import { Route, useLocation } from "react-router-dom";
+import { Redirect } from "react-router-dom";
 import { observer } from "mobx-react-lite";
-import NotFound from "./components/NotFound";
 import { useRootStore } from "./components/StoreProvider";
 import { getPathsFromNavigation } from "./utils/navigation";
 
 import tenants from "./tenants";
 
-interface ProtectedRouteProps {
-  component: React.FC;
-  children: React.ReactNode;
-}
-
-const ProtectedRoute: React.FC<ProtectedRouteProps> = ({
-  component: Component,
-  ...rest
-}) => {
+export const RedirectHome: React.FC = ({ ...rest }) => {
   const rootStore = useRootStore();
-  const { pathname } = useLocation();
   // @ts-ignore
   const tenant = tenants[rootStore.currentTenantId];
   const allowedPaths = getPathsFromNavigation(tenant.navigation);
-
-  if (!allowedPaths.includes(pathname)) {
-    return <NotFound />;
-  }
-  return (
-    <Route {...rest} render={(props) => <Component {...rest} {...props} />} />
-  );
+  return <Redirect {...rest} exact from="/" to={allowedPaths[0]} />;
 };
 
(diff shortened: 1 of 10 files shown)


src/tenants.js
```diff
@@ -27,6 +28,26 @@ export default {
   [core.US_ND]: {
     name: "North Dakota",
     availableStateCodes: [core.US_ND],
+    allowedPaths: ["/community/explore", "/facilities/explore", "/goals"],
```
Catacola, line 31: Is this field necessary? It seems like everything here can be computed from the `navigation` object below.
```js
allowedPaths = Object.entries(tenants[state_core].navigation).flatMap(e => {
  const section = e[0];
  const pages = e[1];
  return pages ? pages.map(p => `/${section}/${p}`) : [`\${section}`];
});
```

src/tenants.js
```diff
@@ -27,6 +28,26 @@ export default {
...
+      facilities: ["explore"],
+    },
+  },
+  [core.US_ID]: {
+    name: "Idaho",
+    availableStateCodes: [core.US_ID],
+    allowedPaths: [
+      "/community/projections",
+      "/facilities/projections",
+      "/methodology",
```
Catacola, line 44: Computing allowed paths from the navigation element will allow for us to choose whether to include `/methodology` in one place instead of having to gate in two places. Right now, this PR will remove methodology from the dropdown but allow users to manually navigate there (which isn't a huge deal but it probably not the functionality we were lookin for).

Catacola (changes requested):
Looking really good. I just have one question about storing and using routes.

src/tenants.js
daschi: Ahh yeah, that makes sense! I made the update - and also updated the redirect home to work with the different tenants.

src/RedirectHome.tsx
```diff
@@ -0,0 +1,33 @@
...
+import { observer } from "mobx-react-lite";
+import { useRootStore } from "./components/StoreProvider";
+import { getPathsFromNavigation } from "./utils/navigation";
+
+import tenants from "./tenants";
+
+export const RedirectHome: React.FC = ({ ...rest }) => {
+  const rootStore = useRootStore();
+  // @ts-ignore
+  const tenant = tenants[rootStore.currentTenantId];
```
Catacola, line 28: It's going to be a big lift to move all of the stores to ts, but it'll help with all of these ts-ignore tags we need to sprinkle in.

https://github.com/Recidiviz/pulse-dashboard/issues/935

Catacola (approved):
🔐

PR #934, opened 2021-03-19
Create restrictedDistrict subsets

### LINKED ISSUE #749: [placeholder] Change district-restrict access enforcement from app-level to API-level
Following up from (discussion in launch checklist ticket [145](https://app.zenhub.com/workspaces/-lantern--5f2cdf04d4d9670013da86ed/issues/recidiviz/zenhub-tasks/145))

The approach we've taken to implementing the district-restricted access is that we are not making an API-level change but an app-level change. All data points, across all districts, are still being fetched by the app and brought into memory/cache. But then the app is limiting which data points the user can see if they should be limited.

This means that the user's computer does have data from other districts (albeit temporarily) and that a technically savvy user could inspect the API requests to look at other districts' data (although, the requests return data in a fairly opaque, optimized format that the user would then need to learn how to decipher).

This was not blocking for the initial launch, but the API-level change is supportable with the recent API changes (shoutout), so this is a placeholder ticket to make that change.

**Additional context**
Whether we prioritize this depends on a longer-term solution to enable org-chart-based views across the DOC. Scoping for that in progress.

jovergaag:
## Description of the change

Creates subsets based on restricted district. 
* Adds restricted district t the query params
* Creates a subset, caches the subset and responds to the network request with the data filtered based on the level_1_supervision_location.

TODO:
- [ ] ~~Add restricted district to refreshCache job? - Currently the response time when you first hit the cache with a restricted district param is quite slow. Should we pre-populate the cache for every subset for every district? The logic for this will be quite complex I think because we will need the response of the supervision location mapping file to know the district ids. And those are asynchronous fetches. We could just get the current cached value for that file and use those districts...~~

## Type of change

## Related issues

> Closes #749

### Development

### Code review


server/utils/cacheKeys.js
```diff
@@ -126,23 +126,29 @@ function getCacheKey({
...
-  getSubsetManifest().forEach(([dimensionKey, dimensionSubsets]) => {
-    const subsetValue = cacheKeySubset[camelCase(dimensionKey)];
-    const subsetIndex = dimensionSubsets.findIndex(
-      (subset) => subsetValue && subset.includes(subsetValue)
-    );
-    if (subsetValue && subsetIndex >= 0) {
-      cacheKey = `${cacheKey}-${dimensionKey}=${subsetIndex}`;
-    }
-  });
+  if (cacheKeySubset && FILES_WITH_SUBSETS.includes(metricName)) {
```
daschi, line 131: I'm not sure it makes sense to check for `cacheKeySubset` after we already de-structure the object on line 129.

server/utils/cacheKeys.js
```diff
@@ -126,23 +126,29 @@ function getCacheKey({
...
+        (subset) => subsetValue && subset.includes(subsetValue)
+      );
+      if (subsetValue && subsetIndex >= 0) {
+        cacheKey = `${cacheKey}-${dimensionKey}=${subsetIndex}`;
+      }
+    });
+  }
 
-  return cacheKey;
+  return appendRestrictedDistrictKey(cacheKey, restrictedDistrict);
```
daschi, line 143: Just thinking aloud here...but maybe we can avoid making this code super specific to US_MO needs by storing state-specific subset manifests in the data bucket that can be cached. Then based on the stateCode and the filters that come in the request we can select the subset manifest to use, and there can be one specifically for US_MO that lists all of the supervision locations as their own subsets.

daschi (commented):
Just a few initial thoughts, happy to chat more about this on slack! 

Re: caching all the subsets for the restricted districts, I think we could wait for the restricted access file to be fetched & cached, and then use that to get the unique location IDs that we would need to create subsets for. The restricted access file would only need to be fetched and cached once and then can just be accessed for every file that gets subsetted/cached.

server/filters/createSubset.js
```diff
@@ -117,23 +130,35 @@ function applyFiltersToDataPoints(dataPoints, filters, filterFn) {
...
+
+  if (
+    !FILES_WITH_SUBSETS.includes(metricName) &&
+    !levelOneSupervisionLocation
+  ) {
     return metricFile;
   }
 
-  const filterFn = getFilterFnByMetricName(metricName, subsetFilters);
+  const filters = getFiltersByMetricName(metricName, subsetFilters);
```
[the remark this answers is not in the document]
jovergaag: Great suggestions!

server/filters/createSubset.js
```diff
@@ -117,23 +130,35 @@ function applyFiltersToDataPoints(dataPoints, filters, filterFn) {
...
+  const filterFn = getFilterFnByMetricName(metricName, filters);
+
+  const skipFilterFn = (dimensionKey) =>
+    !getSubsetDimensionKeys()
+      .concat(["level_1_supervision_location"])
```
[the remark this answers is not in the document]
jovergaag: Yes!

server/filters/subsetFileHelpers.js
```diff
@@ -96,13 +96,15 @@ function createFlattenedValueMatrix(filteredDataPoints, subsetMetadata) {
...
+  const subsetKeys = getSubsetDimensionKeys().concat(
+    "level_1_supervision_location"
+  );
   const transformedDimensionManifest = [];
   dimensionManifest.forEach(([dimensionKey, dimensionValues]) => {
     if (subsetKeys.includes(dimensionKey) && subsetFilters[dimensionKey]) {
       transformedDimensionManifest.push([
         dimensionKey,
-        subsetFilters[dimensionKey].sort(),
+        subsetFilters[dimensionKey],
```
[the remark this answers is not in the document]
jovergaag: Good catch. I had removed it at a point prior to ensuring `level_1_supervision_location` is an array...fixed.

server/utils/cacheKeys.js
```diff
@@ -126,23 +126,29 @@ function getCacheKey({
     cacheKey = `${cacheKey}-${metricName}`;
   }
 
-  if (!cacheKeySubset || !FILES_WITH_SUBSETS.includes(metricName)) {
-    return cacheKey;
-  }
+  const { restrictedDistrict } = cacheKeySubset;
```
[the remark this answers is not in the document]
jovergaag: No. US_PA should never have a `restrictedDistrict` param, in that case this this object is undefined and the cacheKey is not appended.

server/utils/cacheKeys.js
jovergaag: I ended up changing this section a little - hopefully it makes more sense now.

server/utils/cacheKeys.js
jovergaag: I think thats a good idea. That above approach might be a bit complex because some users have a restricted district and others don't - so we would need to add logic around layering the subset dimensions in a way that some subsets have the restrictedDistrict treatment and others don't.

In the spirit of getting this out quickly to move on to the Torch launch, maybe we leave it as is for now and when the second state that requires its own unique `subset_manifest` we can add in the flexibility at that point.

jovergaag:
@daschi This is ready for another review

> Re: caching all the subsets for the restricted districts, I think we could wait for the restricted access file to be fetched & cached, and then use that to get the unique location IDs that we would need to create subsets for. The restricted access file would only need to be fetched and cached once and then can just be accessed for every file that gets subsetted/cached.

I have been doing some QA with @sychang and we decided that the time investment in warming the cache with the restricted-district subsets is not a priority. The added latency of the BE generating the subset on demand is not significant enough to merit the additional time to make the necessary changes to `refreshRedisCache`.

server/utils/cacheKeys.js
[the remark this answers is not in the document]
daschi: Ah okay - I think this would cause an error then if we try to destructure a property that doesn't exist on the `cacheKeySubset`

daschi (changes requested):
There's an issue that I'm seeing on the frontend - if you start with a restricted district on US_MO and then switch with that same user to US_PA, it still makes requests to the API with the restrictedDistrict in the query param and the API response returns 0 data points for all files.

jovergaag:
> There's an issue that I'm seeing on the frontend - if you start with a restricted district on US_MO and then switch with that same user to US_PA, it still makes requests to the API with the restrictedDistrict in the query param and the API response returns 0 data points for all files.

Good find. Fixed.

.travis.yml
```diff
@@ -2,7 +2,7 @@ language: node_js
 before_install:
   - ./travis-setup.sh
 install:
-  - yarn install
+  - yarn install --force
```
daschi, line 5: We might wanna keep an eye on this as it might make our builds a bit slower

src/RootStore/DataStore/helpers.js
```diff
@@ -40,9 +60,12 @@ export function dimensionManifestIncludesFilterValues({
       return true;
     }
     if (dimensionManifest[filterType] === undefined) {
-      throw new Error(
+      // This should only occur with the JSON metric data (never in production).
+      // When fetching optimized data, this logic branch should not be possible
+      console.error(
```
daschi, line 65: Was this causing an error with json demo data?

src/RootStore/DataStore/BaseDataStore.js
```diff
@@ -148,7 +148,10 @@ export default class BaseDataStore {
   }
 
   get filtersQueryParams() {
-    return getQueryStringFromFilters(this.filters);
+    return getQueryStringFromFilters(
+      //  eslint-disable-next-line
+      this.filters, [this.rootStore.restrictedDistrict],
```
daschi, line 153: Why enforce this style?

daschi (approved):
A few more questions but not merge-blocking..

src/RootStore/DataStore/helpers.js
jovergaag: It throws an error when using JSON data rather than optimized data - because we don't have the whole `dimension_manifest` for those files, only what can be built from the `subset_manifest`.

src/RootStore/DataStore/BaseDataStore.js
jovergaag: Changed!

.travis.yml
jovergaag: I removed the `--force` and re-ran the build on travis, which was successful. But I'm worried that without it anytime there is a change in shared-filters we will have this problem again. Or even on master, will master fail to build?

Issue #908, opened 2021-03-15
Use typography styles from component library on projections dashboard


daschi:
**What needs to be done? Why does it need to be done?**
Once the [Component Library](https://github.com/Recidiviz/case-triage/issues) is updated to include all the system typography styles and components, we need to update the core projections dashboard to use these styles directly from the library. 

This includes updating the typography styles (headers, body, and UI text), the icons, and anything else that has been added to the component library that is used on the dashboard.

**Additional context**
There's [an existing issue](https://github.com/Recidiviz/case-triage/issues/69) in the component library to add a rotation prop to the caret icon.

PR #939, opened 2021-03-22
Add timeseries chart to population view

Catacola:
## Description of the change
This PR introduces semiotic to pulse-dash. The chart is styled to look like https://www.figma.com/file/SOdY6JujMpbawgJ4jjDmE8/Population-Projection-MVP?node-id=711%3A0.

Screen Shot 2021-03-22 at 9 20 36 AM

Known issues to address in future PRs:
- Tooltips
- Aligning padding/margin between chat and summaries
- There are a lot of helper functions in the chart file and there are abstraction that could be made for future use. There will be a organization and abstraction diff later -- I find it much easier to develop a new component in a blob like then and then extract the reusable parts when I'm done.

@soniachokshi - Temporary app for testing has been deployed here: https://recidiviz-dashboard-stag-e1108--pop-chart-c0o26tho.web.app/community/projections

## Type of change

## Related issues

80% of https://github.com/Recidiviz/recidiviz-data/issues/6097

### Development

### Code review


src/core/PopulationSummaryMetrics/index.tsx
```diff
@@ -17,31 +17,26 @@
...
-import {
-  recordMatchesSimulationTag,
-  populationProjectionSummary,
-} from "../models/PopulationProjectionSummaryMetric";
+import { recordMatchesSimulationTag } from "../models/PopulationProjectionSummaryMetric";
 import "./PopulationSummaryMetrics.scss";
 
-type ChartDataType = {
-  isLoading: boolean;
+type propTypes = {
```
macfarlandian, line 29: maybe we don't have linting for this but type names should be pascal case?

src/core/PopulationTimeseriesChart/PopulationTimeseriesLegend.tsx
```diff
@@ -0,0 +1,43 @@
...
+};
+
+const PopulationTimeseriesLegend: React.FC<propTypes> = ({ items }) => {
+  return (
+    <div className="PopulationTimeseriesLegend">
+      {items.map((label) => (
+        <div className="PopulationTimeseriesLegend__Item" key={label}>
+          <div
+            className="PopulationTimeseriesLegend__Icon"
+            id={`${label}Legend`}
```
macfarlandian, line 32: I think you would be better off using a modifier class rather than an id to target these for styling

src/core/PopulationTimeseriesChart/PopulationTimeseriesChart.scss
```diff
@@ -0,0 +1,58 @@
...
+  &__Header {
+    display: flex;
+    justify-content: space-between;
+    padding-top: 2.5rem;
+    padding-left: 2.5rem;
+    padding-right: 2.5rem;
+  }
+
+  &__Title {
+    color: #012322;
```
daschi, line 19: I added this in one of my PRs as the scss variable `$pine-1`, I was wondering if we want to add some TODOs in places where we would want to consolidate using the component/design library variables (https://github.com/Recidiviz/pulse-dashboard/issues/908)? I think it's fine either way as long as we know that issue will include updating all of these references eventually..

src/core/PopulationTimeseriesChart/index.tsx
```diff
@@ -0,0 +1,269 @@
...
+  Gender,
+  PopulationProjectionTimeseriesRecord,
+  SimulationCompartment,
+} from "../models/types";
+
+import "./PopulationTimeseriesChart.scss";
+import PopulationTimeseriesLegend from "./PopulationTimeseriesLegend";
+
+// eslint-disable-next-line @typescript-eslint/no-var-requires,global-require
+const ResponsiveXYFrame = require("semiotic/lib/ResponsiveXYFrame") as any;
```
macfarlandian, line 30: semiotic v1.20.6 should come with type definitions, you shouldn't have to do this anymore (warning though, they are a bit spotty)

daschi, line 30: Why does this need to be required like this?

src/core/PopulationTimeseriesChart/index.tsx
```diff
@@ -0,0 +1,269 @@
...
+};
+
+const getDate = (d: PopulationProjectionTimeseriesRecord): Date =>
+  new Date(d.year, d.month - 1);
+
+const prepareData = (
+  data: PopulationProjectionTimeseriesRecord[]
+): PreparedData => {
+  const historicalPopulation = data
+    .filter((d) => d.simulationTag === "HISTORICAL")
```
daschi, line 91: Just curious if there's a reason not to use the `recordMatchesSimulationTag` helper function here

src/core/PopulationTimeseriesChart/index.tsx
```diff
@@ -0,0 +1,269 @@
...
+  months: MonthOptions;
+  data: PopulationProjectionTimeseriesRecord[];
+};
+
+const CURRENT_YEAR = 2021;
+const CURRENT_MONTH = 1;
+
+const SELECTED_GENDER = "MALE";
+const SELECTED_COMPARTMENT = "SUPERVISION";
+const SELECTED_STATUS = "PAROLE";
```
macfarlandian, line 44: i'm assuming these are .... placeholders?

src/core/PopulationTimeseriesChart/index.tsx
```diff
@@ -0,0 +1,269 @@
...
+
+  return { beginDate, endDate };
+};
+
+const PopulationTimeseriesChart: React.FC<PropTypes> = ({ months, data }) => {
+  const filteredData = filterData(
+    months,
+    SELECTED_GENDER,
+    SELECTED_COMPARTMENT,
+    SELECTED_STATUS,
```
daschi, line 158: Once the filters PR is in these should come from the filters store right?

src/core/PopulationTimeseriesChart/index.tsx
```diff
@@ -0,0 +1,269 @@
...
+
+type MonthOptions = 1 | 6 | 12 | 24 | 60;
+
+type PropTypes = {
+  months: MonthOptions;
+  data: PopulationProjectionTimeseriesRecord[];
+};
+
+const CURRENT_YEAR = 2021;
+const CURRENT_MONTH = 1;
```
daschi, line 40: Is the reason we're setting the current month/year because of the status of the fixture data?

src/core/models/PopulationProjectionTimeseriesMetric.ts
```diff
@@ -0,0 +1,41 @@
...
+export function populationProjectionTimeseries(
+  rawRecords: RawMetricData
+): PopulationProjectionTimeseriesRecord[] {
+  return rawRecords.map((record) => {
+    return {
+      year: Number(record.year),
+      month: Number(record.month),
+      compartment: record.compartment as SimulationCompartment,
+      legalStatus: record.legal_status,
+      gender: record.simulation_group as Gender,
```
daschi, line 34: Should we also add these `Gender` and and `SimulationCompartment` type assertions to the the summary metrics?

daschi (approved):
Just a few questions and comments, the chart looks beautiful 💮

src/core/PopulationTimeseriesChart/index.tsx
Catacola: It saves a large number of //@ts-ignore calls. I copied the idea form https://github.com/Recidiviz/covid19-dashboard/blob/master/src/rt-timeseries/RtTimeseries.tsx. The problem is that semiotic is highly dynamic. You can set custom keys and accessors for the data you give it. This means the signature of the data is dynamically controlled by other arguments to the charts and this just leads to TS hell if you try to match signatures.

macfarlandian (commented):
tooltip placement can be a bit of a challenge in this library, you may find [this feature from Spotlight](https://github.com/Recidiviz/public-dashboard/blob/969a7df26da629b8eba39b257465e566a8dbf3a0/spotlight-client/src/charts/XHoverController.tsx#L150) of interest

src/core/PopulationTimeseriesChart/index.tsx
Catacola: I don't believe there's a reason not to; I just didn't think of it. This while function is something I will add to helper classes.

src/core/PopulationTimeseriesChart/index.tsx
Catacola: Correct. I was just hardcoding them for now for testing and development.

src/core/PopulationTimeseriesChart/index.tsx
macfarlandian: we did this in the COVID site because there just weren't any type definitions available at that point. However your point is well taken ... the types that now exist do not account for the ability to use a time scale instead of a linear scale, so e.g. in [this timeseries component](https://github.com/Recidiviz/public-dashboard/blob/master/spotlight-client/src/charts/WindowedTimeSeries.tsx#L0-L1) I did have to suppress half a dozen spurious TS errors. it's pretty annoying! and IMO not clearly better than just doing this (though I wouldn't make it the default practice for using Semiotic, for most use cases the types are adequate, though they do struggle to capture the dynamism of the API)

src/core/PopulationSummaryMetrics/index.tsx
Catacola: Whoops, I've been trying to do that but I must have just missed this one. Thanks!

src/core/PopulationTimeseriesChart/PopulationTimeseriesLegend.tsx
Catacola: What's the advantage of class over ID? I do default to class names for most styling, but when something is very specific and should only have one instance (e.g. a legend entry), I give IDs a shot.

src/core/PopulationTimeseriesChart/index.tsx
Catacola: Yup! They exist for testing. Daniela is working on a filtering diff now and I'll hook these values in to those filters once it's in.

src/core/PopulationTimeseriesChart/PopulationTimeseriesChart.scss
Catacola: Good call. Since we already have the component library, I can start cleaning these up now, but I'll keep an eye out for more places for the future.

src/core/PopulationTimeseriesChart/PopulationTimeseriesLegend.tsx
macfarlandian: I think the main one would be consistency/readability ... when you see classes attached to something you know they are styling hooks, whereas when you see an ID attached to something it may not be obvious why it's there (e.g. it took me several minutes of combing through related files to find the references to it)

src/core/PopulationTimeseriesChart/index.tsx
Catacola: Yup! In the future we will get this information from the metadata of the simulation view when we fetch it.

src/core/models/PopulationProjectionTimeseriesMetric.ts
Catacola: Sure!

macfarlandian (approved):
📉

PR #948, opened 2021-03-23
Little bits of population projection chart polish

Catacola:
## Description of the change

This diff contains small design fixes before I dedicate a big one to the tooltip

- Align axis labels with design
- Use selections from filter bar
- Style points on line
- Select supervision or incarceration view based on path
- Fix chart alignment with the rest of elements on the page

Screen Shot 2021-03-23 at 5 54 24 PM
Note: Fake data

## Type of change

## Related issues

More of https://github.com/Recidiviz/recidiviz-data/issues/6097

### Development

### Code review

### DIFF
--- a/src/core/PopulationTimeseriesChart/index.tsx
+++ b/src/core/PopulationTimeseriesChart/index.tsx
@@ -230,11 +230,14 @@ const PopulationTimeseriesChart: React.FC<PropTypes> = ({ data }) => {
     data: projectedPopulation,
   };
 
+  const populationType =
+    compartment === "SUPERVISION" ? "Supervised" : "Incarcerated";
+
   return (
     <div className="PopulationTimeseriesChart">
       <div className="PopulationTimeseriesChart__Header">
         <div className="PopulationTimeseriesChart__Title">
-          Total Supervised Population
+          Total {populationType} Population
         </div>
         <PopulationTimeseriesLegend items={["Actual", "Projected"]} />
       </div>
--- a/src/core/PopulationTimeseriesChart/index.tsx
+++ b/src/core/PopulationTimeseriesChart/index.tsx
@@ -16,7 +16,10 @@
 // =============================================================================
 
 import React from "react";
+import { useLocation } from "react-router-dom";
 import { scaleTime } from "d3-scale";
+import { observer } from "mobx-react-lite";
+import { usePopulationFiltersStore } from "../../components/StoreProvider";
 import {
   Gender,
   PopulationProjectionTimeseriesRecord,
@@ -25,6 +28,7 @@ import {
 
 import "./PopulationTimeseriesChart.scss";
 import PopulationTimeseriesLegend from "./PopulationTimeseriesLegend";
+import { CORE_VIEWS, getViewFromPathname } from "../views";
 
 // eslint-disable-next-line @typescript-eslint/no-var-requires,global-require
 const ResponsiveXYFrame = require("semiotic/lib/ResponsiveXYFrame") as any;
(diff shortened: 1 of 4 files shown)


src/core/PopulationTimeseriesChart/index.tsx
```diff
@@ -25,23 +28,21 @@ import {
...
 type PropTypes = {
-  months: MonthOptions;
   data: PopulationProjectionTimeseriesRecord[];
 };
 
 const CURRENT_YEAR = 2021;
 const CURRENT_MONTH = 1;
 
-const SELECTED_GENDER = "MALE";
 const SELECTED_COMPARTMENT = "SUPERVISION";
```
daschi, line 45: It looks like you could also delete SELECTED_COMPARTMENT here

src/core/PopulationTimeseriesChart/index.tsx
```diff
@@ -50,15 +51,17 @@ const filterData = (
   status: string,
   data: PopulationProjectionTimeseriesRecord[]
 ): PopulationProjectionTimeseriesRecord[] => {
-  const stepSize = monthRange / 6;
+  const range = monthRange === 1 ? 6 : monthRange;
```
daschi, line 54: Is this just because of the lack of data for 1 month? If so - should we add a TODO here to update when the data is updated?

src/core/PopulationTimeseriesChart/index.tsx
```diff
@@ -248,12 +284,13 @@ const PopulationTimeseriesChart: React.FC<PropTypes> = ({ months, data }) => {
...
         yExtent={[0, chartTop]}
         showLinePoints
+        pointClass="PopulationTimeseriesChart__Point"
         axes={[
           { orient: "left", tickFormat: (n: number) => n.toLocaleString() },
           {
             orient: "bottom",
             tickValues: historicalPopulation
-              .concat(projectedPopulation)
+              .concat(projectedPopulation.slice(1)) // don't double-draw center date
```
daschi, line 293: When I'm looking at 1/6 month time period, the center date moves forward 1 month, is that intended?

daschi (commented):
Just some questions about the data, I also noticed when the filters are on 6months/Male the `PROJECTED` annotation is covered by the data, I think there may be a way to keep those from colliding?

Screen Shot 2021-03-24 at 1 41 34 PM

src/core/PopulationTimeseriesChart/index.tsx
daschi: Oh I see you took this out in the next PR! resolving..

src/core/PopulationTimeseriesChart/index.tsx
Catacola: This is because we show the 6-month projection even when 1 month is selected by design. In a future iteration, we will add an auto tool-tip for the current and 1-month data points or remove the 1 month option altogether.

src/core/PopulationTimeseriesChart/index.tsx
Catacola: This is a data issue. Despite the data being from a january simulation, the February 2021 data is marked as historical and the ranges are based upon which datapoints are marked as historical or not. The data should be addressed by launch, but the chart is working as intended.

Catacola:
> Just some questions about the data, I also noticed when the filters are on 6months/Male the `PROJECTED` annotation is covered by the data, I think there may be a way to keep those from colliding?
> 
> Screen Shot 2021-03-24 at 1 41 34 PM

Thanks for flagging. I've flagged an issue related to this to @soniachokshi already. We'll find a solution.

PR #959, opened 2021-03-26
Modify API for population projections to accept new data

Catacola:
## Description of the change

I've added the real data to our export buckets and the shape of the data has changed slightly. This diff makes a few minor modifications so the FE can process the new shape.

## Type of change

## Related issues

> Closes [#XXXX]

### Development

### Code review

### DIFF
--- a/src/core/PopulationTimeseriesChart/helpers.ts
+++ b/src/core/PopulationTimeseriesChart/helpers.ts
@@ -77,7 +77,7 @@ export const prepareData = (
 
   const projectedPopulation = historicalPopulation.slice(-1).concat(
     data
-      .filter((d) => d.simulationTag !== "HISTORICAL")
+      .filter((d) => d.simulationTag === "BASELINE")
       .map((d) => ({
         date: getDate(d),
         value: d.totalPopulation,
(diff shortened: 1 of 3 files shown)


daschi (approved):
The code looks good but I think you need to rebase on master before merging! This looks like it was branched from `feature/vitals`

Catacola:
> The code looks good but I think you need to rebase on master before merging! This looks like it was branched from `feature/vitals`

Thanks! Just fixed that.

src/core/PopulationTimeseriesChart/helpers.ts
```diff
@@ -77,7 +77,7 @@ export const prepareData = (
 
   const projectedPopulation = historicalPopulation.slice(-1).concat(
     data
-      .filter((d) => d.simulationTag !== "HISTORICAL")
```
daschi, line 80: Do we also need to update HISTORICAL on lines 92, 95?

Also, do we need to update `POLICY_A` to be `PROJECTED`?

daschi: And also update the values in the demo data files

Catacola: My next PR will be to update the summary cards. Sadly, we need to shelve the admission and release projections until we can validate their output. I'll get that change in there.

PR #965, opened 2021-03-29
Add rolling average trend line to vitals chart

### LINKED ISSUE #950: Add the rolling 7 day average line to the VitalsSummaryChart
The values for the rolling 7day average line is the column `7d_avg` on the timeseries data:
```
// proposed data structure
{ "summary" | "ftr_enrolement" | ... : {
  date: string; // YYYY-MM-DD
  value: number;
  7d_avg: number; // needs to be separate.
  // Cannot calculate the earliest 7 days of averages from this data alone
  parent_7d_avg?: number; // average for the parent entity for this metric
}}[]
```

When the user selects a different metric this line should also update to the metric's rolling average value.

N/A

**Additional context**
[Data structure proposal](https://paper.dropbox.com/doc/Data-shape-for-vitals--BHd3ufKZFwyOToFScYvBVqpSAg-j4g9yhvl0VWN1g60vwsqF)
Mocks: https://www.figma.com/file/H08AQJSaAIRxfeMP4iCUPh/UP-MVP?node-id=2022%3A4

daschi:
## Description of the change

This PR adds a rolling average trendline to the vitals chart based on the values from the `weeklyAvg` property. 

On chart hover, the bar should highlight and the line should display a point! Thanks to @Catacola and @macfarlandian to helping me make that possible 🙏 

## Type of change

## Related issues

> Closes #950 

### Development

### Code review


src/core/VitalsSummaryChart/index.tsx
```diff
@@ -32,27 +33,76 @@ interface PropTypes {
...
+        }}
+        baseMarkProps={{ transitionDuration: { default: 500 } }}
+        svgAnnotationRules={(annotation: any) => {
+          if (annotation.d.type === "frame-hover") {
+            const { d, adjustedSize, orFrameState } = annotation;
+            const column = orFrameState.projectedColumns[d.date];
+            const cx = column.middle;
+            const cy = adjustedSize[1] - d.weeklyAvg * 2;
+            setHoveredId(d.index);
+            return <circle cx={cx} cy={cy} r={4} fill={styles.indigo} />;
```
jovergaag, line 85: I noticed that the circle is on the right hand side of the bar instead of the middle (as I would have expected from line 82). 
Screen Shot 2021-03-29 at 12 24 50 PM

jovergaag: Also, sometimes the tooltip hides the datapoint.

Screen Shot 2021-03-29 at 12 26 29 PM

jovergaag (changes requested):
Cool! Nice job. Just a couple of UI polish things.

src/core/VitalsSummaryChart/index.tsx
daschi: I moved the line and the points slightly with -5px, but I'm not sure how to control the tooltip positioning...I tried using the `annotationSettings` for layout but it didn't seem to apply.

src/core/VitalsSummaryChart/index.tsx
```diff
@@ -32,27 +33,73 @@ interface PropTypes {
...
-              width: BAR_WIDTH,
-              stroke: "none",
+            type: "ordinal-line",
+            coordinates: lineCoordinates,
+            lineStyle: {
+              stroke: styles.indigo,
+              strokeWidth: 2,
             },
+            curve: curveCatmullRom,
+            interactive: true,
```
macfarlandian, line 67: i wonder if you'd be better off leaving this false, it makes for slightly weird hover behavior? the tooltip just kind of moves a little if your mouse moves too close to the point

src/core/VitalsSummaryChart/index.tsx
```diff
@@ -32,27 +33,73 @@ interface PropTypes {
...
+            // Shift the point slightly to the left to center it
+            const cx = column.middle - 5;
+            const cy = adjustedSize[1] - d.weeklyAvg * 2;
+            setHoveredId(d.index);
+            return <circle cx={cx} cy={cy} r={4} fill={styles.indigo} />;
+          }
+          setHoveredId(null);
+          return null;
+        }}
+        pieceHoverAnnotation
```
macfarlandian, line 91: similarly I wonder if the UX would be a little smoother if instead of this prop you just passed a `column-hover` annotation instead. if theres a big gap between the line and the bar you wind up with kind of a "dead zone" where the chart doesn't respond to hover

src/core/VitalsSummaryChart/index.tsx
macfarlandian: I haven't yet figured out a good way to generalize this but there is [some logic in Spotlight](https://github.com/Recidiviz/public-dashboard/blob/master/spotlight-client/src/charts/XHoverController.tsx#L150) to place the tooltip off to the left or right depending on which side has more room ... i'm not sure if it will work as-is in an `OrdinalFrame`, haven't tested it, but might be something to go on here

Catacola (commented):
Looks very nice!

I saw that the dots on hover are a little bit off-center compared to the bars. It looks like the left edge of the circle is centered instead of the center. Maybe adding `transform: translateX(-50%)` to the dots might fix this.
Screen Shot 2021-03-29 at 5 17 13 PM

src/core/VitalsSummaryChart/index.tsx
daschi: great suggestion, I think the hover behavior is much better now with column hover

daschi:
> I saw that the dots on hover are a little bit off-center compared to the bars. It looks like the left edge of the circle is centered instead of the center. Maybe adding transform: translateX(-50%) to the dots might fix this.

@Catacola I tried to align the points better, but am finding it really challenging since the annotation `ordinal-line` doesn't provide coordinates that align with the OR frame coordinates. What I came up with is an estimation of the position, based on the column width and the "middle" position, which I think the value is actually the middle of the column width * 2 for what I think is the voronoi hover area. This works OK with the line transformed left by `-0.9%`, but since the value of `cx` on the point is dependent on a dynamic column width, it starts to stray at much bigger or smaller chart widths.

I considered adding a `pixelColumnWidth`, but that makes the chart unresponsive to chart width changes. 

What I think we need to do is set the `transform` value on the `ordinal-line` `path` svg element equal to the calculation in the `svgAnnotationRules` for `cx`, then I think it would align perfectly and be responsive to changing the chart width. However I didn't find a way yet to make that work..

Let me know what you think of it as it is now and if you want to chat / have any ideas!!

src/core/VitalsSummaryChart/index.tsx
```diff
@@ -76,20 +76,30 @@ const VitalsSummaryChart: React.FC<PropTypes> = ({ data }) => {
...
-            const column = orFrameState.projectedColumns[d.date];
-            // Shift the point slightly to the left to center it
-            const cx = column.middle - 5;
-            const cy = adjustedSize[1] - d.weeklyAvg * 2;
-            setHoveredId(d.index);
+          if (annotation.d.type === "column-hover") {
+            const { d, adjustedSize } = annotation;
+            const { pieces, column } = d;
+            const { data: pieceData } = pieces[0];
+            // // Shift the point slightly to the left to center it
```
Catacola, line 83: Nit: double comment

Catacola (approved):
This look so much better! Thank you! I think this is good enough to at least land, if not launch. I say we let Sonia look at it during review and if changes need to be made, we'll do it then.

macfarlandian:
@daschi I have been poking at this some more, I think it's the fixed `BAR_WIDTH` setting that is wreaking havoc with your positioning here. Semiotic naturally wants to size every bar as a percentage of its column width and place it at the center ... overriding that for the bars forces the points and line out of alignment. If you were to do away with that then `column.middle` would actually mean what it says and stuff should start fitting into place without having to fudge the positioning

src/core/VitalsSummaryChart/VitalsSummaryChart.scss
```diff
@@ -39,13 +39,14 @@
...
     path {
-      transform: translate(-5px, 0);
+      transform: translate(-0.9%, -1px);
     }
   }
 }
 
 .VitalsSummaryTooltip {
   @include chart-tooltip;
+  transform: translate(0, 1rem);
```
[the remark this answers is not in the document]
daschi: Oh this is leftover, I probably don't need it anymore!

PR #974, opened 2021-03-31
Vitals summary breadcrumbs

### LINKED ISSUE #870: Breadcrumb navigation for vitals page hierarchy
PRD section: https://docs.google.com/document/d/1moftAcmQ2JWddxwI99btG0DDSC8fBIChwfysb6Hvam4/edit#heading=h.jde0juu0lele
Mock: https://www.figma.com/file/H08AQJSaAIRxfeMP4iCUPh/UP-MVP?node-id=2022%3A4

There will be three levels of hierarchy for ND community vitals. There will be at least 3 in all subsequent states, often more.

- North Dakota statewide view
- Office view (17 district offices)
- Officer view (~150 officers)

Each level of hierarchy will have a very similar page layout, with the only difference being the caseload table. There should be x-level hierarchy to accommodate each of the levels.

From an office or officer view, users can click on the breadcrumbs to return to a previous page (clicking Mandan in the below mock will take a user back to the Mandan view):

This task involves creating the breadcrumb navigation through each page of hierarchy.

**Additional context**

jovergaag:
## Description of the change

Allows the user a path to find their way back to the state level vitals summary page, from the officer view. To do this, we needed to add `entity_type` onto the `VitalsSummaryRecord` model (new fixture files have been added to the bucket so this should run in dev mode)

## Type of change

## Related issues

> Closes #870

### Development

### Code review


src/core/VitalsSummaryBreadcrumbs/index.tsx
```diff
@@ -0,0 +1,54 @@
...
+  entity: VitalsSummaryRecord;
+};
+
+const VitalsSummaryBreadcrumbs: React.FC<PropTypes> = ({
+  stateName,
+  entity,
+}) => {
+  const { entityName, entityType } = entity;
+  const { primary, secondary } =
+    entityType === ENTITY_TYPES.DISTRICT_LEVEL
```
daschi, line 38: I wonder if we will eventually want to have something closer to lantern, like numbered levels (level_1_entity, etc), particularly if other states will have more than these 3 entity types. I think this works well for now, just wanted to bring that question up.

daschi (changes requested):
This is great! It's hard to see if the time series chart is working though because it looks like all of the entities in the demo data have the same values. I wonder if we should randomize those a bit so we can see the chart updating when we change entities. Not a big deal though and we can wait to QA with actual data too!

src/core/VitalsSummaryBreadcrumbs/index.tsx
jovergaag: Hmm, good question. Let me think on that a little. We definitely will have the third level (officer), so you're right that maybe we should lay the groundwork for that here.

jovergaag: I think that sticking with the `primary` and `secondary` naming might be more clear because the different entities will appear in the different breadcrumb divs depending on what entity is currently selected. So sometimes state name is in `primary` and other times office is in `primary`. If we try to do something with `level_1` and `level_2` in the breadcrumbs, it might get confusing since that is how the office/sub-office are categorized. So, for example, we might be trying to insert a `level_1` office id into a `level_2` breadcrumb div which is confusing. Does that make sense?

Once we enable more levels, we can just enable the `tertiary` breadcrumb, etc.

Thoughts?

jovergaag: Or did you mean changing to levels for ENTITY_TYPES?

daschi: Oh yeah I meant changing the data that comes through in entity_types, so we're not needing to come up with entity type names that are specific to each state. So instead of "STATE_DOC" it would be like level_1 and "DISTRICT_LEVEL" would be level_2, etc.

jovergaag: OK, so Carole and I talked about it, and we left STATE, and PO because those are so specific and will have very specific roles in the page, but changed district to LEVEL_1_SUPERVISON_LOCATION to be consistent with Lantern. LEVEL_1_SUPERVISION_LOCATION will always be the most fine grain district level, so if there is a sub-office the sub-office will be level_1 and the district will be level_2.

jovergaag: @daschi :)

src/core/VitalsSummaryTable/VitalsSummaryTable.tsx
```diff
@@ -21,16 +21,25 @@ import cx from "classnames";
...
 import "./VitalsSummaryTable.scss";
 
+function getEntityTypeName(entityType: EntityType): string {
+  switch (entityType) {
+    case ENTITY_TYPES.LEVEL_1_SUPERVISION_LOCATION:
+      return "Office";
+    case ENTITY_TYPES.PO:
+      return "Officer";
+    case ENTITY_TYPES.STATE:
+      return "STATE_DOC";
```
daschi, line 37: It should never get STATE here right, since it will never be a child entity?

daschi (approved):
I like the new entity type naming system, I think it'll be helpful when we have more levels! I just had one question about the entity type case for the table, but this looks good 🥞

src/core/VitalsSummaryTable/VitalsSummaryTable.tsx
jovergaag: yes you're right. I'll remove it

PR #994, opened 2021-04-06
[exploration] E2E tests with WebDriverIO + Cucumber

daschi:
***This PR is not meant to be merged!***

@Catacola Thanks for offering to test out some of the implementation this week! Here's the first one I've set up with an actual test, and would be interested in hearing your thoughts on the WebDriverIO part and also the Cucumber part, since we can also switch out which browser we're using with something like [Playwright](https://playwright.dev/docs/intro) or [Modern Web](https://modern-web.dev/guides/test-runner/getting-started/). 

@ everyone Please let me know if you have any feedback or thoughts about this approach! Tomorrow I am planning to try the BrowserStack integration that will let us run tests against IE11. 

### Run the tests

1. Add these vars to `.env.development`
*note* I created a "test" user for myself for now to get this up in running but I imagine we will want to have a separate email address for different testers?
`TEST_AUTH_USER=[your auth0 email for now]`
`TEST_AUTH_PASSWORD=[your auth0 pw for now]`
`RUN_TESTS_HEADLESS=true `

2. Start the development server `yarn demo` (long term I'm thinking this would be part of the test script and use a different port)
3. Use the command `yarn test-features`

To run in BrowserStack (we have ~290 free minutes to test):
1. Add the env vars:
`BROWSERSTACK_USER=[find this when you login under automate]`
`BROWSERSTACK_KEY=[find this when you login under automate]`
`RUN_BROWSERSTACK=true`

### [WebDriverIO](https://webdriver.io/docs/gettingstarted)

Here's some of the pros & cons I've seen so far with this implementation:

__WebDriverIO Pros__
- You can configure the browser driver to be Selenium based or Chromium based, and it also supports integrating remote browser services like BrowserStack, so we could run our tests against an IE11 remote browser.
- So far I think I like organizing the selectors and page methods within the [`PageObject`](https://webdriver.io/docs/pageobjects/) approach
- Logging into Auth0 was pretty easy & straightforward (though I think it would be just as easy with Playwright, etc.)
- You can access the `puppeteer` browser instance and use the puppeteer API 
- You can select react components by name and props

__WebDriverIO Cons__ (maybe?)
- It adds a lot of dependencies! (8 new dependencies)
- It doesn't come with built in waiting, we need to use `browser.pause()` to wait for the page or elements to load
- I'm sure there are more that I haven't thought of yet because my example is really simple...

#### Cucumber Step Definitions
I don't think we should organize the steps long term in files called `given`, `when`, `then`, but I found it useful to get started here. I think the best practice could be to organize step definitions by domain concepts. I'm not sure yet, but I think as the feature tests grow we can see what makes the most sense to help these from growing out of control and reduce duplication. 

I'm also thinking about setting up another example branch using Playwright so we can compare the two. 

TODOs

- [x] Set up the remote browserstack integration
- [ ] Set up Travis CI to run the test, both with chrome + browserstack
- [ ] Use Typescript

More information about other tools and questions: https://paper.dropbox.com/doc/E2E-Testing-for-pulse--BIXDNLNVU~UAmQ5nbux67qFiAg-BHO2F7AedwfdPgGAeKCB3


macfarlandian:
this is pretty cool ... the Cucumber syntax is easy to follow, it seems like a pretty decent choice here. Pretty simple to get running locally also, although I have yet to be able to get the browserstack test to pass. I suspect that may be down to the lack of a "wait for something" feature, because browserstack is sloooooow. That's the big concern I have coming out of this one, curious to see your comparison branch!

Catacola (commented):
I'm liking this so far. I was afraid cucumber's DSL was going to be cumbersome, but it's much more accessible than I feared.

jovergaag:
Cool! The cucumber tests are so easy to read! 🤩 Similar experience as @macfarlandian, the browserstack tests failed both times I ran them. But if we were to find a way to get this to work it would be awesome, I love that you can go in and replay a session in browserstack to see the failure. Super helpful if we can get it to run consistently!

What will it take to get these running on Travis?

daschi:
@jovergaag / @macfarlandian Interesting about the BrowserStack tests, they have only passed for me...if one of you has time could you show me your dashboard? 

> What will it take to get these running on Travis?

If we want to[ run BrowserStack in Travis ](https://webdriver.io/docs/cloudservices#with-travis-ci-1)I think we just need to download & start `BrowserStackLocal`. I was going to try to add that today too to see how it works.

src/cucumber/pages/index.js
```diff
@@ -0,0 +1,8 @@
+// https://webdriver.io/docs/pageobjects/
+/* eslint-disable class-methods-use-this */
+export default class Page {
+  open(path) {
+    browser.url(path);
+    browser.pause(2000);
```
jessex, line 6: Is it possible to take args into `Page` that let you specify how long you want to pause (or other such features)? I could imagine we'd want to use `Page(pause=10000)` for Lantern pages, for example.

src/cucumber/steps/when.js
```diff
@@ -0,0 +1,8 @@
+import { When } from "@cucumber/cucumber";
+import loginPage from "../pages/loginPage";
+
+When("I login as an {string} user", function (userLevel) {
+  const { username, password } = browser.config[userLevel];
+  loginPage.login(username, password);
+  browser.pause(2000);
```
jessex, line 7: Why do we need this pause in particular? Is it because `login` is asynchronous, it triggers some operation and returns immediately, so we're waiting on that operation to finish? I wonder if there's a better route, like the equivalent of Promises?

src/cucumber/steps/when.js
```diff
@@ -0,0 +1,8 @@
+import { When } from "@cucumber/cucumber";
+import loginPage from "../pages/loginPage";
+
+When("I login as an {string} user", function (userLevel) {
+  const { username, password } = browser.config[userLevel];
```
jessex: Can you explain what this `browser.config` _is_? What state is this and why does `userLevel` return credentials? This line looks magical to me at first glance.

jessex: Oh, I see, it's `wdio.conf.js` below. I think that @Catacola's recommendation below would make this more readable to me!

wdio.conf.js
```diff
@@ -0,0 +1,323 @@
+const config = {
```
jessex, line 1: How much of this file is custom vs copy+paste? It looks like a mixture of things going on here, but it may just be my unfamiliarity with Webdriver.

jessex (commented):
Interesting stuff, @daschi! I left some questions/comments. From my vantage point, I'm curious to know more of your thoughts about how this hooks into the larger dev lifecycle, CI, etc. Also excited to see what the browser-specific story looks like, including IE11.

>What will it take to get these running on Travis?

I wonder if we should take the opportunity to go ahead and migrate to Github Actions instead?

src/cucumber/pages/index.js
daschi: Yeah I think that would make sense, particularly since we're already seeing the IE11 lantern pages vary in load time on BrowserStack.

src/cucumber/steps/when.js
daschi: I added this pause because of the time it took to redirect from the login page to the main lantern page. [WebdriverIO lets you configure whether you want everything to run sync or async mode](https://webdriver.io/docs/sync-vs-async), and at first I tried it as `async`, which was fine but I thought changing it to `sync` would be easier to read. Even with `async` mode, I still think we would need to sprinkle in `browser.pause()` to wait for page loads. 

Here's an example they have on their site to implement this with async, and what made me think it would be less straightforward:
```
describe('suite async', () => {
    it('test', async () => {
        const el = await $('body') // note `await` keyword
        await el.click()

        await browser.pause(500)

        // this WON'T WORK! You can't chain functions like this.
        await $('body').click()
        // instead you need to:
        await (await $('body')).click()
    })
})
```

wdio.conf.js
daschi: Most of it is generated by a CLI tool from the package, the parts I changed were to add `@babel/register` as a required module, adding a `baseUrl`, the `chrome` and `ie11` browser capabilities, and the `credentials` config. I added an environment variable to set running on BrowserStack on/off because we have limited testing minutes and I didn't want to run it every time I ran the tests while setting everything up, but I could see that moving into the config object if we decide to run IE11 on every build.

daschi:
> From my vantage point, I'm curious to know more of your thoughts about how this hooks into the larger dev lifecycle, CI, etc.

I was thinking for the larger dev cycle, we would want to start adding tests for new features and pages as they become more stable and are launched to production.  I chatted with Jen and Tony a bit about the CI process, and we talked about using demo data and running the tests on every build to start. We could keep an eye on how long the builds take and decide whether we need a different cadence as we add more tests. Happy to talk more in depth about it this afternoon! 

> I wonder if we should take the opportunity to go ahead and migrate to Github Actions instead?

I spent most of yesterday looking into how to hook this into the CI process, and I'm going to continue looking into that today. I could look at transitioning to Github Actions to see what that would take, since other products are also moving in that direction.

daschi:
I chatted with @jessex and @Catacola today, and I think for now we want to reduce some of the complexity with this project by removing the requirements of having it run in CI and on BrowserStack for IE11. In the short to medium term we can continue to develop these E2E tests and get them running in a stable environment using `puppeteer` (testing on chrome) via `WebDriverIO`, and later on when we feel more confident with them we can try to integrate it with CI and then eventually explore running it against a BrowserStack IE11 instance. 

There's a couple reasons I think to go this route, integrating with CI may be more complicated and take more time than this one week, and also, we need to reach out to BrowserStack to confirm the pricing for using the Automate product for remote testing.

Let me know if anyone wants to chat about this more or has anything to add!! I'm planning to close this PR and open a separate one that uses just chromium to run the tests. I also plan to add to the test here for the demo.

daschi:
Closing in favor of https://github.com/Recidiviz/pulse-dashboard/pull/999

PR #1011, opened 2021-04-14
Update Lantern race and gender charts for US_PA

### LINKED ISSUE #963: Show all race and genders on the same chart, to match ND's designs
During the PA launch meeting, they provided feedback that the race and gender charts would be more user-friendly if they could see all races and genders on the same chart, similar to how it appears in the revocations by race rate chart on the ND dashboard.

I flagged that the smaller race categories might get hard to distinguish with that design, and less easy to compare across stages of the funnel, which is why we stayed away from it. However, their primary use case is comparing across races for a particular part of the funnel (i.e., supervision population), so that is less of an issue.

This task involves 
- updating the race and gender chart UI to show all races/genders on the same chart, rather than having toggle pills between the races
In addition.
- Changing the race groupings in PA to be "White," "Black," "Hispanic," and "Other," which are the categories most relevant for them.

cc @soniachokshi and @terinpw for mocks or any concerns.

**Additional context**
Open question for Serena: would MO be alright with this change as well.

jovergaag:
## Description of the change

US_PA has decided that they want to show multiple race categories together on the race chart simultaneously in the form of a stacked bar chart. This changes the race and gender charts for US_PA to stacked horizontal bar charts, and maintains the horizontal bar chart by mode for US_MO.

Note: Data to follow. Eventually the bars will all add up to 100%, but the BE work to sum additional race categories into the Other category is still [in progress.](https://app.zenhub.com/workspaces/torch--5ec36552cdfe7bc58f04bdbe/issues/recidiviz/recidiviz-data/6977)

## Type of change

## Related issues

> Closes #963

### Development

### Code review


src/lantern/RevocationsByGender/RevocationsByGender.js
```diff
@@ -31,9 +31,11 @@ const DEFAULT_MODE = "MALE";
 const RevocationsByGender = observer(
   ({ containerHeight, timeDescription }, ref) => {
     const { dataStore } = useRootStore();
+    const { currentTenantId } = useRootStore();
```
daschi, line 34: We already call `useRootStore` on line 33, you could pull `currentTenantId` together with `dataStore`

src/lantern/RevocationsByGender/RevocationsByGender.js
```diff
@@ -31,9 +31,11 @@ const DEFAULT_MODE = "MALE";
...
     const { revocationsChartStore } = dataStore;
     const CHART_TITLE = translate("revocationsByGenderChartTitle");
     const CHART_ID = translate("revocationsByGenderChartId");
+    const stacked = currentTenantId === "US_PA";
```
daschi, line 38: I think we could use the `US_PA` constant from `lanternTenants` here

daschi (approved):
I have a couple small suggestions, do we need to put these behind a flag until the data is ready?

src/lantern/RevocationsByGender/RevocationsByGender.js
jovergaag: doh!

PR #1015, opened 2021-04-15
Vitals Summary - Officer view

### LINKED ISSUE #991: Create officer-level views for Vitals
For the Vitals dashboard, we have both state-level and office-level views. Now, we want to add _officer-level_ views. These will be accessible by clicking on an officer in the table in the office-level Vitals view.

The officer-level view will only include the summary cards and performance over time for each metric. For the MVP, there will **not** be a caseload table in the officer-level view. (We are still determining how to add that table in a way that dovetails nicely with the Case Triage tool.)

Clicking on the individual metric cards at the top will simply update the card underneath with the summary metric and timeseries chart.

The user can access both the office-level view and the state-level views by using the breadcrumb navigation at the top of the page.

Mocks in Figma: https://www.figma.com/file/H08AQJSaAIRxfeMP4iCUPh/UP-MVP?node-id=2624%3A383

**Additional context**

jovergaag:
## Description of the change

Enables the Officer View for the Vitals Page. For now, the table will not be shown in the Officer View.

Two important things to note
- Waiting on time series data for the officers, so this will need to be run in demo mode for testing

## Type of change

## Related issues

> Closes #991

### Development

### Code review


server/core/demo_data/vitals_summaries.json
```diff
@@ -1,8 +1,8 @@
...
-{"entity_id": "STATE_DOC", "entity_name": "State DOC", "entity_type": "state", "parent_entity_id": "STATE_DOC", "overall": "79",  "timely_discharge": "79",  "timely_ftr_enrollment": "56",  "timely_contact": "100",  "timely_risk_assessment": "94", "overall_7d": "-4", "overall_28d": "-10"}
+{"entity_id": "OAKES", "entity_name": "Oakes",  "entity_type": "level_1_supervision_location", "parent_entity_id": "STATE_DOC", "parent_entity_name": "STATE DOC", "overall": "65",  "timely_discharge": "-12",  "timely_ftr_enrollment": "88",  "timely_contact": "100",  "timely_risk_assessment": "100", "overall_7d": "20", "overall_28d": "32"}
+{"entity_id": "OAKES", "entity_name": "Oakes",  "entity_type": "level_1_supervision_location", "parent_entity_id": "OAKES", "parent_entity_name": "STATE DOC", "overall": "65",  "timely_discharge": "-12",  "timely_ftr_enrollment": "88",  "timely_contact": "100",  "timely_risk_assessment": "100", "overall_7d": "20", "overall_28d": "32"}
```
daschi, line 2: I still don't quite understand why we need two summaries for the same data, where one has a `parent_entity_id` that matches its own `entity_id`?

src/core/VitalsSummaryBreadcrumbs/VitalsSummaryBreadcrumbs.scss
```diff
@@ -33,7 +33,12 @@
     }
   }
 
-  &--primary {
+  &--parent::before {
+    content: "/";
```
daschi, line 37: This gets underlined on hover, if we wanted to avoid that I think it might work better to add this as its own span element.

server/core/demo_data/vitals_summaries.json
jovergaag: We had decided to do this originally - but that changed somewhere along the line and I wasn't aware. I'll update the fixtures to match the data here.

src/core/models/types.ts
```diff
@@ -91,7 +91,8 @@ export type VitalsSummaryRecord = {
   entityId: string;
   entityName: string; // i.e. "North Dakota" or "Oakes"
   entityType: EntityType;
-  parentEntityId?: string; // not set for top-level
+  parentEntityId: string;
+  parentEntityName: string;
```
daschi, line 95: I think this `parentEntityName` can be removed from the model now? same question for the metric model

src/core/VitalsSummaryTable/VitalsSummaryTable.tsx
```diff
@@ -73,17 +73,14 @@ const VitalsSummaryTable: React.FC<PropTypes> = ({
...
-                >
-                  {value.entityName}
-                </Link>
-              ) : (
-                value.entityName
-              ),
+            }) => (
+              <Link
+                className="VitalsSummaryTable__link"
+                to={`/community/vitals/${convertIdToSlug(value.entityId)}`}
```
daschi, line 79: This currently is causing an error because there aren't any time series for the POs in the data, do we expect that to be added soon? If not we may want to put a safeguard on line 111 in the `PageVitals/index.tsx` since that's causing the page to crash.

src/core/VitalsSummaryBreadcrumbs/index.tsx
```diff
@@ -18,35 +18,69 @@
...
 import { Link } from "react-router-dom";
 import { ENTITY_TYPES, VitalsSummaryRecord } from "../models/types";
+import { toTitleCase } from "../../utils/formatStrings";
+import { convertIdToSlug } from "../../utils/navigation";
 import "./VitalsSummaryBreadcrumbs.scss";
 
-function formatOfficeName(name: string): string {
-  return `${name} Office`;
+function formatOfficeName(name: string | undefined): string | undefined {
+  return name ? `${toTitleCase(name)} Office` : name;
```
daschi, line 26: One of the entity names already has "Office" in it, maybe we could format it to remove any duplicates?

daschi (changes requested):
Looks good! I just think we should add a safeguard for the missing data so that we can deploy staging without it crashing, or maybe wait to merge until that data is available if it's coming soon?

src/core/VitalsSummaryTable/VitalsSummaryTable.tsx
jovergaag: Yes, you should be able to view it in demo mode. I just confirmed with. @Catacola that the time series data for officer view is a bit down the road so I'll just add this whole officer view behind a flag for now.

src/core/VitalsSummaryBreadcrumbs/index.tsx
jovergaag: I created a ticket to format the `entity_name` more consistently so that we don't have to account for these special cases. Formatting here for now.

Catacola (approved):
Looking good! I have a few minor comments but nothing blocking.

src/core/PageVitals/helpers.ts
```diff
@@ -90,12 +90,16 @@ export function getEntitySummaries(
...
 } {
   const currentEntitySummary = vitalsSummaries.find(
     (d) => d.entityId === currentEntityId
   ) as VitalsSummaryRecord;
   const childEntitySummaryRows = vitalsSummaries
-    .filter((d) => d.parentEntityId === currentEntityId)
+    .filter(
+      (d) =>
+        d.parentEntityId === currentEntityId && d.parentEntityId !== d.entityId
+    )
```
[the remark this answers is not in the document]
jovergaag: It is because there is a single row for the "STATE_DOC" where both `entityId` and `parentEntityId` are STATE_DOC. If we do not include this then we get that row in the table for the top level view.

src/core/PageVitals/helpers.ts
```diff
@@ -107,17 +111,21 @@ export function getEntitySummaries(
...
   timeSeries: VitalsTimeSeriesRecord[],
   selectedCardId: string,
   currentEntityId: string
-): VitalsTimeSeriesRecord[] {
-  return timeSeries.filter(
+): VitalsTimeSeriesRecord[] | undefined {
+  const selectedTimeSeries = timeSeries.filter(
     (d) => d.metric === selectedCardId && d.entityId === currentEntityId
   );
+  return selectedTimeSeries.length > 0 ? selectedTimeSeries : undefined;
```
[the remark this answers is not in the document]
jovergaag: Thats how it was originally but it makes it much easier to guard for when there are no time series that match a given entity. The guards become more complicated and happen in a few places. I'm happy to revert this back to what it was and use those more complicated guards, this just seemed more clean to me.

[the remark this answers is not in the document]
jovergaag: That last commit reverts `getTimeseries` to return an empty array rather than `undefined` if not found. I kind of like it previously but am happy to go either way! Let me know what you think.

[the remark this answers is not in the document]
Catacola: That's a very good reason! I'm happy to leave it that way if it works better for you.

daschi (approved):

PR #1027, opened 2021-04-19
Limit available state codes for out of state Recidiviz user

### LINKED ISSUE #1017: Setup ACL for recidiviz users and state requirements
Certain state data sharing contracts require users accessing their data to reside in the US. We need to modify the authorization for state dashboards that are not viewable by Recidiviz team members that reside outside the US.     

A straightforward way to assign state codes to Recidiviz users that will decide which dashboards they have access to in the app navigation. An idea we talked about was to expand on the `app_metadata` object on the user to include `available_state_codes`, and then use this list to populate the profile state selector. 

An alternative would be to maintain a list of users and their state codes on the client.  

**Additional context**
[This is a small part of the larger requirements for ACL that are found in this here. ](https://paper.dropbox.com/doc/ACL-and-SSO-considerations--BJKtFhDYRLxnrFOUkWv3ypYYAg-V5Z9PhZRNz3zuea5z0M6W)

daschi:
## Description of the change

This update changes the "Set state code in app metadata" rule hook in Auth0 to add a `blocked_state_codes` list to the `app_metadata` specifically for Justine. I don't think this is the ideal long term solution, but I was having difficulty today working through how to correctly set up and request permissions on the front end. 

I was hoping we would be able to setup permissions for a user, something like `access_state_code:us_nd`, and use the `auth0-spa-js` client to request the user's permissions to then decide what to show on the UI, but I don't think we can. We could add an [Authorization Extension](https://auth0.com/docs/extensions/authorization-extension) to make that work, but it looks like they are deprecating that service and prefer users to go with the new Authorization Core features, which doesn't include adding `permissions` to the user object. I think we can make an entirely separate and new request to auth0 for a user's permissions, which I'm going to check out tomorrow.
 
For now, opening this up to get your thoughts or to see if there are any ideas I might be missing...

TODO: 
- [ ] Apply rule change to production tenant

## Type of change

## Related issues

> Closes #1017

### Development

### Code review


src/RootStore/utils/user.js
```diff
@@ -64,7 +80,12 @@ export function getUserStateName(user) {
  */
 export function getAvailableStateCodes(user) {
   const stateCode = getUserStateCode(user);
-  return tenants[stateCode.toUpperCase()].availableStateCodes;
+  const blockedStateCodes = getUserBlockedStateCodes(user);
+  const stateCodes = tenants[stateCode.toUpperCase()].availableStateCodes;
+  return stateCodes.filter((sc) => {
+    if (!blockedStateCodes) return true;
+    return !blockedStateCodes.includes(sc);
+  });
```
Catacola, line 88: The following feels a little more direct:
```js
  const blockedStateCodes = getUserBlockedStateCodes(user);
  const stateCodes = tenants[stateCode.toUpperCase()].availableStateCodes;
  if( !blockedStateCodes) return stateCodes;
  return stateCodes.filter((sc) => !blockedStateCodes.includes(sc));
```

Catacola (approved):
Looks good to me! Just a little note about readability. Feel free to disagree and leave it as is.

PR #1051, opened 2021-04-26
Vitals download data

### LINKED ISSUE #1010: Add Download Data to Vitals dashboard
The user should be able to click on the "Download Data" button in the top right corner of the Vitals dashboard. When they do so, they'll download a CSV of the data of the view that the user is looking at, along with a .txt file with the Methodology. 

Here's an [example](https://docs.google.com/spreadsheets/d/12s5qE-OIWds_teIodXQ2lLTiIq6ICAm5atBecX4yY8c/edit#gid=111700939) of what the data download would look like if the user is looking at the state-wide view. For the office-level view, the "Metrics over time" tab would show just the performance for that office, and the third tab would change to "Metrics by officer."

The Vitals Methodology for the accompanying .txt file is [here](https://docs.google.com/document/d/1moftAcmQ2JWddxwI99btG0DDSC8fBIChwfysb6Hvam4/edit#heading=h.v6pxhexkocuw). 

**Additional context**

jovergaag:
## Description of the change

Adds the download button to the Vitals Page, and formats the summaries and time series data for download.

This feature requires multiple csvs to be downloaded in the zip along with the methodology file, so the `downloadData` logic had to be modified to support this. This means the downloads on the other pages should be tested during QA to ensure nothing broke there.

Note: I may be adding a filters description into the methodology based on design QA, but that will be very straightforward.

## Type of change

## Related issues

> Closes #1010

### Development

### Code review


src/flags.js
```diff
@@ -3,6 +3,6 @@ export default {
   enableRevocationRateByExit: false,
   showMethodologyDropdown: false,
   enableCoreTabNavigation: true,
-  enableVitalsDashboard: false,
+  enableVitalsDashboard: true,
```
daschi, line 6: do we want to merge this to main?

daschi (changes requested):
I think this makes sense, and I tested it on chrome & ie11 on both lantern & core. 

I did notice 1 issue, for the non-vitals Core downloads (i.e. Explore/community, etc.), the CSVs have lost the headers for the time period columns. They used to say `Month` and are not showing `Undefined`.

src/flags.js
jovergaag: oh crap.

daschi (approved):

Catacola (approved):
Probably outside the scope of this diff, but I think it would be a nice touch if we put some details in the name of the zip file like the data and the data downloaded. E.g. `ND_VITALS_2021_04_30.zip`.

PR #1062, opened 2021-04-29
Disable race & gender charts for US_MO restricted access users

### LINKED ISSUE #1057: Hide Race and Gender charts from district-restricted views in MO :(
MO leadership has asked for the race and gender charts to be removed from view for the district-administrator views of the MO Lantern dashboard. They are worried about how the data may be interpreted or misconstrued, even despite the recent changes that were accommodated in the design and calculations. I am following up for official documentation on the rationale and will pass it along any as I receive it.

This task involves hiding the "Race" and "Gender" middle charts from the Missouri dashboard for all users who have restricted district access.

No changes to the following
- Unrestricted views of the MO dash (leadership can still view race and gender charts)
- PA - The good news on the flip side is that PA LOVED the charts and clapped when we showed them.

The next training is the morning of 5/5, and the director of research has asked for the charts to be removed by then, so the target date for this task would be by **5/4**.

**Additional context**
Assigned Daniela because we were just chatting about it, but cc @Catacola and @jovergaag. Let me know if you have further questions or concerns!

daschi:
## Description of the change

This PR removes the race & gender charts for restricted access users in US_MO. 

## Type of change

## Related issues

> Closes #1057 

### Development

### Code review


src/lantern/LanternStore/UserRestrictedAccessStore.ts
```diff
@@ -59,6 +60,16 @@ export default class UserRestrictedAccessStore {
     });
   }
 
+  get disableRaceAndGenderCharts(): boolean {
```
jovergaag, line 63: I wonder if we should make this inclusive rather than exclusive. For example, for a given user with restricted access, what charts are available to them, rather than what charts to disable. That way its easier to add or remove charts, rather than it being this single boolean to control Race and Gender and additional logic is required for other modifications.

src/lantern/LanternStore/UserRestrictedAccessStore.ts
```diff
@@ -59,6 +60,16 @@ export default class UserRestrictedAccessStore {
...
+    return (
+      this.hasRestrictedDistricts && this.rootStore?.currentTenantId === US_MO
+    );
+  }
+
+  get hasRestrictedDistricts(): boolean {
```
jovergaag, line 69: Oh yes this makes more sense here. Nice.

src/lantern/LanternStore/DataStore/RevocationsChartStore.js
```diff
@@ -99,6 +99,15 @@ export default class RevocationsChartStore extends BaseDataStore {
     this.skippedFilters = CHARTS[chartId].skippedFilters || [];
   }
 
+  get availableChartIds() {
```
jovergaag, line 102: See note in the `UserRestrictedAccessStore`

src/lantern/LanternStore/DataStore/RevocationsChartStore.js
```diff
@@ -99,6 +99,15 @@ export default class RevocationsChartStore extends BaseDataStore {
...
+    if (this.rootStore.userRestrictedAccessStore.disableRaceAndGenderCharts) {
+      return Object.keys(CHARTS).filter(
+        (chartId) => !["Race", "Gender"].includes(chartId)
+      );
+    }
+    return Object.keys(CHARTS);
```
jovergaag, line 108: This logic is changing the order of the charts from what is in the designs. Probably not a big deal? Maybe we should sort it so it is in an explicit order.

jovergaag (changes requested):

src/lantern/LanternStore/UserRestrictedAccessStore.ts
daschi: That does sound more extensible, I can move the logic for the charts into the `UserRestrictedAccessStore` and see how that looks like.

daschi: @jovergaag Let me know what you think of the change, and if it doesn't match what you were thinking and you want to chat more about it!

jovergaag (approved):
👎 (not to you - to the disabling of charts)

PR #1094, opened 2021-05-13
Restrict user access using auth0 access and id tokens

daschi:
## Description of the change

This PR got way bigger than I was expecting!! Let me know if it would be helpful for me to talk through the changes at a high level before a PR review. 

**Server Changes**
- Removes `restrictedAccess` endpoint 
- Updates the `newRevocationFile` endpoint to use the jwt `user` field that has the auth0 parsed access token data to create the cache key and filter the metric file
- Removes `restrictedDistrict` from subsetFilters flow and adds a helper `createUserRestrictionsFilters`
- Update `getFiltersByMetricName` to return no filters for files not listed (this is so we don't filter files like supervision_location_ids_to_names`)
- Adds some logic to allow Recidiviz users to toggle between state codes without restriction access on every state code (we may want to change this later to be more fine-grained so that a recidiviz user can have different access levels for different states, but for now we only have US_MO)
- Adds a `api/demoUser` endpoint that returns the properties for a demo user. I added this for the E2E test so we can mock the response and return different user attributes.

**React App Changes**
- Adds a `fetchDemoUser` client that can only be used in demo mode to fetch the user attributes
- Updates the auth0 hooks code to reflect what is currently in `recidiviz-dev`
- Updates the E2E test to mock the demo user endpoint instead of the restricted access endpoint
- Updates the stores to use the restrictions from the user and removes calls to the restricted access endpoint
- Adds a `enableUserRestrictions` property to the `src/tenants` file that helps recidiviz users toggle between tenants while having user restrictions in their user profile. 

**.env-cmdrc changes**
- Adds two new environments: "demo" and "server"
- Adds `METADATA_NAMESPACE` to an env var since it is used in the server, react app, and cucumber tests
- Adds the new environments to the package scripts for starting the app and running tests

TODO
- [ ] Do not merge/deploy until production is deployed next week and the users in auth0 are being updated correctly 
- [ ] Update the access token hook in the Auth0 `recidiviz` production tenant

## Type of change

- [ ] Bug fix (non-breaking change that fixes an issue)
- [ ] New feature (non-breaking change that adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to not work as expected)
- [x] Configuration change (adjusts configuration to achieve some end related to functionality, development, performance, or security)

## Related issues

> Closes #1034
> Closes #1043

### Development

These boxes should be checked by the submitter prior to merging:

- [x] Manual testing against realistic data has been performed locally
[...]

PR #1099, opened 2021-05-17
Consolidate auth0 pre-registration hook and metadata rule

daschi:
## Description of the change

This updates the code we would use for the Auth0 pre-user registration hook and the rule that is setting the state code per user. I did not update the code in Auth0 yet since I wanted to open this PR and get everyone's thoughts on these changes before moving forward. 

The main change is moving setting a user's state_code in app_metadata from a rule that is run every time a user signs in, to a hook that runs when a user first signs up. Since everyone who has already signed up has a state_code in their app_metadata, we can move the logic over to this hook without causing log in issues. It also adds specific logic for ID users, since their email domain doesn't fit into the pattern of having their state code in the domain directly (they have `@idoc.idaho.gov`).

I also added a TODO in the hook for when the recidiviz-data auth endpoint is setup to respond to requests for user restrictions by email address. This will add the user restrictions in the user's app_metadata when they first sign up. Any changes to their user restrictions will then happen on a daily basis with the data pipeline exports triggering an API endpoint that will update all user restrictions based on the latest restrictions file. 

TODO:
- [ ] Once approved, update the rule / hook on the recidiviz-dev and recidiviz Auth0 tenants

## Type of change

## Related issues

> Related to https://github.com/Recidiviz/recidiviz-data/issues/7298

### Development

### Code review


src/auth0/hooks/01-pre-registration-setup.js
```diff
@@ -0,0 +1,78 @@
...
+   * This hook will do three things:
+   * 1. Prevent users from signing up that are not in the allow list
+   * 2. Add the user's state_code to the app_metadata
+   * 3. Fetch any user restrictions and add them to the user's app_metadata
+   *
+   */
+  const response = {};
+
+  /** 1. Domain allow list for registration */
+  const authorizedDomains = []; // add authorized domains here
```
Catacola, line 36: Is this empty because we're not allowing new sign-ups right now?

Catacola (commented):
I think this is a great idea. I just have on little question.

src/auth0/hooks/01-pre-registration-setup.js
daschi: I think @jessex has been updating this directly in Auth0 when we want to give specific domains access. Currently in production this list isn't empty, but I wasn't sure if we wanted to check the domains into source code since they can change depending on when sign ups are happening.

jessex: That's correct, @daschi!

src/auth0/rules/02-update-app-metadata.js
```diff
@@ -15,41 +15,38 @@
...
-    }
   }
 
   // Specific state code restrictions for Recividiz users
   if (user.email === "justine@recidiviz.org") {
     user.app_metadata.blocked_state_codes = ["us_pa"];
   }
 
+  // Normalize existing Idaho state codes
+  if (domain === "idoc.idaho.gov") {
```
jovergaag, line 46: I thought that the state code was being set in the pre-registration hook. Why are we setting it here as well (and on line 37)?

src/auth0/hooks/01-pre-registration-setup.js
```diff
@@ -0,0 +1,78 @@
...
+    // e.g. @doc.mo.gov or @nd.gov, but not @nd.docr.gov
+    let state = domainSplit[domainSplit.length - 2].toLowerCase();
+
+    // Idaho does not use the abbreviation in their email addresses
+    if (state === "idaho") {
+      state = "id";
+    }
+
+    const stateCode = `us_${state}`;
+    if (acceptedStateCodes.includes(state)) {
```
jovergaag, line 66: So if the state is not an accepted state code, the user is returned without a state_code set?

daschi: Yeah, which would trigger a `No state code set for user` auth error  on the frontend. I think this would be very rare, it would mean someone with the correct email domain was able to sign up, but the state_code was somehow parsed incorrectly from the domain.

src/auth0/rules/02-update-app-metadata.js
daschi: I kept the state_code setting for recidiviz and csg users in this rule, in case we need to change how that works on a per user log in basis. The reason Idaho is being set in both places is because we already have ID users registered that have `us_idaho` state codes in their app_metadata, so this is to fix those instances.

jovergaag: Oh that makes sense, smart since we may need to be testing lots of variations.

jovergaag (commented):
I think this looks ok, I'm going to hold off on approving until we get some good testing done on the `daschi-1034-restrict-with-access-token`. Or would you rather merge this into that branch and include it in the big QA push tomorrow when the new schema is available?

daschi:
@jovergaag no need! It doesn't affect that PR so happy to keep it separate and merge afterwards.

jessex:
Sorry I haven't been keeping up-to-date with this PR, @daschi. Would a review from me still be helpful? Let me know!

daschi:
@jessex No worries, I just wanted to keep you updated that we were changing the hooks around. I'll test them in staging before moving them to production, and they're currently blocked behind today's production deploy and also PR #1094.

jessex (commented):
Just took a quick breeze through and have no security concerns!

PR #1107, opened 2021-05-24
Use Redis cache instead of memory cache for all metric types

### LINKED ISSUE #1089: Use Redis cache instead of memory cache for all metric files
It is becoming burdensome to deploy the backend in order to refresh the cache for certain metric files (namely, US_ND). Also, the latency for a request with a cache miss is significant and has been happening more frequently on the ND dashboard so by moving it over to redis we will never see a cache miss unless something has gone terribly wrong.

**Additional context**

jovergaag:
## Description of the change

Uses redis cache instead of memory cache for all metric types. This changes the endpoints to use the Redis cache, as well as updates `refreshCache` endpoint to accept `metricType` param which is then used in `refreshRedisCache` to cache files appropriately depending on the metric type.

Also updates the cron schedule to refresh all of the Vitals, goals, and explore data to be refreshed and cached on a nightly basis.

Note: e2e-lantern tests are currently failing because SSO is enabled and the we don't have test accounts working quite yet.

## Type of change

## Related issues

> Closes #1089

### Development

### Code review

### DIFF
--- a/cron.yaml
+++ b/cron.yaml
@@ -1,12 +1,42 @@
 cron:
-  - description: "Refresh US_PA external cache"
-    url: /api/US_PA/refreshCache
+  - description: "Refresh US_PA newRevocation external cache"
+    url: /api/US_PA/newRevocation/refreshCache
     # Every day at 06:00 UTC / 02:00 EST / 23:00 PST
     schedule: every day 06:00
     target: default
 
-  - description: "Refresh US_MO external cache"
-    url: /api/US_MO/refreshCache
+  - description: "Refresh US_MO newRevocation external cache"
+    url: /api/US_MO/newRevocation/refreshCache
     # Every day at 06:30 UTC / 02:30 EST / 23:30 PST
     schedule: every day 06:30
+    target: default
+
+  - description: "Refresh US_ID populationProjections external cache"
+    url: /api/US_ID/populationProjections/refreshCache
+    # Every day at 07:00 UTC / 03:00 EST / 00:00 PST
+    schedule: every day 07:00
+    target: default
+
+  - description: "Refresh US_ND vitals external cache"
+    url: /api/US_ND/vitals/refreshCache
+    # Every day at 07:00 UTC / 03:00 EST / 00:00 PST
+    schedule: every day 07:00
+    target: default
+
+  - description: "Refresh US_ND goals external cache"
+    url: /api/US_ND/goals/refreshCache
+    # Every day at 07:00 UTC / 03:00 EST / 00:00 PST
+    schedule: every day 07:00
+    target: default
+
+  - description: "Refresh US_ND communityExplore external cache"
+    url: /api/US_ND/communityExplore/refreshCache
+    # Every day at 07:00 UTC / 03:00 EST / 00:00 PST
+    schedule: every day 07:00
+    target: default
+
+  - description: "Refresh US_ND facilitiesExplore external cache"
+    url: /api/US_ND/facilitiesExplore/refreshCache
+    # Every day at 07:00 UTC / 03:00 EST / 00:00 PST
+    schedule: every day 07:00
     target: default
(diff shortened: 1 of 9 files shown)


cron.yaml
```diff
@@ -1,12 +1,42 @@
...
+  - description: "Refresh US_MO newRevocation external cache"
+    url: /api/US_MO/newRevocation/refreshCache
     # Every day at 06:30 UTC / 02:30 EST / 23:30 PST
     schedule: every day 06:30
     target: default
+
+  - description: "Refresh US_ID populationProjections external cache"
+    url: /api/US_ID/populationProjections/refreshCache
+    # Every day at 07:00 UTC / 03:00 EST / 00:00 PST
+    schedule: every day 07:00
```
daschi, line 17: Do we need to space these out at all timewise?

server/core/__tests__/cacheManager.test.js
```diff
@@ -67,24 +67,14 @@ describe("cacheManager", () => {
...
-          IS_DEMO: "false",
-          NODE_ENV: "development",
-        });
-        jest.resetModules();
-        getCache = require("../cacheManager").getCache;
-      });
-      it("returns a redis cache", () => {
-        const cache = getCache("");
-        expect(cache.store.name).toEqual("memory");
+      it("returns a redis cache for facilitiesExplore", () => {
```
daschi, line 75: Now that `getCache` doesn't require a cacheKey anymore, this test can be simplified to just test that it returns a redis cache when the environment is not test or demo.

server/core/__tests__/refreshRedisCache.test.js
```diff
@@ -71,146 +71,193 @@ describe("refreshRedisCache", () => {
...
-    it("returns an error when caching fails", (done) => {
-      const error = new Error("Error setting cache value");
-      mockCache.set.mockImplementationOnce(() => {
-        throw error;
+  describe("when metricType is not newRevocation", () => {
+    describe("refreshing the cache for files without subsets", () => {
+      beforeEach(() => {
+        fileName = "random_file_name";
+        metricType = "vitals";
+        metricFile = { [fileName]: fileContents };
```
daschi, line 225: Since these metric types have more than one file included, can we update this test to reflect that?

server/core/refreshRedisCache.js
```diff
@@ -62,29 +62,38 @@ function cacheEachSubsetFile(
...
+        );
+        cachePromises.push(...subsetCachePromises);
+      } else {
+        console.log(`Setting cache for: ${cacheKey}...`);
+        cachePromises.push(cache.set(cacheKey, metricFile));
+      }
+    });
+  } else {
+    const data = {};
+    Object.keys(files).forEach((fileKey) => {
```
daschi, line 91: Do we need to iterate over files here, or can we just call this directly:

`cachePromises.push(cache.set(cacheKeyPrefix, files))`

server/routes/__tests__/server.test.js
```diff
@@ -224,15 +224,15 @@ describe("Server tests", () => {
...
+        .get("/api/US_PA/newRevocation/refreshCache")
         .then((response) => {
           expect(response.statusCode).toEqual(403);
         });
     });
 
     it("should respond successfully when cron job header is valid", () => {
       return request(app)
-        .get("/api/US_PA/refreshCache")
+        .get("/api/US_PA/newRevocation/refreshCache")
```
daschi, line 235: Should we add some examples for the other metric types?

daschi (changes requested):
Nice! I'm happy that everything is moving towards a similar pattern, just have a few questions/comments.

cron.yaml
jovergaag: I don't think so. They're all really small and don't have the processing like the Lantern files do to create the subsets. They all ran successfully simultaneously last night in staging.

server/core/__tests__/cacheManager.test.js
jovergaag: Good catch, thx!

server/core/refreshRedisCache.js
jovergaag: You're right. I had thought that we needed to rebuild that object but i just tested it and it can just be passed through!

daschi (approved):

PR #1117, opened 2021-05-28
Adds PageVitalsStore

jovergaag:
## Description of the change

Adds a PageVitalsStore and moves the `currentEntityId` and `selectedMetricId` to  the store such that the logic is moved from the `PageVitals` component into the store.

Allows the Vitals specific components to pull their data from the store rather than props.

## Type of change

## Related issues

> Closes [#XXXX]

### Development

### Code review


Catacola (approved):
This is slick and makes so many of the components so much cleaner!

src/core/CoreStore/PageVitalsStore.ts
```diff
@@ -0,0 +1,313 @@
...
+  currentEntityId: string;
+
+  summaries: VitalsSummaryRecord[];
+
+  timeSeries: VitalsTimeSeriesRecord[];
+
+  selectedMetricId: MetricType;
+
+  constructor({ rootStore }: { rootStore: CoreStore }) {
+    makeObservable(this, {
```
daschi, line 57: What was the reason to not use `makeAutoObservable`?

src/core/CoreStore/PageVitalsStore.ts
```diff
@@ -0,0 +1,313 @@
...
+    this.selectedMetricId = METRIC_TYPES.OVERALL;
+    this.summaries = [];
+    this.timeSeries = [];
+
+    when(
+      () => !this.rootStore.metricsStore.vitals.isLoading,
+      () => {
+        const { summaries, timeSeries } = this.rootStore.metricsStore.vitals;
+        this.setSummaries(summaries);
+        this.setTimeSeries(timeSeries);
```
daschi, line 84: I'm wondering if we really need to set these on a `when` action, if we're also setting them initially to an empty array. Since the `metricsStore.vitals.[summaries|timeSeries]` is already an observable, then computed values like `currentEntitySummary` should update whenever the observable updates, without needing the `when` (I think? I haven't tried it)

src/core/CoreStore/PageVitalsStore.ts
```diff
@@ -0,0 +1,313 @@
...
+export function getSummaryStatus(value: number): SummaryStatus {
+  if (value < 70) return "POOR";
+  if (value >= 70 && value < 80) return "NEEDS_IMPROVEMENT";
+  if (value >= 80 && value < 90) return "GOOD";
+  if (value >= 90 && value < 95) return "GREAT";
+  return "EXCELLENT";
+}
+
+export default class PageVitalsStore {
+  protected readonly rootStore;
```
daschi, line 46: If we only ever need the `metricsStore` I wonder if we should initialize it with just that store, instead of the entire core store?

src/core/CoreStore/PageVitalsStore.ts
```diff
@@ -0,0 +1,313 @@
...
+  get timeSeriesDownloadableData(): DownloadableData | undefined {
+    if (!this.currentEntityTimeSeries) return undefined;
+
+    let labels = [] as string[];
+    let ids = [] as string[];
+    const datasets = [] as DownloadableDataset[];
+    Object.values(METRIC_TYPES).forEach((metricType: MetricType) => {
+      const metricData = this.currentEntityTimeSeries
+        .filter((d: VitalsTimeSeriesRecord) => d.metric === metricType)
+        .sort((a, b) => (a.date < b.date ? 1 : -1));
```
daschi, line 231: Do we need a TODO about sorting all the time series data on the backend?

src/core/PageVitals/index.tsx
```diff
@@ -27,56 +26,34 @@ import VitalsSummaryDetail from "../VitalsSummaryDetail";
...
-  );
+  const { metricsStore, tenantStore, pageVitalsStore } = useCoreStore();
+  const { isLoading, isError } = metricsStore.vitals;
+  const {
+    currentEntitySummary,
+    vitalsFiltersText,
+    lastUpdatedOn,
+    timeSeriesDownloadableData,
+    vitalsSummaryDownloadableData,
+  } = pageVitalsStore;
```
daschi, line 49: Since this is all coming from the vitals page store, maybe we can remove the word `vitals` from `filtersText` and `summaryDownloadableData`? I'm okay with it either way but just wanted to make the suggestion!

src/core/models/VitalsMetrics.ts
```diff
@@ -61,31 +61,45 @@ export function createVitalsTimeSeriesMetric(
...
+      timeSeries: observable.ref,
+    });
+
+    this.summaries = [];
+    this.timeSeries = [];
+
+    autorun(() => {
+      if (this.apiData) {
+        this.buildSummaries();
+        this.buildTimeSeries();
```
daschi, line 83: Oh I see these aren't computed anymore, what prompted the change?

src/core/CoreStore/PageVitalsStore.ts
Catacola: Looking at the queries, they should be ordered now. We should be able to drop this now.

src/core/CoreStore/PageVitalsStore.ts
jovergaag: None! Changed.

src/core/models/VitalsMetrics.ts
jovergaag: Both `PageVitals` and `PageVitalsStore` are using `VitalsMeterics.vitals`. With the computed vs an observable object, it was parsing the api data and generating the metric object twice rather than just doing that once when the data is loaded.

daschi: Oh hm that's unexpected, I would think that it wouldn't re-compute unless `this.apiData` has changed..

jovergaag: I thought about moving `isLoading` and `isError` to the `PageVitalsStore` and `PageVitals` watches that observable rather than the metric model observable but it really is the model that we are waiting on to load or to return an error so it didn't seem as direct. If the page were to use those observables from the store rather than the metric then we could turn this back into a computed.

src/core/PageVitals/index.tsx
jovergaag: great suggestion!

src/core/models/VitalsMetrics.ts
jovergaag: changed it back

src/core/CoreStore/PageVitalsStore.ts
jovergaag: Good call!

src/core/CoreStore/PageVitalsStore.ts
jovergaag: So trying to set them outside of a `when` reaction gives this error: `Computed value VitalsMetrics@32.summaries is being read outside a reactive context. Doing a full recompute`

This is the constructor: 
```
    makeAutoObservable(this, {
      summaries: observable.ref,
      timeSeries: observable.ref,
    });
    this.metricsStore = metricsStore;
    this.currentEntityId = DEFAULT_ENTITY_ID;
    this.selectedMetricId = METRIC_TYPES.OVERALL;

    const { summaries, timeSeries } = this.metricsStore.vitals;
    this.setSummaries(summaries);
    this.setTimeSeries(timeSeries);
  }
```

Changing the constructor to:
```
  constructor({ metricsStore }: { metricsStore: MetricsStore }) {
    makeAutoObservable(this, {
      summaries: observable.ref,
      timeSeries: observable.ref,
    });
    this.metricsStore = metricsStore;
    this.currentEntityId = DEFAULT_ENTITY_ID;
    this.selectedMetricId = METRIC_TYPES.OVERALL;

    runInAction(() => {
      const { summaries, timeSeries } = this.metricsStore.vitals;
      this.summaries = summaries;
      this.timeSeries = timeSeries;
    });
  }
```
complains about summaries and timeSeries having no initializer.

src/core/CoreStore/PageVitalsStore.ts
jovergaag: I made this change, then went to merge it into my other open PR and realized that I need `currentTenantId` so I changed this back to `rootStore`.

daschi (approved):

PR #1158, opened 2021-07-01
IE11 banner

### LINKED ISSUE #1141: Improve performance on the projections pages for IE11
Both Community and Facilities projections pages load extremely slow in IE11. They are snappy in Chrome, but very slow in IE11, to the point that it could be interpreted as broken.

It seems to be the annotations in the chart that are taking so long to compute. But the chart is computing the annotations for both INCARCERATION and SUPERVISION compartments on whichever page is loaded first. When switching to the other view, the chart sees to hold those computations and the chart loads very fast. Maybe that is the first place to start, but filtering out data by compartment to be rendered in that page's chart.

Also - lets include a banner about using a different browser when on IE11! 

Banner behavior: 
- Show banner every time the user logs in. 
- Attach the banner to the body of the page, so that if the user scrolls, the banner will hide, but will still be there when they scroll back up to the top. 
- If the user clicks the "X" to dismiss the banner, hide the banner until the next time the user logs in.  

Mocks: https://www.figma.com/file/SOdY6JujMpbawgJ4jjDmE8/Population-Projection-MVP?node-id=2046%3A2

**Additional context**

jovergaag:
## Description of the change

Adds an IE11 banner at the top of the page prompting the user to use a different browser for a better experience. The banner should hide on scroll down the page, or when the close button is clicked. The banner should remain hidden for the rest of the user's session, including if they navigate to a different page.

Design: https://www.figma.com/file/SOdY6JujMpbawgJ4jjDmE8/Population-Projection-MVP?node-id=711%3A0

## Type of change

## Related issues

> Closes #1141

### Development

- [x] E2E have been run for Lantern changes ([Steps in the Readme](https://github.com/Recidiviz/pulse-dashboard#running-e2e-tests))

### Code review


src/components/IE11Banner.tsx
```diff
@@ -0,0 +1,86 @@
...
+      <Banner visible={pageStore.ie11BannerIsVisible}>
+        <BannerText>
+          Looks like you’re using Internet Explorer 11. For faster loading and a
+          better experience, use Microsoft Edge, Google Chrome, or Firefox.
+        </BannerText>
+        <BannerClose>
+          <button
+            type="button"
+            className="close"
+            onClick={() => pageStore.hideIE11Banner()}
```
Catacola, line 75: Nit: since you're just calling a function without passing or adding any args, you don't need to wrap it in an anonymous function:
```suggestion
            onClick={pageStore.hideIE11Banner}
```

src/components/TopBar/TopBar.tsx
```diff
@@ -16,35 +16,33 @@
...
 import React from "react";
-import PropTypes from "prop-types";
 import cn from "classnames";
 
-import { usePageState } from "../../contexts/PageContext";
+import { useRootStore } from "../StoreProvider";
 
-const TopBar = ({ children, isHidable = false, isWide = false }) => {
-  const { hideTopBar } = usePageState();
+interface Props {
```
Catacola, line 23: Three cheers for typescript!

Catacola (approved):
This all looks good!

One minor thing I noticed was that the text seems cramped on narrow screen. Maybe we can reduce the margins when the screen gets narrow?

Screen Shot 2021-07-02 at 2 50 02 PM

jovergaag:
> This all looks good!
> 
> One minor thing I noticed was that the text seems cramped on narrow screen. Maybe we can reduce the margins when the screen gets narrow?
> 
> Screen Shot 2021-07-02 at 2 50 02 PM

I ended up adjusting all of the margins per suggestion from @soniachokshi. This is what the cramped screen looks like now: 

Screen Shot 2021-07-06 at 12 00 55 PM

src/components/IE11Banner.tsx
jovergaag: Good catch!

Catacola (approved):

PR #1237, opened 2021-08-19
[Security] Gate API requests by state code

### LINKED ISSUE #1227: [Pen Test Finding] Fix authorization bug allowing users to access other state's data
**Describe the bug**
Users are able to copy an auth token into an API request for a state that they are not authorized to view, and receive the other state's data. 

For more details, see the [Cobalt finding](https://app.cobalt.io/recidiviz/pulse-dashboard-and-case-triage-application-august-2021-pt7441/findings/4)

**To Reproduce**

**Expected behavior**

**Screenshots**
Screenshots revealing sensitive information _must not_ be shared. If a screenshot is helpful in diagnosing and resolving the bug, please send this directly to a contact on the Recidiviz team.

 - OS [e.g. Windows 10]: 
 - Version [e.g. 22]: 

 - Device: [e.g. iPhone6]: 
 - OS: [e.g. iOS8.1]: 
 - Version [e.g. 22]: 

**Additional context**

Catacola:
## Description of the change

This PR adds validation to all routes of the form `/api/:stateCode/*/` (state_code being of the form us_xx) to ensure that the requester has the same state_code in their `app_metadata` from Auth0 (or is a Recidiviz user).

There is now an explicit list of routes which don't get their jwt validated and processed. All other routes (even those not of the form above) will get checked for a valid jwt by default.

Beyond the unit tests below, I tested by exploring the entire site and by curling unprotected routes.

## Type of change

## Related issues

Closes #1227 

### Development

### Code review


server/app.js
```diff
@@ -115,10 +124,18 @@ function errorHandler(err, _req, res, next) {
   }
 }
 
-if (isDemoMode) {
-  checkJwt = (req, res, next) => {
-    next();
-  };
+if (!isDemoMode) {
```
Catacola, line 127: @jovergaag From your demo branch, I am under the impression that this is set for local demo mode but not deployed demo mode so these checks will be enabled for the pen testers when they retest. Please correct me if I'm wrong.

server/app.js
```diff
@@ -24,16 +24,25 @@ const multer = require("multer");
...
+const { validateStateCode } = require("./utils/validateStateCode");
 
 const app = express();
 
 const upload = multer();
 
+const unprotectedRoutes = [
+  "/api/demoUser",
+  "/_ah/warmup",
+  "/api/:stateCode/:metricType/refreshCache",
```
jessex, line 42: Why is this one "unprotected?" Is there something else happening to ensure only admins can call it?

server/utils/validateStateCode.js
```diff
@@ -0,0 +1,57 @@
...
+ */
+
+const { pathToRegexp } = require("path-to-regexp");
+
+const { getAppMetadata } = require("./getAppMetadata");
+
+const validateStateCode = (exemptRoutes) => {
+  return (req, res, next) => {
+    const matchingRoutes = exemptRoutes.filter(
+      (route) => pathToRegexp(route).exec(req.originalUrl) !== null
```
jessex, line 28: This is case-insensitive, I presume?

server/utils/validateStateCode.js
```diff
@@ -0,0 +1,57 @@
...
+
+    if (matchingRoutes.length > 0) {
+      next();
+      return;
+    }
+
+    const reqStateCode = req.params.stateCode.toLowerCase();
+
+    // only filter on state codes (us_xx)
+    if (/us_[a-z][a-z]/.exec(reqStateCode) === null) {
```
jessex, line 39: Why is it that there is separate treatment for "exempted routes" and routes without stateCode in the path? Why can't the latter be explicitly listed in the former? The upside to that is it should make it so that if someone in the future mistakenly adds a route without stateCode when they meant to include stateCode, it would mean that instead of just letting requests silently next their way through, it would fail at dev time, because presumably we'd have some sort of test that ensures all routes either have stateCode or are exempted.

jessex (changes requested):
Thank you for this, @Catacola ! Just a few questions from me.

server/utils/validateStateCode.js
```diff
@@ -0,0 +1,57 @@
...
+      next();
+      return;
+    }
+
+    const metadata = getAppMetadata(req);
+    let userStateCode = metadata.state_code || "";
+    userStateCode = userStateCode.toLowerCase();
+
+    if (userStateCode === "recidiviz" || userStateCode === reqStateCode) {
```
jovergaag, line 48: CSG users have a `lantern` state code, which should allow access to US_PA and US_MO.

server/app.js
```diff
@@ -24,16 +24,25 @@ const multer = require("multer");
...
+  "/file/:name",
```
jovergaag, line 43: I think we need to add `/api/generateFileLink` to this list.

server/utils/validateStateCode.js
```diff
@@ -0,0 +1,57 @@
...
+    if (matchingRoutes.length > 0) {
+      next();
+      return;
+    }
+
+    const reqStateCode = req.params.stateCode.toLowerCase();
+
+    // only filter on state codes (us_xx)
+    if (/us_[a-z][a-z]/.exec(reqStateCode) === null) {
+      next();
```
jovergaag, line 40: All of our routes that are protected are formatted as `/api/:stateCode/*`. Do we want to call next here if the state code is not formatted properly? I would think we would want to send a 400?

server/utils/__tests__/validateStateCode.test.js
```diff
@@ -0,0 +1,109 @@
...
+  expect(resp.sendStatus.mock.calls.length).toBe(1);
+  expect(resp.sendStatus.mock.calls[0][0]).toBe(401);
+};
+
+describe("validateStateCode", () => {
+  beforeEach(() => {
+    jest.resetModules();
+  });
+
+  it("should call next() if stateCode in request matches state code in app_metadata", () => {
```
jovergaag, line 68: We may want to add a test with an upper case state code just to be safe.

jovergaag (changes requested):
This looks awesome! Just a few comments.

server/app.js
jovergaag: Or maybe not? We want it to be "protected" in the sense that we are using `checkJwt` but not validating the state code. Do we need to differentiate between the two?

server/utils/validateStateCode.js
jovergaag: @jessex I agree. The only route as this is written that is not "unprotected" but does not include stateCode is `/api/generateFileLink`. I think it makes sense to add that to the exempted list and add the `checkJwt` back into it.

Maybe the exempted list can be renamed to `routesWithoutStateCode` and we add the `checkJwt` back into each route individually to be very explicit. 

Since @Catacola is OOO today, I'll push up a commit to see what you think.

server/utils/validateStateCode.js
jovergaag: Yes. I added some tests to confirm that.

jovergaag:
@jessex @daschi This is ready for another look!

daschi (approved):
🔒

server/app.js
```diff
@@ -124,18 +116,10 @@ function errorHandler(err, _req, res, next) {
...
-  app.use(
-    checkJwt.unless({
-      path: unprotectedRoutes.map((r) => pathToRegexp(r)),
-    })
-  );
-
-  // Verify that the user has access to state-specific date
-  // for all state-specific routes
-  app.use("/api/:stateCode/", validateStateCode(unprotectedRoutes));
+if (isDemoMode) {
```
jessex: Sorry if I missed this, but doesn't this mean that in both local mode and in demo mode there won't be any stateCode validation, so when we deploy this and ask the pin testers to retest they'll just find the same vulnerability has before?

server/app.js
```diff
@@ -157,6 +159,21 @@ app.get("/_ah/warmup", () => {
...
+  "/api/demoUser",
+  "/_ah/warmup",
+  "/api/:stateCode/:metricType/refreshCache",
+  "/file/:name",
+  "/api/generateFileLink",
+];
+
+// Verify that the user has access to state-specific date
+// for all state-specific routes
+app.use(
```
jessex, line 172: Will this apply even in demo mode above?

jessex (commented):
One question!

server/app.js
jovergaag: In the deployed demo app, `isDemoMode` currently resolves to `false`. I know its confusing, I was in the middle of clearing up the terminology and ENV vars and in the code when the Vitals flurry started, so it's still in this in between state. I've got a commit with this DEMO -> OFFLINE name change that I can add here to make it more clear, but I lean towards keeping those two changes isolated since they are such big changes.

The new ENV vars are in `1Password` if you wanted to look at them. In the deployed demo app environment, `IS_DEMO = false`

I did realize yesterday evening that moving the state code validation middleware to the bottom of the file, means that it was never being hit, so the commit from today fixes that. I moved it back up and the `server.test.js` now reflects the new error message from the custom middleware state code validation.

jessex: Thanks for the explanation! Yes let's keep those changes isolated from one another. I see that this should work as expected now with the latest commit and test coverage.

jessex (approved):
I think we're good to go now! Thanks so much, @jovergaag and @Catacola!

server/app.js
jovergaag: Yes. That `isDemoMode` block only overrides the `checkJwt` middleware since in the tests and in local mode we are not authorizing users. The request will fall through to the rest of the middleware and routes in demo mode (which, until I can change the name, is actually local/test mode). In the deployed demo app, `isDemoMode = false` so the the user will be authorized on every route that specifies `checkJwt`

I really apologize for the confusion here. I was trying to do the demo mode transition in steps and got pulled away before having the chance to clean everything up.

jovergaag (approved):

PR #1267, opened 2021-08-31
download fixture script

jovergaag:
## Description of the change

Script to download fixture files to be used offline. It can be used to download .json files, or .txt files. If the requested file is .txt, it also downloads and formats the metadata file as well. Updates the README with instructions for use.

## Type of change

## Related issues

> Closes [#XXXX]

### Development

### Code review


daschi (approved):
This is cool and will be helpful if only a few of the fixtures are outdated. Were you also thinking of adding a similar script to the recidiviz-data backend to download all files at once?

server/core/demo_data/download_fixture.sh
```diff
@@ -0,0 +1,25 @@
+#!/usr/bin/env bash
+echo "Downloading data file"
+gsutil cp $1 server/core/demo_data/
```
Catacola, line 3: Let's add a helpful message to the user if they don't supply an argument. It may also be worth doing some validation to make sure they supplied something ending in `.txt` or `.json`. For testing sake, I tried running this with a wildcard. It managed to download the files but then didn't process any of them for obvious reasons.

We could also update this script to loop over all files returned and process them one-by-one.

Catacola (commented):
Thanks! This should make like easier.

How do we plan on using this with the contractors? Will we give them a list of files they should download and they'll need to manually feed them into this one-by-one?

What do you think of having a second script that will download all of them demo bucket and run it through the decoding below if it's a `.txt` file?

jovergaag:
> This is cool and will be helpful if only a few of the fixtures are outdated. Were you also thinking of adding a similar script to the recidiviz-data backend to download all files at once?

I was not planning on that. What is the use case to download all of the files? I thought the intention was to download the fixture for a single metric file to send to the contractor to enable them work on a specific page/chart?

daschi:
> I was not planning on that. What is the use case to download all of the files? I thought the intention was to download the fixture for a single metric file to send to the contractor to enable them work on a specific page/chart?

I think I was considering the case of needing to send them more than 1 metric at a time, running this script individually for each might become cumbersome. We could also see how this goes and if there's a need to download more at a time, update it then

PR #1279, opened 2021-09-02
Pathways navigation skeleton

### LINKED ISSUE #1263: Create navigation skeleton
Create navigation skeleton to prep for contractor work. No styling needed

**Additional context**

jovergaag:
## Description of the change

Adds the Pathways navigation skeleton to unblock work for simultaneous engineering work on Pathways next week.

This adds the Pathways Navigation header, the Section Navigation panel on the left side of the page, the Prison page, and the Supervision page. It also includes the permissions settings such that any Pathways page or chart can be hidden per user. By default, both the Prison and Supervision pages are restricted, except for Recidiviz users. 

I also tried to standardize the terminology for Pathways/Core path such that the path is laid out as `/view/page/section`. So for the Core path `/community/practices`, "community" is the view and "practices" is the page. For the Pathways path `/pathways/prison/populationOverTime`, "pathways" is the view, "prison" is the page", and "populationOverTime" is the section. These terms were not being used consistently so I tried to tighten them up everywhere I found them, hopefully was able to find all of the uses of these terms.

The only way to navigate to the Pathways view at this point is to manually set your URL: `localhost:3000/pathways`. You should be able to play around with the permissions by setting your Auth0 `routes` metadata:
  ```
"routes": {
    "prison_populationOverTime": true,
    "prison_gender": false,
    "pathways_prison": true,
    "pathways_supervision": false
  }
```

## Type of change

## Related issues

> Closes #1263

### Development

### Code review

### CODE REVIEW

File: src/core/CoreNavigation.tsx [hunk 1]
```diff
@@ -68,8 +73,8 @@ const CoreNavigation: React.FC = () => {
       </div>
       <div className="CoreNavigation__right">
         {flags.enableCoreTabNavigation && (
-          <CorePageSelector
-            currentSection={currentSection}
+          <PageSelector
```
jovergaag, line 76: Part of the Ilya's work will be to make the styling on this `PageSelector` dynamic depending on whether it is for Core or Pathways. So the Pathways navigation is pretty ugly right now...but it will get better!

src/core/PathwaysPageTemplate/PathwaysPageTemplate.tsx
```diff
@@ -0,0 +1,38 @@
...
+
+import PropTypes from "prop-types";
+import React from "react";
+
+import PathwaysSectionNav from "../PathwaysSectionNav";
+
+const PathwaysPageTemplate: React.FC = ({ children }) => (
+  <div className="main-content PathwaysPageTemplate">
+    <div className="PathwaysPageTemplate__body">
+      <PathwaysSectionNav />
```
jovergaag, line 28: I considered adding this `PathwaysSectionNav` into `PathwaysLayout` and using the more generic `PageTemplate` instead of this `PathwaysPageTemplate`, but the section nav will have a section specific title and copy that seemed too specific to be adding into the high level layout. Definitely open to thoughts on this!

src/core/PathwaysSectionNav/PathwaysSectionNav.tsx
```diff
@@ -0,0 +1,50 @@
...
+
+import { useRootStore } from "../../components/StoreProvider";
+import { useCoreStore } from "../CoreStoreProvider";
+
+const PathwaysSectionNav: React.FC = () => {
+  const { pathname } = useLocation();
+  const { userStore } = useRootStore();
+  const { setSection } = useCoreStore();
+  const navigationLayout = userStore.userAllowedNavigation;
+  const [currentView, currentPage] = pathname.split("/").slice(1, 3);
```
jovergaag, line 30: Should we add the view and page to the store instead of reading them from the pathname everywhere?

src/AuthWall/AuthWall.tsx
```diff
@@ -28,8 +29,11 @@ import { ERROR_MESSAGES } from "../constants";
  * Verifies authorization before rendering its children.
  */
 const AuthWall: React.FC = ({ children }) => {
+  const { pathname } = useLocation();
+  const currentView = pathname.split("/")[1];
   const { userStore, currentTenantId } = useRootStore();
   const { userAppMetadata } = userStore;
+  const unprotectedViews = ["profile", ""];
```
daschi, line 36: Another option could be to have a `SHARED_VIEWS` constant in `App.js` and just concat them for each layout, so we don't need to check unprotected views here. Open to either though so if you feel more strongly about this no worries.

src/core/PageNavigation.tsx
```diff
@@ -15,34 +15,34 @@
 // along with this program.  If not, see <https://www.gnu.org/licenses/>.
 // =============================================================================
 
-import "./CorePageSelector.scss";
+import "./PageNavigation.scss";
```
daschi, line 18: I like this rename!

daschi (approved):
Beautiful...I think the organization of `/view/page/section` makes sense. 🌈

src/core/PathwaysSectionNav/PathwaysSectionNav.tsx
daschi: I like that idea!

jovergaag: I played around with this a little this morning. I think its going to be really hard to add the currentView and currentPage to the store because we would need to use the `withRouteSync` functionality to sync the store with the path. However, the way our paths are being used to route to specific pages, we cannot add the view and page as params in the path. I'm open to ideas on how to do this, but I think it would take a fairly significant re-write to our overall routing strategy to accomplish. So I'm happy leaving it as is!

jovergaag:
@daschi Can you check the little boxes pretty pls.

src/AuthWall/AuthWall.tsx
jovergaag: Oh yes, thats better. Thanks!

Issue #1264, opened 2021-08-30
[Pathways] Design copy storage/fetch system


Catacola:
**Is your feature request related to a problem? Please describe.**
We want a system where designers can update copy by state without needing to publish a PR.

This issue is to put together a design doc for this system.

Requirements:
1. Editable by prod/design
2. Partitioned by state
3. Versioned (could be weak versioning, i.e. labeled by user and overwritable)

**Describe the solution you'd like**
Store the copy as plaintext in a GCP bucket and fetch it either at view time or compoile time.

**Describe alternatives you've considered**

**Additional context**
Add any other context about the feature request here.

PR #1287, opened 2021-09-06
Add state-specific overrides for view copy

Catacola:
## Description of the change

Adds state-specific overrides to copy system.

Right now, there are only overrides for US_ID, but can easily be added.

All fields and views are optional in the state-specific copy files. That is, you can override any view and within that view, you can override any field without needing to respecify any field you want to leave unchanged.

`retrieveViewCopy()` has been changed to a hook, `useViewCopy()`. This allows us to access the current tenant from this function so that we can automatically return the correct copy without needing to pass the tenant in as an argument.

## Type of change

## Related issues

Part of #1264 

### Development

### Code review


src/core/content/view/us_id.ts
```diff
@@ -0,0 +1,40 @@
...
+    summary: `This chart includes individuals who are admitted to IDOC facilities
+      and CAPP, including termers, riders, parole violators, and people held in county
+      jails under IDOC jurisdiction.`,
+  },
+  supervision: {
+    summary: `This charts includes all individuals that are currently on probation
+      and/or parole, including those in the Limited Supervision Unit.`,
+  },
+  supervisionToPrison: {
+    summary: `This charts includes people who have been incarcerated in a IDOC facility
```
jovergaag, line 34: Should this be "in an IDOC facility"? AN? My grammar bells are going off...but admittedly those bells have never been in tune so feel free to disregard. @soniachokshi

src/core/PagePrison/PagePrison.tsx
```diff
@@ -18,12 +18,12 @@
...
 import React from "react";
 
-import { retrieveViewContent } from "../content/retrieveContent";
+import useViewContent from "../hooks/useViewContent";
 import PathwaysLeftPanel from "../PathwaysLeftPanel";
 import PathwaysPageTemplate from "../PathwaysPageTemplate";
 
 const PagePrison: React.FC = () => {
-  const { title, summary } = retrieveViewContent("prison");
+  const { title, summary } = useViewContent("prison");
```
jovergaag, line 26: You know, as I'm looking at this and at the designs, we may not even need dedicated page components for each page. We could probably do a generic page and just pass in the view name. It's probably fine for now since its still early stages, but as we build things out, we may want to consider simplifying.

jovergaag (approved):
NICE. I love it.

src/core/PagePrison/PagePrison.tsx
Catacola: I was considering the exact same thing while writing this PR. I think all we'd need to add beyond this is a way to list the metrics that will be shown on each page and maybe the filters. I say maybe because I think we could derive the filter bar from the metrics shown, but it's possible we'll want more control than that.

jovergaag:
@Catacola Still looks good after the update from the base branch!

PR #1301, opened 2021-09-13
Enable users to navigate between compartments and transitions within Pathways 

### LINKED ISSUE #1262: Enable users to navigate between compartments and transitions within Pathways
The primary navigation _within_ the Pathways tool will be a set of tabs at the top of the page to allow the user to navigate between compartments (e.g., Prison) and transitions (e.g., Supervision → Prison). Each state will have a different number of tabs, depending on which compartments and transitions will have data and charts available. See the [Pathways MVP x States spreadsheet](https://docs.google.com/spreadsheets/d/1jmY_yKjLs7HUeCx3UKFHSCJdhTckuYokWr5fv9EEVfs/edit#gid=0) to see the list of pages for each state.

Go to [section in PRD →](https://docs.google.com/document/d/1IvWek2k1qfrfJLCWB1kijUihcln-LKH2e8-WQVHCcFo/edit#heading=h.h6w0s3e6ifoq)
Go to [mocks in Figma →](https://www.figma.com/file/Mw6kSEP2jdr9i7UVEjPuCI/UP-Consolidation?node-id=1141%3A2171)
Go to [prototype in Figma →](https://www.figma.com/proto/Mw6kSEP2jdr9i7UVEjPuCI/UP-Consolidation?page-id=776%3A2729&node-id=1141%3A2171&viewport=250%2C48%2C0.14&scaling=scale-down-width&starting-point-node-id=1141%3A2171&hide-ui=1)

A lot of the framework for this has been finished. The primary focus for this ticket is to add the styling to `PageNavigation`. That component is shared between two views (Pathways and the Core views), so we will need to pass in the view name to use as part of the classname for view specific styling. The styling on the component currently is for Core views, so please preserve that styling such that the core views (ie pages like `/community/practices`) still have the correct styling on their header.

The pages are added/removed for each state in `src/tenants.ts` in the `TENANTS` object. You can see that for US_ID, the "prison" and "supervision" pages have been enabled for the Pathways view. Eventually, there will be 4 pages enabled (Prison, Supervision, Supervision to Prison, and Supervision to Liberty), but only two are enabled right now.  If you want to add the two additional pages to help with the styling, you add them here.
```
navigation: {
      pathways: ["prison", "supervision"],
}
```

Please use the colors and fonts from `CoreConstants.scss`. These are our standardized styles across Recidiviz products. Let Jen know if you need a color or font that is not in that file. Thank you!

nasaownsky:
## Description of the change

Added styles for navigation within Pathways

## Type of change

## Related issues

> Closes #1262 

### Development

### Code review


src/core/CoreConstants.scss
```diff
@@ -18,6 +18,7 @@ $crimson-dark: #a43939;
 $crimson-dark-50: rgba(164, 57, 57, 0.5);
 $data-forest: #25636f;
 $data-forest-uncertainty: rgba(37, 99, 111, 0.15);
+$data-forest-light: rgba(37, 99, 111, 0.07);
```
jovergaag, line 21: The naming convention that we're trying to move to (but is not super apparent in this file) for our colors is `color-name-opacity-percentage`. So this name would be `$data-forest-07`

src/core/PathwaysNavigation/PathwaysNavigation.scss
```diff
@@ -21,18 +21,56 @@
...
-  display: flex;
   align-items: center;
   justify-content: space-between;
   border-bottom: $core-border-color 1px solid;
 
-  &__logo {
-    margin: 1.5rem 0 2rem 0rem;
-    // min-width hack so IE11 doesn't clobber the logo
-    min-width: 6.5rem;
+  .PageNavigation {
```
jovergaag, line 28: While this works and looks great, we're trying to keep styles for a given component entirely within that component's scss file. So, the preferred strategy would be to pass the view name in to `PageNavigation` and use that view name as a modifier to the `PageNavigation` classname. Then the `PageNavigation.scss` file would have view specific stiles in it, something like this:

```
.PageNavigation {
  view agnostic styles...

  &--Pathways {
     Pathways specific styles...
  }

  &--Core {
    Core specific styles...
  }
}
```

This is just an idea, if you find a better way to accomplish this please feel free to move forward. But we do want to keep all `.PageNavigation` styles inside of that scss file.

src/core/PathwaysNavigation/PathwaysNavigation.scss
```diff
@@ -21,18 +21,56 @@
...
+    & li {
+      flex-grow: 1;
+      flex-basis: 0;
+      margin: 0;
 
-    &-image {
-      height: 1.5rem;
+      .PageNavigation--Option {
```
jovergaag, line 34: Thank you for using BEM classnames! I know that you are working off of the classnames that already exist within the `PageNavigation` component. I think we should use this opportunity to clean them up a little bit! We should use lower case for the "element" and "modifier" names.  Also, in this component, `option` is the element and `selected` is the modifier, so I think these classnames should be `.PageNavigation__option--selected`, etc.

Thanks for doing this cleanup!

src/core/PathwaysNavigation/PathwaysNavigation.scss
jovergaag: Within the next few months the Core views will be deprecated, so we can clean this component/styling up at that point.

jovergaag (changes requested):
This looks great! I have a few requests, mostly around naming conventions and practices that we try to use within this code base.

Thank you.

nasaownsky:
Fixes arrived! 🚀

src/core/PageNavigation.scss
```diff
@@ -8,32 +8,92 @@
...
+
+          &:focus {
+            color: $pine-3;
+          }
+        }
+      }
+    }
+  }
+
+  &--Pathways {
```
jovergaag, line 48: This is a little bit redundant and messy now...but it will be short lived and we can clean it up once Core is deprecated. Thanks for making this change!

jovergaag (approved):

PR #1338, opened 2021-09-23
Shared profile page, Practices page inside of Pathways, and Pathways permissions

### LINKED ISSUE #1336: Enable a smooth transition for specific US_ID users from Core -> Pathways while maintaining Core for everyone else
How the heck are we gonna transition this?? :)

Make profile and practices available on Pathways for specific users, and make pathways their default redirect

**Additional context**

jovergaag:
## Description of the change

In anticipation of our transition to Pathways, we need a way to allow US_ND and US_ID users to have access to whichever layout is enabled for them. To do this, we need to make a few adjustments:
- Move the Profile outside of a layout, and style it depending on which layout the user came from
- Make the Practices page available from a `/practices` path, rendered inside of the `PathwaysLayout`
- Adjusts the route permissions such that Pathways pages/chart are restricted by permission, as well as the new `/practices` path

I still need to do some styling on the Practices page, but just putting this up for your viewing pleasure and for initial feedback!

@nasaownsky If you get started on the new Profile page styling before this is merged, you can use one or more of the commits from this branch!

## Type of change

## Related issues

> Closes #1336

### Development

### Code review


src/utils/navigation.ts
```diff
@@ -82,11 +100,15 @@ export function getAllowedNavigation(
 }
 
 export function getPathWithoutParams(pathname: string): string {
-  const navItems = pathname.split("/");
-  // navItems[0] is "" because of the leading /
-  const view: string = navItems[1];
-  const page: string = navItems[2];
-  return page ? `/${view}/${page}` : `/${view}`;
+  const viewsAndPages = Object.values(CORE_VIEWS)
```
jovergaag, line 103: This is pretty ugly, I know. I tried to pass the params into this function, but we call this from inside of `ProtectedRoute` and params aren't available inside of react-routr-dom Routes. This seemed like the best way to be able to cleanly determine which parts of the path are not params. Other suggestions welcomed!

Catacola: If params are of the form `/some/set/of/pages?cool_param=true&bad_param=false`, could you do something like this?
```
return pathname.slice(0,pathname.indexOf("?");
```

Catacola (commented):
What do you think of just having two different profile pages and components? That is, we keep the current profile page at `\profile` and build a new one at `\pathways\profile`? The link in the pathways nav bar can then just go directly to `\pathways\profile` and we don't need to do any checking of which site we're coming from to render the correct page. I believe this will also get us the access control on the new profile page so no one will accidentally land on it before they have access to pathways.

src/utils/navigation.ts
jovergaag: No because `:entityId` is a URI param, not a query param so it is to the left of the "?".

src/components/Profile.js
```diff
@@ -25,24 +24,25 @@ import StateSelector from "./StateSelector";
 import { useRootStore } from "./StoreProvider";
 
 const Profile = () => {
+  const { push, location } = useHistory();
+  // Use this boolean to render/style the correct profile page
+  const pathways = new URLSearchParams(location.search).get("pathways");
+  // clear the pathways indicator from the path until we consolidate to single Profile page
+  if (pathways) {
+    push({ pathname: "/profile" });
+  }
```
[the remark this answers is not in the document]
jovergaag: I did not know that `replace()` was an option. Amazing, thank you!

jovergaag:
> What do you think of just having two different profile pages and components? That is, we keep the current profile page at `\profile` and build a new one at `\pathways\profile`? The link in the pathways nav bar can then just go directly to `\pathways\profile` and we don't need to do any checking of which site we're coming from to render the correct page. I believe this will also get us the access control on the new profile page so no one will accidentally land on it before they have access to pathways.

I spent quite a bit of time trying to make this happen. The problem is that with the way we currently have set up our `/view/page/section` path, our page navigation and page permissions are tied pretty tightly to that format. So if we try to add `/pathways/profile`, we need to write a bunch of extra custom code to give the user permission to see the profile page, and even more custom code to allow the profile page as a valid page but not include it in the page navigation's dynamic list of pages.

I ended up making the path `/pathways-profile` which isn't amazing but its not awful...? Eventually, once we have deprecated the other profile page we can change that. LMK what you think!

Catacola (approved):

PR #1344, opened 2021-09-24
Add supervision to liberty page and chart

### LINKED ISSUE #1343: Create supervision to liberty metric and chart

jovergaag:
## Description of the change

HOLY MOLY it was easy to add this whole page and chart. 🎉 

Passed design review.

## Type of change

## Related issues

> Closes #1343

### Development

### Code review


src/core/ChartNote/ChartNote.tsx
```diff
@@ -32,6 +32,8 @@ const ChartNote: React.FC<Props> = ({ note, isLoading = false }) => {
     );
   }
 
+  if (!note) return <div />;
```
Catacola, line 35: Would it make sense to roll this into the above check so that we return the same black state when loading or when no text is specified? That is, is there a difference between a div.ChartNote with a break inside and an empty div?

src/core/CoreStore/MetricsStore.ts
```diff
@@ -115,10 +119,22 @@ export default class MetricsStore {
         FILTER_TYPES.SUPERVISION_TYPE,
       ],
       chartTitle: "Admissions to prison from supervision",
-      noteCopy: `The revocation rate is calculated by counting the number of revocations
-                 per month and dividing it by the total supervision population in the
-                 given month or year. In this chart, the total supervision population
-                 includes only those on in-state probation, parole, or dual supervision.`,
```
Catacola, line 121: I'm assuming the copy was removed since it didn't actually apply to the chart we were showing?

src/core/VizCountOverTimeWithAvg/VizCountOverTimeWithAvg.tsx
```diff
@@ -47,14 +47,14 @@ const VizCountOverTimeWithAvg: React.FC<VizCountOverTimeWithAvgProps> = ({
 
   const data = dataSeries.map((d, index) => ({
     index,
-    date: new Date(d.year, d.month),
+    date: new Date(d.year, d.month - 1),
```
Catacola, line 50: So that explains why I was seeing data from the future.

Something, something, JS... something something, in a fire.

src/core/content/page/us_id.ts
```diff
@@ -21,14 +21,26 @@
 import { StateSpecificPageCopy } from "../types";
 
 const content: StateSpecificPageCopy = {
+  prison: {
+    title: "Prison",
```
Catacola, line 25: If a field doesn't differ from the default, it's not necessary to specify it. However, if you think this makes it easier to follow state-specific copy, I'm open to changing the spec.

Catacola (approved):
I have some minor comments but otherwise this looks good.

src/core/CoreStore/MetricsStore.ts
jovergaag: @soniachokshi  asked that we remove it because the methodology hasn't been finalized yet for this chart. I think we will add it back in once that has been formalized.

@soniachokshi is that correct, or did you intend for me to remove it ONLY for the demo deployment?

src/core/VizCountOverTimeWithAvg/VizCountOverTimeWithAvg.tsx
jovergaag: If I hadn't been wrestling with a test a couple of weeks ago that taught me about `new Date(year, month INDEX)` this would have taken me such a long time to figure out :) 

Thanks js.

src/core/content/page/us_id.ts
jovergaag: You're right, these aren't necessary. I ran the script again without these state specific titles but fixed the thing that was actually causing my issue and these titles aren't needed. Removed! Good catch.

src/core/ChartNote/ChartNote.tsx
jovergaag: Yes. Not sure why I didn't do that to begin with :)

Issue #1166, opened 2021-07-16
Enable SSO and notify team


jovergaag:
**What needs to be done? Why does it need to be done?**
It's time to enable SSO for Recidiviz team members! Once it is enabled, send out an email to team@ to notify them of this change, and provide instructions for FE developers on how to create a test user under SSO.

**Additional context**
Add any other context about the task here.

PR #1348, opened 2021-09-28
[Auth] Add flag to app_metadata to block syncing of permissions

Catacola:
## Description of the change

Update the auth0 action used to sync user permissions. If the field `skip_sync_contacts` is set to true, Auth0 will not try to contact the recidiviz backend to update the user's permissions. This is useful in the context of test and temporary users.

## Type of change

## Related issues

Part of #1166 

### Development

### Code review


src/auth0/actions/04-update-user-restrictions.js
```diff
@@ -21,17 +21,27 @@
...
+  
+  const testerEmails = [];
+  const e2eTestEmails = [];
+
+  if (app_metadata.skip_sync_permissions ||
+    (email && (testerEmails.includes(email) ||
+      e2eTestEmails.includes(email)))
+  ) {
+    return;
+  }
```
Catacola, line 35: This is the real change; everything else is just `prettier`.

The bypass for testers and e2e testers was already in auth0, but it hadn't been synced to our repo. My change add the check for `skip_sync_permissions`.

jovergaag: Love it. Were you thinking we will only deploy this to `recidiviz-dev` or were you thinking also to `recidiviz` tenant?

jovergaag (approved):

src/auth0/actions/04-update-user-restrictions.js
Catacola: I don't see a specific need for it in prod, but it's going to be tricky to remember to remove it when we copy this script to prod. We could split this script into two, but then it would be a pain to remember to maintain them both. I don't think this will lead to a security hole. What are your thoughts?

PR #1454, opened 2021-11-16
Add "As of [date]" to prison population by facility chart

### LINKED ISSUE #1443: Pathways TN no data error state crashing page
On Tennessee > System-Level Trends > Prison > Prison Population over time chart, selecting the filters time period=5 years, gender=female, facility=MCCX throws an error because the chart returns no data that meets the criteria (there are no females incarcerated in the MCCX facility). Here is the URL: https://dashboard-staging.recidiviz.org/system/prison/countOverTime?ageGroup=All&facility=MCCX&gender=Female&legalStatus=All&timePeriod=5%20years

Currently, the page crashes and shows the following error:

The intended behavior is that a "no data" error state should be shown with a prompt to reset the filters, similar to what happens on the ID charts when there is no data.

https://www.figma.com/file/Mw6kSEP2jdr9i7UVEjPuCI/UP-Consolidation?node-id=1777%3A454

image

**Additional context**
cc @soniachokshi @jovergaag

### LINKED ISSUE #1446: Add "As of [date]" to prison population by facility chart
On the Tennessee > System-Level Trends > Prison population by facility chart, we want to make sure users know that the data is capturing the latest data that has been transferred to us. 

This task involves adding an "as of [date]" to the title of the "Prison Population by Facility" chart. 

The title string should now say: Prison Population by Facility as of [date].
The styling should be the same as the styling for the "List of people in prison as of [date]" chart. Here is the example of how it looks on the "List of people in prison" chart. 

**Additional context**

nasaownsky:
## Description of the change

1. Added "As of [date]" to prison population by facility chart
2. Resolved Pathways TN no data error state crashing page

## Type of change

## Related issues

Closes #1443 
Closes #1446 

### Development

### Code review


src/core/PopulationTimeSeriesChart/PopulationTimeSeriesChart.tsx
```diff
@@ -56,8 +56,10 @@ const PopulationTimeSeriesChart: React.FC<Props> = ({
   }));
 
   const { beginDate, endDate } = getDateRange(
-    historicalPopulation[0].date,
-    historicalPopulation.slice(-1)[0].date,
+    historicalPopulation.length > 0 ? historicalPopulation[0].date : new Date(),
+    historicalPopulation.length > 0
+      ? historicalPopulation.slice(-1)[0].date
+      : new Date(),
```
jovergaag, line 62: I think we probably want to return undefined or an empty div on line 57 if `historicalPopulation.length < 1`. If there is no historical population data we will not want to render the chart

src/core/VizPopulationSnapshot/VizPopulationSnapshot.tsx
```diff
@@ -49,6 +51,8 @@ const VizPopulationSnapshot: React.FC<VizPopulationOverTimeProps> = ({
     value: d.totalPopulation,
   }));
 
+  const latestUpdate = max(dataSeries.map((d) => d.lastUpdated)) || new Date();
```
jovergaag, line 54: These snapshots will all have the same date, so you can just take the first item in `dataSeries` and use its `lastUpdated` to improve performance of this chart. We should also default to "Unknown" here if there is no `lastUpdated` date.

I'm realizing now that we should make this same adjustment in `VizPopulationPersonLevel.tsx` on line 37.

jovergaag (changes requested):

src/core/PopulationTimeSeriesChart/PopulationTimeSeriesChart.tsx
nasaownsky: This task is performed by ModelHydrator. The error throws when interpreter tries to access historicalPopulation[0] element. The empty div or null doesn't work in that case.

src/core/VizPopulationSnapshot/VizPopulationSnapshot.tsx
nasaownsky: I don't quite understand what you mean by "Unknown". formatDate accepts only Date type, so if there's no data we need Date object. It will not be used anywhere, just prevents error throwing. Everything else handled by ModelHydrator.

src/core/PopulationTimeSeriesChart/PopulationTimeSeriesChart.tsx
jovergaag: I agree, I think we need to return prior to ever accessing historicalPopulation[0]. 

I also agree that this task is supposed to be performed by the `ModelHydrator`, but we are seeing the error often enough that it seems there may be a timing issue where we try to access the non-existent element here before the `ModelHydrator` catches and renders the no data state. I'm not sure exactly what is going on but we are definitely seeing an error from trying `historicalPopulation[0].date`

src/core/VizPopulationSnapshot/VizPopulationSnapshot.tsx
jovergaag: I think we would need to do something like this on line 54:
`const latestUpdate =formatDate(dataSeries[0].lastUpdated, "MMMM dd, yyyy") || "Unknown";`

Then just add `latestUpdate` directly to the string in the chart title.

nasaownsky:
@jovergaag So I took the most optimal approach in my opinion, that could help to avoid those errors in the future. We just need to keep in mind that in cases where we trying to access elements in objects that could possibly be empty we should use the optional chaining operator `?.`

src/core/PopulationTimeSeriesChart/helpers.ts
```diff
@@ -128,6 +128,12 @@ export const getDateRange = (
   beginDate.setDate(beginDate.getDate() - offset);
   endDate.setDate(endDate.getDate() + offset);
 
+  if (!firstDate && !lastDate)
+    return {
+      beginDate: new Date(9999, 11, 31),
+      endDate: new Date(9999, 11, 31),
```
jovergaag, line 134: I don't really agree with the use of the arbitrary dates here. It feels a little like a hack. I would prefer to guard against the case that we don't have historical data in the chart component directly to make it easier to read and maintain for future developers.

src/utils/formatStrings.ts
```diff
@@ -192,6 +192,8 @@ function formatISODateString(date: string): string {
 }
 
 function formatDate(date: Date, pattern = "M/d/yy"): string {
+  if (!date) return "Unknown";
```
jovergaag, line 195: I like this solution!

jovergaag (changes requested):

src/core/PopulationTimeSeriesChart/helpers.ts
nasaownsky: That is the case, we already guard against the case that we don't have historical data in the chart component directly by using the optional chaining operator `?.`, but then `getDateRange` receives nothing and the App crashes agains, so we must guard against this either. And `getDateRange` must also necessarily return `Date` type, because otherwise other parts of the App start to crash.

nasaownsky: This is the case, we already guard against the case that we don't have historical data in the chart component directly by using the optional chaining operator ?., but then getDateRange receives nothing and the App crashes agains, so we also must guard against this. And getDateRange must also necessarily return Date type, because otherwise other parts of the App start to crash.

jovergaag (approved):

Issue #940, opened 2021-03-23
Move width of filter bars from JS to CSS


Catacola:
**What needs to be done? Why does it need to be done?**
The different filter dropdowns in FilterBar.tsx have their widths defined in js. I should move them to css to keep future developers from going crazy and hunting me down.

**Additional context**
Add any other context about the task here.

PR #1492, opened 2021-11-29
Add multi-select component and logic to Pathways filters

### LINKED ISSUE #1437: Create multi-select filter for pathways
And clean up the multi-select filtering logic.

**Additional context**

jovergaag:
## Description of the change

Adds multi-select component and logic to the Pathways filters.
-enables Multi-select dropdown in `PathwaysFilterBar`
-adds configuration to `filterOptions` to specify which filters are multi select on a per state level (some states will use multi-select for some filters while other states will remain single select for the same filter)
-adds the logic such that all filters live in the `filtersStore` as arrays, such that any filter can be converted from single select to multi-select with a one line change.
-adjusts the filtering logic such that the filter component calls `setFilters` on change instead of `setQuery`. `withRouteSync` now just watches the `filtersStore` and `pagePracticesStore` and updates the path when a filter/selectedMetricId changes
-re-orders some of the filters to match the new order specified in the designs
-collapses `Download` and `Methodology` into an ellipse button in the filter bar now that the multi-select filters are wider and was forcing the details to wrap

## Type of change

## Related issues

> Closes #1437 #940 

### Development

### Code review


src/core/PathwaysFilterBar/PathwaysFilterBar.scss
```diff
@@ -0,0 +1,36 @@
+// Recidiviz - a data platform for criminal justice reform
+// Copyright (C) 2020 Recidiviz, Inc.
```
Catacola, line 2:
```suggestion
// Copyright (C) 2021 Recidiviz, Inc.
```

src/core/PathwaysFilterBar/PathwaysFilterBar.scss
```diff
@@ -0,0 +1,36 @@
...
+  }
+  #supervisionType {
+    width: 11rem;
+  }
+  #ageGroup {
+    width: 16rem;
+  }
+  #facility {
+    width: 11rem;
+  }
```
Catacola, line 35: Thankyouthankyouthankyouthankyouthankyouthankyou

src/core/controls/MultiSelect/CoreMultiSelect.js
```diff
@@ -126,6 +129,7 @@ CoreMultiSelect.propTypes = {
   options: PropTypes.arrayOf(optionPropType).isRequired,
   className: PropTypes.string,
   summingOption: optionPropType.isRequired,
+  isChanged: PropTypes.bool,
```
Catacola, line 132: I don't see this prop used anywhere in the component. The only thing I could see it doing is being passed to `ReactSelect` via `{...props}`, but I don't see an `isChanged` prop in the react select docs: https://react-select.com/props

src/core/models/PopulationOverTimeMetric.ts
```diff
@@ -55,7 +55,8 @@ export default class PopulationOverTimeMetric extends PathwaysMetric<PopulationT
       facility,
       ageGroup,
     } = this.rootStore.filtersStore.filters;
-    const monthRange: MonthOptions = parseInt(timePeriod) as MonthOptions;
+    // the timePeriod filter will only ever be single-select so always use the 0 index
+    const monthRange: MonthOptions = parseInt(timePeriod[0]) as MonthOptions;
```
Catacola, line 59: This is beginning to tickle my senses. If we need to put this comment everywhere, it feels like we're doing something wrong. However, I can't think of a good way to make the timePeriod filter just return one value without making the typing weird for all the filters. I can think of two options, but I'm unsure of their drawbacks:
1) Create a special filter class for things we know will always be single-select and use that for this filter.
2) Make a small helper function along the lines of `parseTimeFilter = (timePeriod: someType): MonthOptions => parseInt(timePeriod[0]) as MonthOptions`. This is a repeated enough pattern in this codebase and there are three things going on here (indexing, parsing, and casting), so it doesn't feel that frivolous.

src/core/models/__tests__/PopulationOverTimeMetric.test.ts
```diff
@@ -97,6 +97,7 @@ describe("PopulationOverTimeMetric", () => {
         FILTER_TYPES.AGE_GROUP,
         FILTER_TYPES.FACILITY,
       ],
+      enabledMoreFilters: [],
```
Catacola, line 100: I know these are only in tests, but seeing as we need to specify `enabledMoreFilters: []` over and over, what do you think of making this an optional prop that defaults to `[]`?

src/core/withRouteSync.tsx
```diff
@@ -95,31 +97,77 @@ const withRouteSync = <Props extends RouteParams>(
...
               convertLabelsToValues(
                 cleanQuery as PopulationFilterLabels,
                 filtersStore.filterOptions
               )
             );
+          }
+          setInitialLoad(false);
+        }
       })
     );
```
Catacola, line 131: If you pass an empty array as the second argument to use effect (e.g. `useEffect(() => {some_code()}, [])`), it should only run on the initial render. I think you then   can drop the `initialLoad` state.

src/utils/formatStrings.ts
```diff
@@ -82,11 +82,11 @@ function raceValueToHumanReadable(raceValue: string): string {
 }
 
 function toHtmlFriendly(string: string): string {
-  return string.replace(/\W+/g, "-");
+  return string && string.replace(/\W+/g, "-");
 }
 
 function toHumanReadable(string: string): string {
-  return string.replace(/[-_]/g, " ");
+  return string && string.replace(/[-_]/g, " ");
```
Catacola, line 89: I'm not sure what these `&&` checks are for. Typescript says these should always be strings, not nullable strings, so this check shouldn't be necessary. If the string is nullable, you should update the types and you can probably change them to `string?.replace()`.

Catacola (changes requested):
Everything works great, thanks! I have a few comments below. Feel free to push back on anything if you feel I'm offbase.

src/core/types/filters.ts
```diff
@@ -60,7 +60,7 @@ export interface PopulationFilters {
 export type PopulationFilter = {
   type: keyof PopulationFilters;
   title: string;
-  width: string;
+  isMultiSelect: boolean;
```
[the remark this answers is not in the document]
jovergaag: Great idea. I ended up switching it to `isSingleSelect` since there are far fewer single select filters and defaulting it to false.

src/core/controls/MultiSelect/CoreMultiSelect.js
jovergaag: It is being used in `coreSelectCustomStyles` which lives in another component. I will move it out to avoid confusion.

src/core/models/PopulationOverTimeMetric.ts
jovergaag: I wrestled with this for a while as well (both before your review and after, to no avail). I agree that it is smelly. If we make the custom single-select class it generates a ton of additional code to handle both arrays and strings.

I ended up adding a getter to the FiltersStore which returns the `monthRange`. It is definitely better than it was. Let me know if you still think it could use some work.

src/core/withRouteSync.tsx
jovergaag: whaat? I did not know this.

src/utils/formatStrings.ts
jovergaag: They were causing problems at some point, but seem to be working fine without the guard now that things are more clean.

jovergaag:
@Catacola Thank you for the amazing feedback! It is so much better now.

src/core/withRouteSync.tsx
Catacola: Yup! This is because the second argument is the list of props that, when changed, trigger a rerun of the hook added in the first argument. If the argument is `undefined`, it will run whenever the component is rerendered. However, if it is supplied, it will only rerender if a supplied argument prop changed and since there are no props in `[]`, it will never rerun.

For a more official sounding explanation, see the note after https://reactjs.org/docs/hooks-effect.html#tip-optimizing-performance-by-skipping-effects.

src/core/withRouteSync.tsx
```diff
@@ -95,29 +96,76 @@ const withRouteSync = <Props extends RouteParams>(
       viewId === PATHWAYS_VIEWS.operations || pageId === CORE_PAGES.practices
         ? metricQueryParams
         : filterQueryParams;
-    const [query] = useQueryParams(queryParams);
+    const [query, setQuery] = useQueryParams(queryParams);
     const cleanQuery = removeUndefinedValuesFromObject(query);
 
     // eslint-disable-next-line react-hooks/exhaustive-deps
```
Catacola, line 102: You may not still need this line. I'm honestly not sure.

Catacola (approved):
Nice work

PR #1530, opened 2021-12-17
Creates US_DEMO filter options and adds Pathways fixtures to repo

### LINKED ISSUE #1499: Create demo mode filter options
A clear and concise description of what task needs to be accomplished and why.

**Additional context**

jovergaag:
## Description of the change

We were seeing quite a bit of fixture churn every time we added a new state because each state has different dimension values. This body of work allows the app to use a pre-defined set of demo filter options when the app is running in `demo` or `offline` modes such that the same set of fixtures can be used regardless of the state.

Now that the fixtures will remain constant for a while, I added them to the repo to make it easier to version control them.

It will be easiest to review this commit by commit since there are so many fixtures that were added in the last commit. This is also deployed to the demo app if you want to QA there: dashboard-demo.recidiviz.org

NOTE: Do not merge until the Tuesday production deploy. This relies on the schema change in that deploy.

## Type of change

## Related issues

> Closes #1499 

### Development

### Code review


src/core/models/SupervisionPopulationOverTimeMetric.ts
```diff
@@ -57,12 +57,14 @@ export default class SupervisionPopulationOverTimeMetric extends PathwaysMetric<
...
         return (
+          Math.abs(monthsOut) <= monthRange &&
           gender.includes(record.gender) &&
           supervisionType.includes(record.supervisionType) &&
           district.includes(record.district) &&
           mostSevereViolation.includes(record.mostSevereViolation) &&
           numberOfViolations.includes(record.numberOfViolations) &&
-          Math.abs(monthsOut) <= monthRange
+          ["ALL"].includes(record.supervisionLevel) &&
+          ["ALL"].includes(record.race)
```
macfarlandian, line 67: i don't necessarily have a problem with this but why not just use `=== "ALL"` instead?

macfarlandian (approved):

src/core/models/SupervisionPopulationOverTimeMetric.ts
jovergaag: Because this sets us up to quickly and easily change to `race.includes(record.race)` once the race filter is enabled in January. The fixtures include the `race` dimension, so this is allowing us to support the future proofed fixtures while waiting for the new filters to be implemented.

macfarlandian: nice. makes sense!

PR #1587, opened 2022-01-25
Dynamic filter bar

### LINKED ISSUE #1511: Set up filter order per compartment and dynamically add filters to modal based on width
Let's set up a static filter order per compartment (Prison/Supervision) and dynamically order and add filters to the primary filter bar or hide in the modal based on filter width such that the maximum number of filters always show up in the primary filter bar. This will remove some of the complexity in `enabledFilters` and reduce our need to keep the filters in that file ordered manually.

Filter order Prison:
TIME_PERIOD
FACILITY
GENDER
LEGAL_STATUS
AGE
RACE
OFFENSE_TYPE
NUM_PRIOR_INCARCERATIONS (not yet implemented)
TIME_SERVED (not yet implemented)

Filter order Supervision
TIME_PERIOD
DISTRICT
GENDER
SUPERVISION_TYPE
MOST_SEVERE_VIOLATION
NUMBER_OF_VIOLATIONS
AGE
RACE
LENGTH_OF_STAY
OFFENSE_TYPE

May be able to use the filter widths in `PathwaysFilterBar.scss` to determine the widths, or use another method if it is easier!

**Additional context**

nasaownsky:
## Description of the change

Set up filter order per compartment and dynamically add filters to modal based on width

## Type of change

## Related issues

Closes #1511 

### Development

### Code review


nasaownsky:
@jovergaag So I was able to create this feature, but struggle with bug: if we select page/section with fewer enabled filters, the next selected page/section will display fewer enabled filters too. I thinks the issue is in state managing somewhere in `useResizeFilterBar`. Can you take a look at it please?

jovergaag:
@nasaownsky I apologize for not looking at this sooner, we are focused on launching US_ND this week. Once that is out the door I will take a look at this bug. I'm glad you picked up some other issues in the meantime!

jovergaag:
@nasaownsky It is time to pick this up again! Can you please rebase it and I'll take a look at the bug that you were running into.

nasaownsky:
Good to go, take a look!

jovergaag:
> UPD: also, do we actually need this change? With the new alignment of the filters they look decent in their current form.

Yes I think so. We have been getting lots of feedback from the design team that they don't like the wrapping, and there are some pages that wrap on smaller screens.

src/core/PageSystem/PageSystem.tsx
```diff
@@ -41,10 +43,39 @@ const PageSystem: React.FC = () => {
...
+    FILTER_TYPES.AGE_GROUP,
+    FILTER_TYPES.RACE,
+  ];
+
+  const sortedEnabledFilters: EnabledFilters =
+    compartment === "SUPERVISION"
+      ? supervisionFiltersOrder.filter((item: any) =>
+          enabledFilters.includes(item)
+        )
+      : prisonFiltersOrder.filter((item: any) => enabledFilters.includes(item));
```
jovergaag, line 77: Can we just do this in the filters store in the `enabledFilters` getter? We should have access to the `metricStore.current` to be able to see what compartment it is. Then we can just store the filter order in the `utils/constants.ts` maybe?

src/hooks/useResizeFilterBar.ts
```diff
@@ -0,0 +1,70 @@
...
+//
+// You should have received a copy of the GNU General Public License
+// along with this program.  If not, see <https://www.gnu.org/licenses/>.
+// =============================================================================
+import { useEffect, useState } from "react";
+
+import { EnabledFilters } from "../core/types/filters";
+
+const useResizeFilterBar = (
+  filtersRef: any,
```
jovergaag, line 22: You can use `React.MutableRefObject<HTMLElement>` type here instead of any as wella s on `containerRef`. You will just need to case them in `PathwaysFilterBar`
```
const filtersRef = React.useRef() as React.MutableRefObject<HTMLElement>;
  const containerRef = React.useRef() as React.MutableRefObject<HTMLElement>;
```

jovergaag: You will also need to add the `number` return type for the function to make lint happy!

src/hooks/useResizeFilterBar.ts
```diff
@@ -0,0 +1,70 @@
...
+    );
+
+    const handleResize = () => {
+      const containerWidth = containerRef.current?.offsetWidth;
+      const count = getRenderCount(filtersWidths, containerWidth, 140);
+      setRenderCount(enableMetricModeToggle ? count - 1 : count);
+    };
+
+    window.addEventListener("resize", handleResize);
+    handleResize();
```
jovergaag, line 56: I am resizing my window and don't see the the filters responding as I resize. When I refresh the page the new filters appear/disappear. Do you see it appearing during resize? I don't think this is super important for this feature, what really matters is that the filters look good when the page is loaded on any screen width. I just want to make sure I'm not missing something!

jovergaag:
I made a couple of comments here. Overall, this is super slick! Its working so well, I'm super excited about it. I haven't had a chance to look more into that bug you mentioned but will look at it in the morning!

src/hooks/useResizeFilterBar.ts
```diff
@@ -0,0 +1,70 @@
...
+      if (total + array[i] > outerWidth) {
+        return i;
+      }
+      total += array[i];
+    }
+    return array.length;
+  };
+
+  useEffect(() => {
+    const filtersWidths = Array.from(filtersRef.current?.children).map(
```
jovergaag, line 45: I think this is the source of the problem. When looking at `filtersRef.current.children` when switching from a chart WITH the metric mode toggle to one WITHOUT, you can see that the children don't change. 

I think what we need to do is force the `PathwaysFilterBar` to re-render whenever the metric id (or maybe `chartTitle`?) changes which will update the ref and cause the children to be re-populated as the whole set before being trimmed down to the smaller set. The way it is now, the filter bar is not re-rendering so the children never change when switching. The easiest way to do this is to add `key={metric.id}` as a prop to `PathwaysFilterBar` in `PageSystem` which will force that component to re-render when the metric id changes.

Give that a try, let me know how it goes.

jovergaag:
@nasaownsky This is ready for another look!

src/core/CoreStore/MetricsStore.ts
```diff
@@ -336,6 +336,7 @@ export default class MetricsStore {
       id: "supervisionPopulationOverTime",
       tenantId: this.rootStore.currentTenantId,
       sourceFilename: "supervision_population_time_series",
+      compartment: "SUPERVISION",
```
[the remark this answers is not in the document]
nasaownsky: That's a good point. Actually, can I have some sort of list telling what exactly metric is `SUPERVISION` and what is not?

jovergaag:
> errorInfo: { code: 'app/invalid-credential', message: "Failed to read credentials from file {PATH_TO_CREDENTIALS_FILE}: Error: ENOENT: no such file or directory, open '{PATH_TO_CREDENTIALS_FILE}'" }

No it was not only you. We have a new "product" that is being added to pulse-dashboard and have had some bumps getting the environment right to support the new firestore. 

We made a few changes, so now if you copy the `offline` section of `env-cmdrc.example` into your `.env-cmdrc` you should be up and running. If not, check your firebase version `firebase -V` and make sure you are on 10.4.0 or higher.

@nasaownsky

jovergaag (approved):
YESS

PR #1603, opened 2022-01-26
Use count from ALL row to determine totalCount

jovergaag:
## Description of the change

Moves the FE away from using the `total_population` column and instead uses the count from the "ALL" row as the `totalCount`.

## Type of change

## Related issues

> Closes [#XXXX]

### Development

### Code review


src/core/models/PrisonPopulationSnapshotMetric.ts
```diff
@@ -33,14 +34,51 @@ import { filterTimePeriod } from "./utils";
...
+    this.totalCount = -1;
     this.download = this.download.bind(this);
+
+    reaction(
+      () => !this.isLoading,
+      () => {
+        this.totalCount = this.getTotalCount;
+      }
+    );
+  }
```
macfarlandian, line 55: this is a weird pattern ... `totalCount` should just be the getter and return -1 if there aren't any records

macfarlandian (changes requested):

src/core/models/PrisonPopulationSnapshotMetric.ts
jovergaag: I did it this way because we access `totalCount` several times over the life of the metric, and if it is just the getter it will filter for the ALL row multiple times rather than once when the metric is created.

macfarlandian: if it's a mobx computed getter that only depends on `this.allRecords`, it should only run once and be cached from then on

jovergaag: OK I'll give that a try!

jovergaag: GREAT suggestion! Worked like a charm, and making the `dataSeries` a computed getter as well it reduced the number of times the dataSeries was computed. Thank you!

macfarlandian: 😎  `computed`s are the mobx superpower

jovergaag: I think I knew that at one point but clearly somehow I lost it. So glad to have it back <3

macfarlandian (approved):

PR #1623, opened 2022-02-01
Default values to Unknown when a filter is enabled for that state, otherwise to ALL

### LINKED ISSUE #1619: Use `enabledFilters` per state to determine if a metric attribute should be cast to 'ALL' or to 'Unknown'
A clear and concise description of what task needs to be accomplished and why.

**Additional context**

jovergaag:
## Description of the change

We build the default to "ALL" functionality to support the development of new filters against fixture data while allowing `main` to remain deployable for states with a more limited set of dimensions enabled in the metric view. This ended up causing problems when a dimension value was undefined, it was cast to "ALL" when it was indeed just an unknown value. The result of this was incorrect counts throughout the app.

We now look at the enabled filters per state, and if a filter is enabled for that state, default the undefined value to "Unknown" , and if the filter is not enabled continue to default it to "ALL".

I have done quite a bit of QA on this and our counts for ND look much better. Most notably, the counts/rates charts now make sense (before, the rates were adding up to more than 100%). I also went back to check US_ID and the counts there have remained steady which is important since the state is already launched for some charts.

## Type of change

## Related issues

> Closes #1619

### Development

### Code review


src/core/models/utils.ts
```diff
@@ -77,13 +78,22 @@ const lengthOfStayMap = {
...
     {},
     defaults,
-    ...Object.entries(record).map(([k, v]) =>
-      v === undefined ? {} : { [k]: v }
-    )
+    ...Object.entries(record).map(([k, v]) => {
+      if (v === undefined) {
+        return enabledFilters.includes(k as EnabledFilter)
+          ? { [k]: "Unknown" }
+          : {};
```
jovergaag, line 93: This is where the actual change was made.

Catacola (commented):
Works for me. Thanks for all the solid testing.

Catacola (approved):

PR #1633, opened 2022-02-07
Various chart improvements

### LINKED ISSUE #1630: Ensure that y-axis is whole numbers (not decimals) on count charts
When we are counting people events (all of the charts except for the "rate" charts) we want the y-axis to always be whole numbers. I have found that we are seeing decimals when the numbers are very small, like this on a count over time chart.

**Additional context**

### LINKED ISSUE #1631: X-axis chart labels - rotate and wrap
Some of the chart labels on the x-axis are getting squished. Let's rotate the x-axis labels 45 degrees for the facility, district, judicial district, and race charts. We should also set a max width of those text boxes and wrap the text if needed.

Screenshots of the issue with ND data:

Example of what the rotated and word wrapped labels should look like:

**Additional context**

nasaownsky:
## Description of the change

-Rounded Y-axis ticks to whole numbers
-Rotated and wraped X-axis chart labels

## Type of change

## Related issues

Closes #1630 
Closes #1631 

### Development

### Code review


src/core/VizPopulationSnapshot/VizPopulationSnapshot.tsx
```diff
@@ -176,7 +177,17 @@ const VizPopulationSnapshot: React.FC<VizPopulationOverTimeProps> = ({
...
+              <text textAnchor="middle">
+                {isRotateLabels
+                  ? accessorLabel.split("/").map((wrapPart) => (
+                      <tspan dy="1.5em" x="0">
+                        {wrapPart}
+                      </tspan>
+                    ))
+                  : accessorLabel}
+              </text>
+            );
```
nasaownsky, line 190: It turns out it's not that easy to wrap text that's rendered as svg. So I end up with this solution, but we need some character or space, that we can use as separator to split string and wrap it. Go ahead and play with filter labels to see how it works. I think this issue needs to be discussed with the PM.

jovergaag: Why don't we make the character a vertical dagger `|`, and add one of those characters after the `/` in the filter option label. This is a character that will never be used in the labels so it will be safe to split the string on that character. That way the slash can remain in the label in the charts and we have control over where the string is wrapped.

If we do this, we will also want to use this character to text wrap the label in the filter dropdown as well which will also be nice for that dropdown formatting on these longer labels!

jovergaag: I think splitting it this way is great! I just changed this to a regex so that the `/` would remain in the label string.

jovergaag (approved):
Everything is starting to look super polished! 🎉

PR #1638, opened 2022-02-08
Refactor metrics to remove filtering by each dimension individually

### LINKED ISSUE #1596: Refactor metrics to remove filtering by each dimension individually
Let's keep a list of dimensions per metric and iterate through them so as to be less brittle in the filters on each metric. This list can also be used in the `totalCount` getter on the metrics and the grouping inside the filters.

**Additional context**

nasaownsky:
## Description of the change

-Added tooling to group filtered records

## Type of change

## Related issues

Closes #1596 

### Development

### Code review


nasaownsky:
@jovergaag So I've added this tooling I've created in last PR, and it seems to work all right! 
Would you take a look at filtering? I tried some approaches, but none has worked.

jovergaag (commented):
@nasaownsky Take a look at the commit that I pushed up. The `SupervisionPopulationSnapshotMetric` has the start of what I think will be a good implementation of the filters. I have not tested this much, and there are lots of typescript errors but it should hopefully get you headed in the right direction!

I ended up creating a new type called `Dimension` because we have some dimensions (charts: like Length of Stay) that will never be filters. This is causing some problems in the filtering logic because we're trying to use a `dimensionId` to index the filters...but see if you can make typescript happy. Note: Our filter types are not great, so feel free to overhaul them if needed.

nasaownsky:
I'm not exactly understand now how the `priorLengthOfIncarceration` and `lengthOfStay` charts should work. Earlier we had list with options in `filterOptions`, which were used to display specific values. For `priorLengthOfIncarceration` it was ALL, 0, 1 etc. They were not meant to be used as filters, instead they were used for highlighting chosen chart columns. And now that they're deleted, I'm not understand, where from to get them (not only to highlight, but and for simply displaying). @jovergaag

jovergaag:
> I'm not exactly understand now how the `priorLengthOfIncarceration` and `lengthOfStay` charts should work. Earlier we had list with options in `filterOptions`, which were used to display specific values. For `priorLengthOfIncarceration` it was ALL, 0, 1 etc. They were not meant to be used as filters, instead they were used for highlighting chosen chart columns. And now that they're deleted, I'm not understand, where from to get them (not only to highlight, but and for simply displaying).

`lengthOfStay` and `priorLengthOfIncarceration` will never be filters, they will only ever be charts. On those charts, since we don't have the corresponding filter, the "highlighting" functionality is not necessary, the bars will always be the default color. Does that make sense?

src/core/models/dimensions.ts
```diff
@@ -40,10 +40,7 @@ export const dimensionsByMetricType = {
     DIMENSION_TYPES.FACILITY,
     DIMENSION_TYPES.AGE_GROUP,
   ],
-  prison_population_projection_time_series: [
-    DIMENSION_TYPES.GENDER,
-    DIMENSION_TYPES.LEGAL_STATUS,
-  ],
+  prison_population_projection_time_series: [DIMENSION_TYPES.GENDER],
```
jovergaag, line 43: Why was LEGAL_STATUS removed here? And on line 78?

src/core/VizPopulationSnapshot/VizPopulationSnapshot.tsx
```diff
@@ -58,8 +58,9 @@ const VizPopulationSnapshot: React.FC<VizPopulationOverTimeProps> = ({
     enableMetricModeToggle,
   } = metric;
 
-  // @ts-ignore
-  const accessorFilter = filters[accessor];
+  const isNotFilters = ["priorLengthOfIncarceration", "lengthOfStay"].includes(
+    accessor
+  );
```
jovergaag, line 63: I see what you mean about this dimension/filter change causing issues here (along with all of the typescript problems it has created). I need to spend some more time looking at this to see if that change is causing more problems than its solving. But it is late today and my brain is tired. I'll take a look at this first thing in the morning.

src/core/utils/filterOptions.ts
```diff
@@ -250,20 +250,32 @@ export const DefaultPopulationFilterOptions: PopulationFilters = {
     setFilters: setFilters(FILTER_TYPES.JUDICIAL_DISTRICT),
     options: [
       { label: "All", value: "ALL" },
-      { label: "JUDICIAL_DISTRICT_1", value: "JUDICIAL_DISTRICT_1" },
-      { label: "JUDICIAL_DISTRICT_2", value: "JUDICIAL_DISTRICT_2" },
-      { label: "JUDICIAL_DISTRICT_3", value: "JUDICIAL_DISTRICT_3" },
-      { label: "JUDICIAL_DISTRICT_4", value: "JUDICIAL_DISTRICT_4" },
-      { label: "JUDICIAL_DISTRICT_5", value: "JUDICIAL_DISTRICT_5" },
+      {
+        label: "JUDICIAL_DISTRICT_1",
```
jovergaag, line 266: Let's make these labels "District 1", "District 2" etc

jovergaag (commented):
Overall these metric filters are looking SO MUCH BETTER! Thank you! I pushed up a commit which has a solution for the time period filter.

I'm going to take another look at the dimension/filter type problem that I created first thing in the morning when I have a fresh brain.

src/core/VizPopulationSnapshot/VizPopulationSnapshot.tsx
jovergaag: I'm sure that removing `lengthOfStay` as a filter is also why the Length Of Stay charts aren't working anymore. I think it is something in `filterRecords` but will have to take a closer look at it tomorrow. 

Thanks for your patience with PR reviews, things are moving quickly over here now trying to get ME and TN data loaded into the dashboard so I have not been diligent with reviews. Your PRs are my priority tomorrow!

src/core/models/PopulationProjectionOverTimeMetric.ts
```diff
@@ -42,27 +42,24 @@ export default class PopulationProjectionOverTimeMetric extends PathwaysMetric<P
...
         const monthsOut =
           (record.year - simulationDate.getFullYear()) * 12 +
           (record.month - (simulationDate.getMonth() + 1));
         return (
-          gender.includes(record.gender) &&
-          status.includes(record.legalStatus) &&
           Math.abs(monthsOut) <= monthRange &&
-          monthsOut % stepSize === 0
+          monthsOut % stepSize === 0 &&
+          status.includes(record.legalStatus) &&
```
nasaownsky, line 61: > Why was LEGAL_STATUS removed here? And on line 78?

Because in this metric the record name is the same, and filter name depends on compartment. If we use it like before from dimensions we would always get `legalStatus.includes(record.legalStatus)` which is not what we want.

src/core/VizPopulationSnapshot/VizPopulationSnapshot.tsx
nasaownsky: Yeah, at first I was also thinking, that the problem is in `filterRecords`, but then I was checkout to `main`, and it turns out, that the `lengthOfStay` charts aren't also work. That was strange, especially with working `priorLengthOfIncarceration` chart, which is practically the same. So the problem appeared earlier. In short: the problem is in `utils.ts` on lines 147 and 177. If we remove all that logic and only keep `lengthOfStay: record.length_of_stay,`, the charts will work as usual. So the problem is in `lengthOfStayMap`, because only fixture file that comtains `length_of_stay` data is `supervision_to_prison_population_snapshot_by_dimension`, and it has another values for `length_of_stay`. Rather than `months_0_3, months_3_6, months_6_9...` and so on it has `[ "length_of_stay", [ "0", "12", "18", "24", "6", "all" ] ]`.

jovergaag: Good find. I'll update the fixtures. We changed the schema a couple of weeks ago and I failed to update the fixtures to match!

jovergaag: Fixed.

src/core/models/PrisonPopulationSnapshotMetric.ts
```diff
@@ -58,12 +58,10 @@ export default class PrisonPopulationSnapshotMetric extends PathwaysMetric<Priso
...
       (record: PrisonPopulationSnapshotRecord) => {
         return (
-          // #TODO #1596 create tooling to reduce listing every dimension in filters
-          record.gender === "ALL" &&
-          record.legalStatus === "ALL" &&
-          record.facility === "ALL" &&
-          record.ageGroup === "ALL" &&
-          record.lengthOfStay === "ALL" &&
+          this.dimensions.every(
+            // @ts-ignore
```
jovergaag, line 62: I played around a little bit with creating different dimension types `PrisonDimensions` and `SupervisionDimensions` and `LibertyDimensions`. It kind of worked and was helpful...but not really. I think that this `ts-ignore` is fine for now. I will create a ticket to work out our typing in a couple of weeks when things have stabilized a bit with all of our launches.

PS - we are launching US_ME next week and US_TN a couple of weeks after! Moving fast :)

src/core/models/PrisonPopulationSnapshotMetric.ts
```diff
@@ -115,25 +98,28 @@ export default class PrisonPopulationSnapshotMetric extends PathwaysMetric<Priso
...
+          }
+        );
+        return {
+          count: sumBy("count", dataset),
+          populationProportion: (
+            (sumBy("count", dataset) * 100) /
+            this.totalCount
+          ).toFixed(),
+          ...datasetWithoutCount[0],
+        };
```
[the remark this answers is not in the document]
jovergaag: The `avg90Day` should be pulled out of the object, similar to `count` for the metrics that have the that attribute.

jovergaag:
> Because in this metric the record name is the same, and filter name depends on compartment. If we use it like before from dimensions we would always get legalStatus.includes(record.legalStatus) which is not what we want.

I see. Its not ideal but thats the way the data is laid out so we will work with it. I think its fine since the projections page is not using the new filtering function. Within the next few months US_ND will move to the "projections" charts on the prison and supervision pages, we can work on the projections metrics at that point.

jovergaag (changes requested):
The only thing left is to copy that `datasetWithoutCount` logic to the rest of the metrics. Looking good!

src/utils/datasets.js
```diff
@@ -121,6 +121,6 @@ export function sortByLabel(dataPoints, labelKey) {
     if (a[labelKey] === "All") return -1;
     if (b[labelKey] === "All") return 1;
 
-    return a[labelKey].localeCompare(b[labelKey]);
+    return a[labelKey].localeCompare(b[labelKey], "en", { numeric: true });
```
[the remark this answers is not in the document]
nasaownsky: It filters labels not only by alphabet, but in numeric-right way. So instead of `1, 12, 24, 6` we will get `1, 6, 12, 24`.

nasaownsky:
I've noticed, that we have problem with first render, when switching charts, and it came after merging later changes from main. If we change any filter in filter bar the render passes. Maybe it's `PageSystem` fix?

jovergaag:
> I've noticed, that we have problem with first render, when switching charts, and it came after merging later changes from main. If we change any filter in filter bar the render passes. Maybe it's PageSystem fix?

I don't see the problem. On first load, when switching charts, everything looks find to me. Can you give me more details about what you're seeing?

nasaownsky:
It's gone with revert. So I merged main again and now it should be good to go.

jovergaag (approved):
This is HUGE! Such a big improvement and less bug prone. Thank you!

PR #1648, opened 2022-02-10
Clean up console warnings and errors in tests from the FiltersStore

### LINKED ISSUE #1640: Clean up console warnings and errors in tests from the FiltersStore
The `reaction` that I added last week is causing tons of noise in the console and tests. Let's clean up those warnings/errors, if it means updating the disabled filters in a different way that's fine!

**Additional context**

nasaownsky:
## Description of the change

Got rid of console errors caused by reaction in `FiltersStore`. Also got rid some more console errors, which I was able to catch.

## Type of change

## Related issues

Closes #1640 

### Development

### Code review


src/core/CoreStore/FiltersStore.ts
```diff
@@ -70,10 +70,11 @@ export default class FiltersStore {
     this.getFilterLongLabel = this.getFilterLongLabel.bind(this);
 
     reaction(
-      () => this.rootStore.metricsStore.current.filters,
+      () => this.rootStore.metricsStore?.current?.filters,
       (filters: Filters) => {
         this.clearDisabledFilters(filters);
-      }
+      },
+      { delay: 300 }
```
jovergaag, line 77: I don't love this. Anytime we have to add a delay to something it sets off red flags in my head that its going to be buggy. Can we maybe put the `clearDisabledFilters` in a `useEffect` with a `filters` dependency in `PathwaysFilterBar`?

jovergaag (changes requested):
This is a great progress! I'm still seeing lots of warnings in the tests which would be nice to clean up:

`[MobX] Since strict-mode is enabled, changing (observed) observable values without using an action is not allowed. Tried to modify: CoreStore@12.metricsStore`
and
Screen Shot 2022-02-10 at 2 25 16 PM
and
Screen Shot 2022-02-10 at 2 25 22 PM


I've also got several console warnings and errors: 
- In demo mode, I see an error for Intercom. We can just check `isDemoMode` and not `initIntercomSettings` if we are in demo mode.
- I also see these in my console which would be nice to clean up.

Screen Shot 2022-02-10 at 2 17 31 PM

src/core/CoreStore/FiltersStore.ts
nasaownsky: I tried many options, and they're all overwhelmed like my previous with useState, or aren't work. This is the compromise. Besides, this delay is official mobx `reaction` option, so I think there will no problems with it.

nasaownsky:
@jovergaag I've managed to clear majority of that errors/warnings except two: the `Cannot update a component ('PageSystem')` and the keys errors.

jovergaag:
> I've managed to clear majority of that errors/warnings except two: the `Cannot update a component ('PageSystem')` and the keys errors.

These look great! The keys error is because we're generating the methodology via markdown and its impossible to insert a key into the bulleted list in the markdown. So that one is fine!

I pushed up a commit that gets rid of the `PageSystem` error. Basically, that component didn't need to be an observer, all of its children could be observers and get observables directly from the stores. There is still a test failing and it creates an infinite loop somewhere in the `PathwaysFilterBar` when changing charts or changing pages but I think that's just a matter of putting a different guard on line 49. Maybe also check for `isLoading`?

Let me know your thoughts on this!

src/App.js
```diff
@@ -59,9 +59,12 @@ import { PATHWAYS_TENANTS } from "./RootStore/TenantStore/pathwaysTenants";
 import { initI18n } from "./utils/i18nSettings";
 import initFontAwesome from "./utils/initFontAwesome";
 import initIntercomSettings from "./utils/initIntercomSettings";
+import { isDemoMode } from "./utils/isDemoMode";
 
 initFontAwesome();
-initIntercomSettings();
+if (!isDemoMode) {
+  initIntercomSettings();
+}
```
jovergaag, line 67: YAY thank you.

src/core/CoreStore/index.ts
```diff
@@ -47,7 +47,7 @@ export default class CoreStore {
 
   filtersStore: FiltersStore;
 
-  metricsStore: MetricsStore;
+  metricsStore: MetricsStore = new MetricsStore({ rootStore: this });
```
jovergaag, line 50: That's nice! Can you please also do that for the filtersStore and pagePracticesStore for consistency.

jovergaag:
> I pushed up a commit that gets rid of the `PageSystem` error.

Oops, I failed to enter my ssh password to push the commit so it never pushed. Should be here now :)

src/core/CoreStore/index.ts
nasaownsky: Seems like we need determine `PagePracticesStore` in the old way, because it's throws an error if not. Also, new logic of `PathwaysFilterBar` looks great, but could you please fix its testing? I can't get why they aren't work.

jovergaag (approved):
So much nicer!! THANK YOU

PR #1687, opened 2022-03-10
[Practices v2] base models and subscription logic

macfarlandian:
## Description of the change

This models most of the basic domain objects and database reads for the core Practices interactions of searching for clients, filtering by your own caseload or your set of saved officers, and viewing a single client in greater detail. It doesn't include 100% of the data we would be reading but it's enough to establish some basic patterns I think. Also worth noting is that it only works in the tests, since there is no Firestore integration yet. (I would probably recommend starting there to get a sense of how the data will be consumed.)

This is informed by a [mini design-doc](https://paper.dropbox.com/doc/Data-fetching-for-practices--BdgpdzefhFN4eo6SnXnDRxfFAQ-R5knS9wa6IEOkXYmCzWlH) that I wrote up in the course of writing all this, which outlines the key requirements and Firestore constraints we need to work around. At this point I am going to try letting the Client objects fetch their own additional data rather than issuing multiple search queries and trying to "join" the returned results, because it's much cleaner in application code and I believe the Firestore SDK is pretty smart about minimizing the overhead of concurrent subscriptions. Once we load-test this with realistic data volume we'll find out if this was a good idea! 

This code is obviously not production-ready but it is gated behind a call to the `hydrate` method on the main data store; nothing else will happen until that is called, and we aren't calling that, so it should be safe to merge. I have verified that I don't see any side effects when accessing the rest of the dashboard app.

## Type of change

## Related issues

https://github.com/Recidiviz/recidiviz-data/issues/11352, https://github.com/Recidiviz/recidiviz-data/issues/11372, https://github.com/Recidiviz/recidiviz-data/issues/11421, https://github.com/Recidiviz/recidiviz-data/issues/11427

### Development

- ~E2E have been run for Lantern changes ([Steps in the Readme](https://github.com/Recidiviz/pulse-dashboard#running-e2e-tests))~ n/a

### Code review


macfarlandian:
Updated this with a cleaner and more complete implementation that now includes the actual relationship between the Client and the Opportunity

macfarlandian:
I am rethinking this approach a bit actually ... the more I analyze the required functionality the more I think the subscription model is not really the right fit. There are too many situations where the same UI functionality can be applied to an arbitrarily large dataset (e.g. searching for a user, which could be applied to a single officer's caseload, multiple officers' caseloads, an entire district or the entire state) ... rather than having to decide on the fly whether or not we should subscribe based on some heuristic rule I think it would be simpler to just never do it at all!

macfarlandian:
Taking this out of draft status, I think it has stabilized enough to move on with

src/PracticesStore/ClientUpdate.ts
```diff
@@ -0,0 +1,34 @@
...
+  clientId: string;
+
+  readonly compliantReportingStatus?: Readonly<
+    ClientUpdateRecord["compliantReportingStatus"]
+  >;
+
+  constructor(record: ClientUpdateRecord) {
+    this.clientId = record.personExternalId;
+    this.compliantReportingStatus = record.compliantReportingStatus;
+  }
```
Catacola, line 33: I'm assuming this class is where we will add any future client-level changes users make in the app?

src/PracticesStore/PracticesStore.ts
```diff
@@ -0,0 +1,200 @@
...
+          });
+        } else {
+          throw new Error(
+            `Unable to retrive user record for ${authenticatedUser.email}`
+          );
+        }
+      } else {
+        // We expect the user to already be authenticated
+        throw new Error("Missing email for current user.");
+      }
```
Catacola, line 102: You don't need to take this suggestion, but I when there are multiple checks like this that can throw errors, I like to invert the outer ones to throw the error in the then-clause of the if-statement so we don't separate the logic from the error and need to balance checks with errors. For example:

```suggestion
      if (!authenticatedUser?.email) {
        // We expect the user to already be authenticated
        throw new Error("Missing email for current user.");
      }
      
      const userRecord = await getUser(authenticatedUser.email);
      if (userRecord) {
        runInAction(() => {
          this.user = userRecord;
          this.setDefaultCaseload(userRecord);
          this.updateStateData(userRecord.info.stateCode);
        });
      } else {
        throw new Error(
          `Unable to retrieve user record for ${authenticatedUser.email}`
        );
      }
```

You could also invert the `userRecord` check as well, but since it's so small of a clause, it's not the same sort of issue.

src/PracticesStore/__fixtures__/index.ts
```diff
@@ -0,0 +1,78 @@
...
+
+export const mockOfficers: UserRecord[] = [
+  {
+    name: "Foo Fakename",
+    id: "OFFICER2",
+    stateCode: mockUser.info.stateCode,
+    district: "1",
+  },
+  {
+    name: "Bar Realname",
```
Catacola, line 66: Very realname. Much trust.

src/firestore/firestore.ts
```diff
@@ -0,0 +1,67 @@
...
+  ClientUpdateRecord,
+  CombinedUserRecord,
+  OpportunityType,
+  UserRecord,
+} from "./types";
+
+export async function getUser(
+  email: string
+): Promise<CombinedUserRecord | undefined> {
+  throw new Error("Not implemented");
```
Catacola, line 30: Nice thinking

Catacola (approved):

src/PracticesStore/PracticesStore.ts
macfarlandian: yeah good call this is much cleaner

PR #1709, opened 2022-03-22
[Practices] More details in client profile

macfarlandian:
## Description of the change

Displays in the client profile all of the top-level client data we expect to have as of now (some more P1/P2 stuff might come in later but this is the basic set), including some additional fields that I added as part of this change based on what I have seen in https://github.com/Recidiviz/recidiviz-data/pull/11182. 

Screen Shot 2022-03-21 at 5 35 09 PM

## Type of change

## Related issues

> Closes https://github.com/Recidiviz/recidiviz-data/issues/11616

### Development

### Code review


src/core/PracticesClientProfile/Details.tsx
```diff
@@ -0,0 +1,117 @@
...
+  letter-spacing: -0.01em;
+  line-height: 1.14;
+  margin-bottom: ${rem(spacing.sm)};
+  margin-top: ${rem(spacing.lg)};
+`;
+
+const DetailsList = styled.dl``;
+
+const DetailsSubheading = styled.dt`
+  color: rgba(53, 83, 98, 0.5);
```
Catacola, line 43: Should we define this and the one on L52 in the palette?

Catacola (commented):
This all looks good to me. I was able to get the site to run in offline mode, but I was unable to get the client profile to load. I tried `http://localhost:3000/practices/compliantReporting/100` and `http://localhost:3000/practices/compliantReporting`, but all I got were blank pages. Selecting an officer on `http://localhost:3000/practices` lead to nothing.

macfarlandian:
yeah you have to hack in a value for `PracticesStore.selectedOfficers` for now until the selection UI is merged in (the existing select doesn't do anything it's just for show) ... may I suggest `["OFFICER1"]`

src/core/PracticesClientProfile/Details.tsx
macfarlandian: these aren't even named styles in Figma as yet so I am not sure what their status is re: the design system ... would be a question for @agronomic when he gets back maybe

macfarlandian:
I was thinking about this actually and realized it could happen in the wild also, if a user is linked directly to this URL but doesn't have a default caseload. I will try to solve for that here!

macfarlandian:
OK this page will now attempt to fetch the specified client's data if we don't already have it. @ohaibbq this decouples the client list from the subscription so it should also help somewhat with the list behavior by exposing fewer intermediate subscription states.

ohaibbq (approved):
🦮

PR #1760, opened 2022-03-30
Adjust authorize logic in UserStore to fix timing issue

jovergaag:
## Description of the change

Fixes bug in Lantern dashboard where `revocations_matrix_supervision_location_ids_to_names` was being fetched before the token was received. 

## Type of change

## Related issues

> Closes [#XXXX]

### Development

### Code review


src/RootStore/UserStore.ts
```diff
@@ -147,15 +147,15 @@ export default class UserStore {
       }
       if (await auth0.isAuthenticated()) {
         const user = await auth0.getUser();
-        await runInAction(async () => {
-          this.userIsLoading = false;
+        await authenticate(await auth0.getTokenSilently());
```
macfarlandian, line 150: we still should not call this unless the user's email is verified

macfarlandian (changes requested):

src/RootStore/UserStore.ts
jovergaag: Yes, good point. Fixed.

ohaibbq (approved):

src/RootStore/UserStore.ts
```diff
@@ -147,19 +147,21 @@ export default class UserStore {
...
             this.getToken = (options?: GetTokenSilentlyOptions) =>
               this.auth0?.getTokenSilently(options);
             this.logout = (...p: any) => this.auth0?.logout(...p);
             this.isAuthorized = true;
-          } else {
+            this.userIsLoading = false;
+          });
+        } else {
+          runInAction(() => {
             this.isAuthorized = false;
```
macfarlandian, line 162: need to set `this.userIsLoading = false` in this branch also?

jovergaag: No - the user will not have loaded if it doesn't get authorized.

jovergaag: `this.user = user;` is in the branch above so the user will not be loaded in the else branch.

macfarlandian: OK, just making sure: before this change, that property was not set conditionally. Maybe that was part of the problem?

jovergaag: Oh I see what you're saying. I need a break, I haven't stood up or eaten anything yet today so my mind is stressed. I'll be back in a few to look at this with fresh eyes.

macfarlandian: I hear that!

jovergaag: After looking at this for a while, I think that setting `userIsLoading` prior to checking that the user was verified or authorized was a bug. The `DistrictsStore` and `BaseDataStore` (among others) are checking `userIsLoading` before fetching data. If the user is not verified, or not authorized, it should never fetch. I believe that this current configuration is correct, and I have adjusted a couple of tests to cover this.

macfarlandian (approved):
I think this makes sense for now, but I also think our general confusion on this matter is an argument for moving toward checking `user.isAuthorized` in the future, when that is semantically what we care about

PR #1771, opened 2022-04-05
[Workflows] Add flag to app metadata ot prevent logging in workflows

Catacola:
## Description of the change

Adds the optional flag `workflows_no_not_log` to app metadata. If this field is present and set to true, this will suppress most logging of the user's interactions. Specifically, it will short-circuit all calls to `track()` in `analytics.ts` and it will prevent logging a form as completed if printed.

This IS NOT a read-only mode. If a user edits a form or changes the eligible/deny reason, it will get stored to firestore.

## Type of change

## Related issues

> Closes [#XXXX]

### Development

- ~[ ] E2E have been run for Lantern changes ([Steps in the Readme](https://github.com/Recidiviz/pulse-dashboard#running-e2e-tests))~ not lantern related

### Code review


src/PracticesStore/__tests__/Client.test.ts
```diff
@@ -155,6 +157,15 @@ test("mark client as completed when printing form", () => {
   );
 });
 
+test("don't record a completion if user has userAppMetadata.workflows_do_not_log = true", async () => {
+  // @ts-ignore
```
macfarlandian, line 161: what are we ignoring here? this directive is generally frowned upon, not clear to me why it's needed

macfarlandian (changes requested):
what is the actual use case we are addressing here? I guess I am not sure why a somewhat unpredictable "write sometimes but don't log" mode is preferable to a true "read only" mode

Catacola:
@macfarlandian It's now a true read-only mode.

src/RootStore/UserStore.ts
```diff
@@ -36,6 +36,10 @@ import { authenticate } from "../firestore";
 import tenants from "../tenants";
 import { isOfflineMode } from "../utils/isOfflineMode";
 import { getAllowedMethodology } from "../utils/navigation";
+import {
+  setWorkflowsCanWrite,
```
macfarlandian, line 40: ESLint is yelling about this

src/__tests__/analytics.test.ts
```diff
@@ -0,0 +1,68 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ * =============================================================================
+ *
+ */
+
+import { trackReferralFormPrinted } from "../analytics";
+import mockWorkflowsReadOnly from "../utils/test/mockWorkflowsReadOnly";
+
+const trackMock = jest.spyOn(window.analytics, "track");
```
macfarlandian, line 24: nice

src/firestore/__tests__/firestore.test.ts
```diff
@@ -0,0 +1,59 @@
...
+ * You should have received a copy of the GNU General Public License
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ * =============================================================================
+ *
+ */
+
+import mockWorkflowsReadOnly from "../../utils/test/mockWorkflowsReadOnly";
+import { maybeSetDoc } from "../firestore";
+
+// const mockSetDoc = jest.spyOn(firebase.firestore, "setDoc");
```
macfarlandian, line 24: delete?

src/utils/workflowsCanWrite.ts
```diff
@@ -0,0 +1,37 @@
...
+ * along with this program.  If not, see <https://www.gnu.org/licenses/>.
+ * =============================================================================
+ *
+ */
+
+import UserStore from "../RootStore/UserStore";
+
+let userCanWrite = true;
+
+export const setWorkflowsCanWriteFromUserStore = (userStore: UserStore) => {
```
macfarlandian, line 25: got a bunch of lint warnings in this file

macfarlandian, line 25: i think it would be "safer" if this just accepted a flag (or maybe an app metadata object) rather than the whole datastore, just to limit the potential for weird mobx-related bugs or other nastiness in the future

macfarlandian (changes requested):
this is pretty nice ... I do think that for a belt to go with these suspenders, we should also check for the read_only flag in the Firestore rules. We are already stuffing some app metadata into the Firebase token when we construct it on the server, so it would be pretty trivial to add that field I think

PR #1802, opened 2022-04-11
[Workflows] add parole special conditions

macfarlandian:
## Description of the change

Incorporates a new separate field for "board conditions" aka special conditions of parole. They contain different information than the ones associated with supervision sentences but thankfully are already stored as valid JSON!

Screen Shot 2022-04-11 at 3 16 24 PM

## Type of change

## Related issues

https://github.com/Recidiviz/recidiviz-data/issues/11959

### Development

### Code review


src/core/PracticesClientProfile/Details.tsx
```diff
@@ -141,9 +141,30 @@ function getSpecialConditionsMarkup(client: Client): JSX.Element {
 export const Details = observer(({ client }: ClientProfileProps) => {
   return (
     <DetailsSection>
-      <DetailsHeading>Special Conditions</DetailsHeading>
+      <DetailsHeading>Probation Special Conditions</DetailsHeading>
```
danawillow, line 144: Does this mean there will always be sections for both probation and parole special conditions? If so would it be possible to only display the header for the appropriate one? I could see it being confusing to see both headers even if only one section has data.

macfarlandian: good question ... this is what @sychang requested, unless I misinterpreted. I am not sure whether we know for sure which type of SCs could be expected for a given client? If we can infer that from some other piece of data (e.g. the supervision type) then yeah it would be easy to display the fields conditionally

Catacola: Serena is in the sky for the next few hours, so we should just make a call here. Let's go with both, or somewhat cleaner, can we only display them if they have values?

Catacola (approved):

src/core/PracticesClientProfile/Details.tsx
macfarlandian: I think the latter is probably not desirable because users want to know affirmatively that there are no special conditions for a client, and that could become ambiguous if the field is simply missing

macfarlandian: we can also just wait on this, she did [explicitly say](https://docs.google.com/document/d/1UihGqjQRWZIAPjVUHdfKs5IMRL3vQTNWyN7_lq6FKZI/edit?disco=AAAAXaq45wE) this is not blocking for tomorrow

danawillow: Thanks for the explanation! I'm fine with merging or waiting, with a weak preference for merging since this displays information we weren't displaying before.

danawillow (approved):

Issue #1720, opened 2022-03-23
Length of stay charts - redesign


jovergaag:
**Is your feature request related to a problem? Please describe.**
On any of the survival curve chargs (Length of stay), when a filter is selected, break each dimension value into its own line on the chart and highlight the selected line(s)

**Describe the solution you'd like**

**Describe alternatives you've considered**

**Additional context**
Add any other context about the feature request here.

jovergaag:
@agronomic @sychang I did some looking at this last week and it gets to be really complicated, really quickly. It works if only a single filter is selected, but what if we select the gender filter AND the race filter? Do we show a line for every permutation of those options? Or if there is more than one filter selected, show only the single "total" line? and not the dimension (multi-line) breakout?

sychang:
Going to reassign this one to @niclui to do a bit of scoping, since he's been thinking about how to surface comparisons in a useful way.

niclui:
Hi @jovergaag @sychang, here's the PRD [(doc)](https://docs.google.com/document/d/1TYNWMp_tj2F2iXoDgpVqgMSchXVaG7LOW-1x_e2By1w/edit?usp=sharing) with lo-fi wireframe [(slides)](https://docs.google.com/presentation/d/1dn5qh4CNjkxbfYDePrdXgqxGgXR6aLE1d9lOSCNqJRU/edit?usp=sharing). Feel free to schedule a time on my calendar if you'd like to chat through it

sychang:
I think this looks great! Going to take a small amount of design work so reassigning to @soniachokshi for now

PR #1823, opened 2022-04-19
[Pathways] Length of stay charts - update denominator

jovergaag:
## Description of the change

Updates the denominator in the length of stay charts to be the sum of the cohort - instead of the total count of the compartment population. Also adds a 0 point at 0 months to ensure the line is tied to the intersection of the x and y axis.

Adds the LOS charts to all enabled states that do not have it enabled.

Awaiting design and QA sign off.

## Type of change

## Related issues

> Part of #1720

### Development

### Code review


src/core/VizLengthOfStay/VizLengthOfStay.tsx
```diff
@@ -50,6 +47,7 @@ const VizLengthOfStay: React.FC<VizLengthOfStayProps> = ({ metric }) => {
       cohortProportion: (accumulatedCount * 100) / totalCount,
     };
   });
+  data.unshift({ lengthOfStay: 0, count: 0, cohortProportion: 0 });
```
danawillow, line 50: If the data already has data points with `lengthOfStay: 0`, what would happen here? Or is that impossible?

src/core/VizLengthOfStay/VizLengthOfStay.tsx
```diff
@@ -31,16 +31,13 @@ type VizLengthOfStayProps = {
...
-    chartTitle,
-    chartXAxisTitle,
-    chartYAxisTitle,
-    totalCount,
-  } = metric;
+  const { dataSeries, chartTitle, chartXAxisTitle, chartYAxisTitle } = metric;
 
   const latestUpdate = formatDate(dataSeries[0]?.lastUpdated, "MMMM dd, yyyy");
 
+  const totalCount = dataSeries.reduce((accumulator, d) => {
```
danawillow, line 38: I'm not 100% clear yet on what the other way of counting was doing, can you explain the difference to me? Also is it possible that other charts that use the same base data would want these counts or are they very specific to length of stay?

src/core/VizLengthOfStay/VizLengthOfStay.tsx
jovergaag: It is possible but not likely. It would just show a flat line from the (0,0) point to the (for example) (months_0_6, 0) point.

src/core/VizLengthOfStay/VizLengthOfStay.tsx
jovergaag: The original `totalCount` was the count from the 'ALL' row - which is the full count/population from the entire compartment.

The count that we are using here now for the length of stay charts is the total count for the population that is currently being represented in the chart for the given filters set. We want this total specifically for length of stay so that the line chart ends up at 100% (or close to it) at the right side of the chart, to show the growth from 0 to 100%. 

We might(?) want this cohort total in some of the other demographic charts eventually once we implement the proportional horizontal bar charts for race/gender/etc. But for right now this is the only chart where we want the cohort total.

danawillow: That makes sense, thank you!

danawillow (approved):

PR #1852, opened 2022-04-27
enroll in sentry performance monitoring

danawillow:
## Description of the change

Enrolls us in Sentry performance monitoring so we can get a baseline for how long Pathways requests are taking, and hopefully see improvements with the new backend.

This failed quietly and mysteriously until I upgraded Sentry, so it comes along with the upgrade.

It's working now, but only partially. If I refresh a page, it'll get logged in Sentry, but if I just click on the header it doesn't. This is obviously not what we want, but I think it requires more React knowledge than I have to get it working. Any ideas?

I followed the guides at:
* https://docs.sentry.io/platforms/javascript/guides/react/performance/
* https://docs.sentry.io/platforms/javascript/guides/react/configuration/integrations/react-router/ (this one it seemed like you only needed to do more than what I've done here if you want parameterized transaction names, but it's clearly not working so maybe not)

The practices change is unrelated, but for some reason the linter has decided now that it's a problem and didn't before 🤷‍♀️ 

## Type of change

## Related issues

Closes https://github.com/Recidiviz/recidiviz-data/issues/12446

### Development

~- [ ] E2E have been run for Lantern changes ([Steps in the Readme](https://github.com/Recidiviz/pulse-dashboard#running-e2e-tests))~

### Code review


src/App.js
```diff
@@ -89,7 +99,7 @@ const App = () => (
   }>
     <StoreProvider>
       <SentryErrorBoundary>
-        <Router>
+        <Router history={history}>
```
jovergaag, line 102: I've been clicking around the app on my local branch and i've only seen one of the paths that I've navigated to populate. I'm going to wait to check in on this tomorrow - maybe it takes a while to show up in Sentry, but I'm also not seeing the Sentry network request. I don't think this is updating every time the path changes.

This might be because we aren't updating the history every time we navigate to the new chart path?

danawillow: Ok, here's where I'm at: for some reason using Router instead of BrowserRouter was working this morning, but the tests were failing. This afternoon, using Router instead of BrowserRouter at all is giving me the following error:
```
helpers.ts:121 Uncaught TypeError: Cannot read properties of undefined (reading 'split')
    at AuthWall (AuthWall.tsx:35:1)
```

The line of code that's erroring is:
```
  const { pathname, search } = useLocation();
  const currentView = pathname.split("/")[1];
```

so something is going wrong with the call to `useLocation()`, which is pretty clearly relevant since that's a history-related call, but I'm not sure why it's returning undefined. Everything I can google says it returns undefined if the call happens outside of the `<Router>` component, but `<AuthWall>` is inside of `<Router>`. I think this is as far as I'm going to be able to get today.

jovergaag: Do you need to check if `pathname` exists first before trying to split it? What about 

```
pathname && pathname.split("/")[1];
```

danawillow: Got some help from @Catacola and we figured out that we could go back to using `BrowserRouter` but rework things a bit so that we call `Sentry.init` from inside a React component, which gives us access to the `useHistory()` hook instead of creating and passing our own history to the `Router`.

This seems to work- when I update Sentry to allow localhost, I'm seeing a whole bunch of events show up in the performance tab (I turned localhost back off so if you want to pull and test you'll have to turn it back on). It is initializing Sentry twice though- as far as I can tell that's fine, but heads up in case there's an easy way to fix it.

jovergaag (approved):
Works beautifully! So excited about this :)

src/App.js
```diff
@@ -70,10 +69,18 @@ if (!isDemoMode()) {
...
+// Sentry.init({
+//   environment: process.env.REACT_APP_SENTRY_ENV,
+//   dsn: process.env.REACT_APP_SENTRY_DSN,
+//   integrations: [
+//     new BrowserTracing({
+//       routingInstrumentation: Sentry.reactRouterV5Instrumentation(history),
+//     }),
+//   ],
+//   tracesSampleRate: 1,
+// });
```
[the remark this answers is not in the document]
danawillow: whoops thank you!

PR #1853, opened 2022-04-27
[Pathways] Support UNKNOWN dimension values

### LINKED ISSUE #1850: [Pathways] Prepare the FE to support "Unknown" dimension values
This includes adding "Unknown" to the filter options for some states and some dimensions, and handling the Unknown length_of_stay values.

**Additional context**

jovergaag:
## Description of the change

We will soon stop filtering out UNKNOWN dimension values in the data platform because we want to surface those values in the FE. This handles the UNKnOWN values, by adding it to the filter options where it shows up, as well as adding additional handling for UNKNOWN length_of_stay values.

Should be deployed in coordination with https://github.com/Recidiviz/recidiviz-data/pull/12486

I have exported data for all Pathways states from a sandbox from the above branch - everything looks good but feel free to do something similar during your QA.

## Type of change

## Related issues

> Closes #1850

### Development

### Code review


src/core/utils/filterOptions.ts
```diff
@@ -402,6 +402,50 @@ export const DefaultPopulationFilterOptions: PopulationFilters = {
 
 export const IdPopulationFilterOptions: PopulationFilters = {
   ...DefaultPopulationFilterOptions,
+  [FILTER_TYPES.AGE_GROUP]: {
```
danawillow, line 405: How come these are added to the state-specific filters instead of to the default filters?

jovergaag: Thats a great question. I had thought about adding it to default filters - but for most states and most dimensions - there aren't any unknowns that show up in the data, so we would have a filter option that always results in a blank chart. Maybe that's ok? I just added it where it is needed with the intention that we will one day be building the filter options dynamically based on the response from an endpoint. But we're not there yet. Maybe we should be safe and just include it in every filter for every state.

@sychang Thoughts?

sychang: Definitely defer on what's the best technical implementation. My two cents: if most states and most dimensions won't have an unknown, we'd want to avoid having several extraneous blank charts. In that sense, agree with the current implementation of adding where it's needed since the eventual vision is to filter options dynamically based on response from an endpoint.

jovergaag: @danawillow Given this input from Serena, I think that we're ready for another review here!

danawillow (approved):
Sorry for the delay, looks good!

PR #1951, opened 2022-06-08
[Pathways] make a more generic Differ, enable it for LibertyPopulationSnapshotMetric

danawillow:
## Description of the change

* Moved the diffing logic into `PathwaysMetric` 
* Created a base `Differ` class with two subclasses `TimeSeriesDiffer` and `SnapshotDiffer` (eventually we'll need to add `PersonLevel` and `PopulationTimeSeries`)
* Enabled diffs for all LibertyPopulationSnapshotMetrics

I didn't end up figuring out how to get the tests to have different mocks for "with diffs" and "without diffs", but since it's testing that the right API endpoint gets called and there's unit tests for the differs, this is probably good enough for code that's intended to be temporary anyway.

## Type of change

## Related issues

#1834

### Development

~~- [ ] E2E have been run for Lantern changes ([Steps in the Readme](https://github.com/Recidiviz/pulse-dashboard#running-e2e-tests))~~

### Code review


src/core/models/backendDiff/__tests__/SnapshotDiffer.test.ts
```diff
@@ -0,0 +1,136 @@
...
+      count: 20,
+    },
+    {
+      ageGroup: "25-29",
+      count: 10,
+    },
+  ];
+
+  expect(differ.diff(oldData, newData).entries).toHaveLength(0);
+});
```
macfarlandian, line 136: sorry if this is a dumb question but why wouldn't you want this case to be caught? are the extra fields definitely irrelevant in all cases?

src/core/models/types.ts
```diff
@@ -220,11 +220,32 @@ export type LibertyPopulationTimeSeriesRecord = {
...
 
-export interface TimeSeriesDataRecord {
+export type TimeSeriesDataRecord = {
   count: number;
   year: number;
   month: number;
-}
+};
+
+/* Superset of all the SnapshotRecords, but with all fields except "count" optional */
```
macfarlandian, line 229: could you not get this by running existing record types through `Partial<T>`?

Catacola: I believe it would need to be `Partial<PrisonPopulationSnapshotRecord & SupervisionPopulationSnapshotRecord & LibertyPopulationSnapshotRecord> & {count: number}`

src/core/models/backendDiff/__tests__/SnapshotDiffer.test.ts
danawillow: not dumb! the old backend always returns all fields and the new backend always returns only the requested fields, so yes. Added some comments to clarify.

src/core/models/types.ts
danawillow: ahh thank you, I was trying to do it with Partial before but chose a much more convoluted method and decided it looked worse, this is way better. Done!

PR #1954, opened 2022-06-09
[Pathways] guard for case when filter option not found

### LINKED ISSUE #1899: [Pathways] Clear filters when changing tenants
I think we already do this but we still are seeing errors so it must have a bug.

**Additional context**

jovergaag:
## Description of the change

We had a bug where the page would crash when changing tenants. I thought this was because we were not clearing the filters when changing tenants, but it turns out we already were. 

It is instead when changing from any page (despite the tenant) that has an enabled filter set to a page that does not have that enabled filter so the filter option cannot be found. This just guards for that case before accessing the value in `PathwaysFilterBar`

## Type of change

## Related issues

> Closes #1899

### Development

### Code review


danawillow (approved):
Good catch!

PR #1967, opened 2022-06-16
add person-level differ, use it for PrisonToSupervisionTransitionsPersonLevel

danawillow:
## Description of the change

The bulk of this change is the Person level differ, which compares the full record for each combination of state id, time_period, age_group.

Some other changes to support this:
* Removing groupBy as something you need to set in order to call the new BE
* Adding `backtrackSinceToFirstOfMonth` since for the over time charts we want to have our "since" query include everything for a given month, but not for the snapshot or person-level ones
* Having the number of diffs sent to Sentry be configurable. Person-level diffs should set it to 0 outside of testing to avoid sending PII to Sentry. I thought about sanitizing them but realistically we need to know who the diff is on, and anything that can identify who the diff is on is PII, so probably easier just to send the counts and debug locally.
* Enable the new backend for `PrisonToSupervisionTransitionsPersonLevel`

Applying the DO_NOT_MERGE label to remember to wait until we have `race` populated and available, and also until the changes from https://github.com/Recidiviz/recidiviz-data/pull/13551 are live in staging.

One issue I'm having- because we're using `since`, there are going to be periods of time each day where we see some guaranteed diffs. Easier to explain by example: today is June 16, but the latest export is from June 13 (this is true for staging right now). Because we calculate the time period for the old data at export time but the new data at query time, individuals with a transition on Dec 14 are present in the old data, but not in the new data. This is actually going to be a diff scenario for all the non-over-time metrics (those aren't affected because of the backtracking to the first of the month, which matches the export).

## Type of change

## Related issues

#1834

### Development

~~- [ ] E2E have been run for Lantern changes ([Steps in the Readme](https://github.com/Recidiviz/pulse-dashboard#running-e2e-tests))~~

### Code review


danawillow:
The dependencies for this are all resolved now.

PR #1971, opened 2022-06-21
[Workflows] identify and exclude almost eligible clients

### LINKED ISSUE #1965: [Workflows] Update CR-eligibility filtering logic
A data platform change is in progress to represent eligibility as a composite of two fields instead of just one; in addition to the existing `compliant_reporting_eligible` field (which we recast as `eligibilityCategory` during ETL) there will be a second field indicating whether there are criteria for which the client is almost eligible rather than fully eligible. This would be a breaking change for existing use cases as there is no functionality in place to filter out the new "almost" cases.

To unblock the platform change we should update the filtering in the Workflows data models to consider this new field, when it is present, and include it in the filtering logic. This change should support both the existing data format and the new one so that the data platform release requires no further coordination.

**Additional context**
[Related platform PR](https://github.com/Recidiviz/recidiviz-data/pull/13534#discussion_r898755986)

macfarlandian:
## Description of the change

Uses an expected new field to ensure almost-eligible clients are filtered out of the existing list of eligible clients. Part of making sure https://github.com/Recidiviz/recidiviz-data/pull/13534 is backwards-compatible and can be released without further coordination. We expect this field will also be useful for ungating different levels of almost-ness during rollout and beta testing.

No real data was available to test with yet but I did verify behavior by editing some live Firestore documents in staging.

## Type of change

## Related issues

> Closes #1965 

### Development

### Code review


jovergaag (approved):
Looks good to me!

PR #1975, opened 2022-06-23
[Workflows] Add feature flags managed by Firestore

### LINKED ISSUE #1941: Enable per-user rollout of almost-eligible features
We need to control the rollout of the features in this Epic so that they are accessible to a small group of users for a test period before going into wide release. (We may also want to do a phased rollout of the wide release, though that is TBD.)

There are various plausible ways of accomplishing this: 
- create and manually curate a table of flags in BigQuery that gate the export of almost-eligible data by supervising officer
- manually add the flag to user metadata in Auth0 and use that to turn the feature off on the frontend
- use the `feature-variants` feature from Case Triage to manually curate flags in Cloud Storage, fetch them from the new consolidated backend, and turn the feature off on the frontend
- build a whole new thing for this that we control via the admin panel (may also just be a new iteration of `feature-variants`)

The first option has the benefit of suppressing data that won't actually be used (reducing noise for analysis and limiting pointless bytes downloaded), but ties the feature to the data platform release cycle. The other manual options are equally cumbersome in their own ways (lots of manual JSON editing) but can be updated more freely. Building or revamping a fuller feature-flag system would be the most eng work but could let us add convenience features such as "turn on for everyone" (which requires a code change in any of the current or proposed allowlist-based solutions to remove all of the gating code entirely) or being more flexible about where we apply the gate (e.g. in BQ via a CloudSQL connection or in the application)

**Additional context**
[PRD section](https://docs.google.com/document/d/14bca--FPrZO8HOR8owzvRTS0RMIxXtS9yBTBBORYctc/edit?disco=AAAAaAlMk3k)

macfarlandian:
## Description of the change

Introduces a new read-only Firestore Collection that will contain feature flags indexed by user, plus the logic for consuming and exposing this data in the `PracticesStore` so it can be used to gate UI features for the upcoming "almost eligible" release.

Upon further investigation and discussion with @danawillow, given the current state of the new backend and the intended direction, there didn't actually seem to be much tangible benefit to trying to reuse the Case Triage version of this feature on a temporary basis; it's not that great of an incremental step toward the desired solution, would require significant refactoring, and is pretty cumbersome to actually use. Using Firestore as a temporary solution is much more lightweight and comes with the benefit of letting us manage the flags via the Firebase Console. 

For consistency I did maintain the same feature set as in Case Triage, even though I'm not sure if we'll actually need to use all of it: the values are not simple flags but objects that can optionally contain an active date and a variant string for segmentation within the feature. The logic for determining whether the feature is active as of the current moment is encapsulated within the `PracticesStore`.

No real feature is using this yet but I did test it against Firestore with a dummy component to verify the behavior.

## Type of change

## Related issues

> Closes #1941 

### Development

### Code review


danawillow (approved):
Seems reasonable! How are you planning on getting the data into Firestore?

jovergaag (approved):

macfarlandian:
@danawillow we would just be creating and editing it manually, there isn't really any data source we could use for ETL in any useful way that I can think of

PR #2018, opened 2022-07-07
[Workflows] Use sandbox collection for almost-eligible data

macfarlandian:
## Description of the change

Uses the new feature flag to redirect the application to an alternate data source for caseload records. This is accomplished by basically having the `firestore` module monkeypatch itself after receiving data from `featureVariants` and then forcing new caseload data to be fetched. It should react immediately to changes made in the Firestore console to add or remove the flag.

The main goal here was to unblock development against staging data while https://github.com/Recidiviz/recidiviz-data/pull/13534 is in flight, which will probably be for another couple of weeks. We can test schema changes and the like in a sandbox without interfering with or being clobbered by the regularly scheduled ETL runs, or having to generate fixture data, which could be fairly tedious considering all of the new fields and combinations thereof that will be introduced by this feature. 

I have populated the new sandbox collection with an export from a sandbox view based on that branch, and somewhat by chance discovered a bug where we incorrectly show the CR module on the profiles of almost-eligible clients! I will address that in a separate PR but in the meantime it is a useful way to validate that this feature works; you can visit, e.g., [this client in the preview app](https://recidiviz-dashboard-stag-e1108--sandbox-1l0ec2aq.web.app/workflows/client/4eIxZcjHN_eegNGo) and then add/remove the `CompliantReportingAlmostEligible` flag for yourself in [the Firestore console](https://console.firebase.google.com/project/recidiviz-dashboard-staging/firestore/data/~2FfeatureVariants); you should see the opportunity module appear and disappear as you make the change.

## Type of change

## Related issues

> https://github.com/Recidiviz/recidiviz-data/issues/12476

### Development

### Code review


src/firestore/firestore.ts
```diff
@@ -144,6 +144,17 @@ const collections = {
   ) as CollectionReference<FeatureVariantRecord>,
 };
 
+/**
+ * Replaces the clients collection reference in place with a reference
+ * to the clients sandbox collection
+ */
+function patchClientDataSandbox(useSandbox: boolean) {
```
Catacola, line 151: I'm wondering if the `useSandbox` argument is necessary. That is, should we just hardcode `featureVariants?.CompliantReportingAlmostEligible !== undefined` inside this function? I'm asking this from a maintenance perspective. If we ever update the eligibility reason for the sandbox, we'd have to make sure to update all of the call-sites for this function. Is there a case where you would want to patch the client collection in when calling in one place but not patch it when calling from another?

Catacola (commented):
Nice changes. Very easy to follow. I have one minor comment but otherwise this looks good to me.

Catacola (approved):

PR #2036, opened 2022-07-08
[Workflows] Refactor opportunity eligibility logic

macfarlandian:
## Description of the change

Addresses an issue where we were not properly checking the new `remainingCriteriaNeeded` field before showing the Compliant Reporting module on the client profile page (as discussed in #2018), and also refactors opportunity logic to be more generic overall (in prep for adding more opportunity types in the future).

More logic is pushed down to the Client class, which is now responsible for determining which opportunities it is eligible for. The datastore filters on this property to produce lists of eligible people. In addition, mappings over `OpportunityType` replace public opportunity-specific methods and components consume these mappings more generically.

## Type of change

## Related issues

> #1965, https://github.com/Recidiviz/recidiviz-data/issues/12476 

### Development

### Code review


src/PracticesStore/Client.ts
```diff
@@ -426,7 +441,7 @@ export class Client {
 
   printCompliantReportingReferralForm(): void {
     if (this.currentUserEmail) {
-      if (this.eligibilityStatus.compliantReporting) {
+      if (this.reviewStatus.compliantReporting !== "DENIED") {
```
Catacola, line 444: `"DENIED"` was only used in 5 places before this PR, but it's now feeling like we should turn the set of form/opportunity states into enums instead of magic strings. This includes `"PENDING"`, `"IN_PROGRESS"`, and `"COMPLETED"`.

Catacola (commented):
Looks really nice. I can see how this will be easier to build upon

src/PracticesStore/Client.ts
macfarlandian: the values are enforced at the type level, do you think that is not enough?

PR #2069, opened 2022-07-22
[Workflows] Creates new Opportunity interface and class

macfarlandian:
## Description of the change

This refactors a lot of disparate opportunity-related logic into a `CompliantReportingOpportunity` class, which implements a generic `Opportunity` interface that exposes what the UI needs in order to display the opportunity lists and modules. There shouldn't be any observable change in behavior as a result of this despite all the churn in source code. The desired end result is that opportunity-specific code is collocated as much as possible and delegated to data models, so that UI components can be extremely thin and it will be hopefully more straightforward to add new opportunities. If any of the new organization seems unintuitive or otherwise difficult to follow, that is the entire point of this change so please flag it!

We will probably want to do something similar with the opportunity-specific form data, but this is large enough as it is and that wasn't in the direct path of the almost-eligible feature so I have left all that stuff alone.

(The MobX update was a minor version bump to bring in [this bugfix](https://github.com/mobxjs/mobx/pull/2873), which I quite vexingly encountered in the course of this work! Debugging that was a major contributor to this taking two full days longer than I expected it to 😢)

## Type of change

- [x] Refactor

## Related issues

n/a

### Development

### Code review


src/PracticesStore/__tests__/PracticesStore.test.ts
```diff
@@ -588,6 +586,11 @@ test("filter out clients who are almost eligible", async () => {
         ...eligibilityFields,
         eligibilityCategory: "c2",
         remainingCriteriaNeeded: 3,
+        almostEligibleCriteria: {
```
daschi, line 589: Question about the data...are we going to later display clients that have >1 criteria needed?

src/RootStore/index.ts
```diff
@@ -62,6 +62,12 @@ if (process.env.NODE_ENV !== "test") {
     observableRequiresReaction: false,
   });
 
+  if (process.env.NODE_ENV !== "production") {
+    configure({
+      disableErrorBoundaries: true,
```
daschi, line 67: what is the need for this in, is it because of the state of the current data?

src/core/PagePracticesV2/PagePracticesV2.tsx
```diff
@@ -82,7 +82,7 @@ const PagePracticesV2: React.FC = () => {
             sidebarContents={
               <>
                 <SidebarHeading>Compliant Reporting</SidebarHeading>
-                <OpportunityCaseloadSelection opportunity="compliantReporting" />
+                <OpportunityCaseloadSelection opportunityType="compliantReporting" />
```
daschi, line 85: do you see this eventually coming from the tenant store?

daschi (approved):
I think this makes sense to me...just had a few questions.

src/PracticesStore/Opportunity/CompliantReportingOpportunity.ts
```diff
@@ -0,0 +1,592 @@
...
+    if (e instanceof OpportunityValidationError) {
+      return undefined;
+    }
+    // don't handle anything unexpected, it's probably a bug!
+    throw e;
+  }
+}
+
+// type-only export to force instantiation through creator function
+//  export type { CompliantReportingOpportunity };
```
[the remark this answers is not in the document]
macfarlandian: oh yes thanks, meant to delete that

src/PracticesStore/__tests__/PracticesStore.test.ts
macfarlandian: yes that's the expectation ... the original spec was to only show clients missing one and then gradually roll out more. Not 100% sure if that is still the release plan but we can be flexible enough that it doesn't really matter

src/RootStore/index.ts
macfarlandian: no it's just to aid with debugging in test and dev environments. without it Mobx swallows all the errors which generally just leads to mysterious UI behavior without a useful error message. It should only apply outside of deployment builds though

src/core/PagePracticesV2/PagePracticesV2.tsx
macfarlandian: probably from the `PracticesStore` ... I think once we add multi-opportunity support this will be factored out to a URL parameter, which we can then [pipe into mobx](https://github.com/Recidiviz/pulse-dashboard/blob/4eb67fa5a0f8037fc124ffc9492142e29e4c1a2a/src/core/PracticesRoute/PracticesRoute.tsx) via `PracticesRoute` (more things that need to be renamed to "Workflows"!)

daschi: :) okay sounds good...I am planning to do that practices -> workflows swich this week, prob after this PR merges

Issue #2087, opened 2022-07-26
[Workflows] Update tenant object to specify workflow opportunities by state


Catacola:
The current tenant objects have a `workflows` key to enable workflows, but the value is currently set to `[]` and does nothing. This PR is to update our config parser to use the value in that key to determine which opportunity is enabled in a state.

This value should be used to populate the `this.opportunities` variable in the constructor of `PracticesStore`.

PR #2093, opened 2022-07-26
[Workflows] Enable Workflows for ND and add basic support for multiple opportunities

### LINKED ISSUE #2092: [Workflows] Enable Workflows for ND with basic Early Termination support
A clear and concise description of what the problem is, e.g. we don't currently convey...

**Additional context**

jovergaag:
## Description of the change

- Enables workflows for ND
- Updates the workflows opportunity route and redirects `/workflows` to `/workflows/compliantReporting` for TN users, and `/workflows` to `/workflows/clients` for ND users
- Adds the opportunity types to the tenant navigation object's `workflows` key
- Only allows access to the opportunity listed in the tenant's navigation object `workflows` key (TN cannot see Early Termination and ND cannot see Compliant Reporting)
- Routes to appropriate opportunity page per tenant

## Type of change

## Related issues

> Closes #2092 and #2087 and #2096

### Development

### Code review


src/PracticesStore/PracticesStore.ts
```diff
@@ -84,6 +85,8 @@ export class PracticesStore implements Hydratable {
 
   private officers?: SubscriptionValue<StaffRecord[]>;
 
+  opportunities: OpportunityType[] = [];
```
macfarlandian, line 88: might rename this to `opportunityTypes` or something so it doesn't get confused with actual Opportunity objects

src/PracticesStore/PracticesStore.ts
```diff
@@ -136,6 +139,15 @@ export class PracticesStore implements Hydratable {
...
     );
+    reaction(
+      () => this.rootStore.currentTenantId,
+      (currentTenantId) => {
+        this.opportunities =
+          (currentTenantId &&
+            tenants[currentTenantId].workflowsOpportunities) ??
+          [];
+      }
+    );
```
macfarlandian, line 150: seems like this could be a computed get method instead of a manual reaction

src/PracticesStore/PracticesStore.ts
```diff
@@ -347,6 +359,7 @@ export class PracticesStore implements Hydratable {
   get opportunityCounts(): Record<OpportunityType, number | undefined> {
     return {
       compliantReporting: this.compliantReportingEligibleCount?.current(),
+      earlyTermination: undefined,
```
macfarlandian, line 362: maybe we update the type to be a `Partial<Record<whatever>>` so that we don't have to do this? (I'm not sure we even need this member anymore honestly, we barely use it and it could just be replace by an `<array>.length` ... it has been on my list to get rid of this but haven't gotten around to it

src/core/ViewNavigation/ViewNavigation.tsx
```diff
@@ -108,7 +111,7 @@ const ViewNavigation: React.FC<ViewNavigationProps> = ({
       <NavLink
         activeClassName="ViewNavigation__navlink--active"
         className="ViewNavigation__navlink"
-        to={`/${PATHWAYS_VIEWS.workflows}`}
+        to={`/${PATHWAYS_VIEWS.workflows}/${opportunities[0]}`}
```
macfarlandian, line 114: I know the routing is a little convoluted but this one was actually intended to go directly to the workflows root url (regardless of whatever happens to be there, allegedly it will be the full caseload search eventually)

src/core/WorkflowsLayouts/WorkflowsNavLayout.tsx
```diff
@@ -86,6 +88,9 @@ const BrandedNavLink = styled(NavLink).attrs({ exact: true })`
 `;
 
 export const WorkflowsNavLayout: React.FC = ({ children }) => {
+  const { practicesStore } = useRootStore();
+  const { opportunities } = practicesStore;
```
macfarlandian, line 92: could be `{practicesStore: { opportunities } } = useRootStore()`

src/core/WorkflowsLayouts/WorkflowsNavLayout.tsx
```diff
@@ -103,11 +108,15 @@ export const WorkflowsNavLayout: React.FC = ({ children }) => {
...
-              </li>
+              {opportunities.map((opportunity) => {
+                return (
+                  <li>
+                    <BrandedNavLink to={workflowsUrl(opportunity)}>
+                      {OPPORTUNITY_LABELS[opportunity]}
+                    </BrandedNavLink>
+                  </li>
+                );
+              })}
```
macfarlandian, line 119: love this. elegant solution

src/core/views.ts
```diff
@@ -73,7 +73,7 @@ export const PATHWAYS_PATHS: Record<string, string> = {
   methodology: `/${PATHWAYS_VIEWS.methodology}/:dashboard`,
   methodologySystem: `/${PATHWAYS_VIEWS.methodology}/system`,
   methodologyOperations: `/${PATHWAYS_VIEWS.methodology}/operations`,
-  practices: `/${PATHWAYS_VIEWS.workflows}`,
+  practices: `/${PATHWAYS_VIEWS.workflows}/:opportunityId`,
```
macfarlandian, line 76: I think we still need a root page, even if we are redirecting everyone in TN away from it at the moment? I doubt we are going to want to replicate that experience in ND because it was basically just a corner we cut to accelerate the first TN release

macfarlandian (changes requested):

src/App.js
```diff
@@ -87,6 +87,7 @@ const App = () => (
                     <ProtectedRoute path={PATHWAYS_PATHS.practices} component={PagePracticesV2} />
                     <Route path="/profile" component={Profile} />
                     <Redirect from="/system" to="/system/prison" />
+                    <Redirect from="/workflows" to="/workflows/compliantReporting" />
```
daschi, line 90: The Recidiviz logo in the `WorkflowsFormLayout` is linking to `/workflows`, which on the ND page is redirecting to here. Wondering if we want to make this state-specific at this point, or will it be something that'll be figured out once we have a landing page?

src/core/ViewNavigation/ViewNavigation.tsx
jovergaag: I had played around with leaving it as the root url, but in this in between state until the root url points to the full caseload search, we are redirecting `/workflows` to `/workflows/compliantReporting` to ensure consistency for TN users. However, that redirect is not valid for ND.

I figured that we have two options
1. Point this global nav link at the first opportunity for now - which works as expected for both TN and ND - then change it to be the full caseload search when we are ready for that to be the homepage at which point we remove the TN redirect
2. Leave this global nav pointing at `/workflows` but use a hook to navigate to the correct opportunity page based on the tenant, and remove the redirect.

The first seemed more straightforward for now and is a temporary solution but I am open to it if you think we should implement #2.

daschi: I like option 1 for now...it seems like option 2 would be useful if we decide not to show the caseload search as the landing page for certain states, and if the opportunities will be prioritized, but that doesn't seem difficult to implement later when it's needed.

src/App.js
jovergaag: Hm, I think you have an old version of the branch. If you pull down the latest version you should see ND navigating to the correct opportunity type.

daschi: Strange I am still seeing it route to `/workflows/compliantReporting` after pulling

jovergaag: After you go to the Profile and select "North Dakota", then clicking on the workflows icon? If I navigate to ND from the profile page it defaults to Early Termination.

jovergaag: Oh I see what you mean, you're manually navigating to `/workflows` while on ND. I see. I'll think about that.

src/core/ViewNavigation/ViewNavigation.tsx
macfarlandian: I guess maybe it depends on the timing of the ND launch versus the TN homepage changeover (a question for @sychang maybe, which one she thinks will happen first?) If we are OK launching in ND without a proper homepage, or if we think TN will be ready to switch over pretty soon, then I think option 1 does work (but there may be more change-management later if we have two or more groups of users whose homepage is going to change). 

Otherwise I think option 2 is better  ... once there are three states then TN will be the exception from the default, and once they switch over we can remove the special-case redirect feature entirely without affecting anything in any other states

src/App.js
daschi: Not just manually - it's by clicking on the Recidiviz logo in the side nav

src/core/ViewNavigation/ViewNavigation.tsx
jovergaag: [This commit](fada03170b27f5702ab8593ce4428365a6ddac77) is a rough attempt at option 2. It allows us to redirect the root url to whatever we want it to be based on the tenant. I think I like it better than option 1 now that it is written out. LMK what you think. If we like it I will write some tests.

macfarlandian: yeah that is pretty good actually ... I thought you were going to have to do something way grosser than that!

sychang: Maybe not relevant anymore, but ND launch will definitely happen before TN changeover.

macfarlandian: no that's good to know! I suspected that would be the case. all the more reason to go with jen's new solution

src/core/views.ts
```diff
@@ -253,6 +254,12 @@ export function getSectionIdForMetric(metric: MetricId): PathwaysSection {
   return PATHWAYS_SECTION_BY_METRIC_ID[metric];
 }
 
+export const WORKFLOWS_PATHS: Record<string, string> = {
+  practices: `/${PATHWAYS_VIEWS.workflows}`,
+  opportunityType: `/${PATHWAYS_VIEWS.workflows}/:opportunityType`,
+  practices404: `/${PATHWAYS_VIEWS.workflows}/not-found`,
```
Catacola, line 260: While we're here, can you update theses to `workflows`?

jovergaag: they will be changed in Daniela's PR I think - I will rebase once that is merged and see where we're at!

src/PracticesStore/PracticesStore.ts
jovergaag: It wasn't being used anywhere so I removed it!

macfarlandian: amazing

src/tenants/US_TN.ts
```diff
@@ -96,6 +96,7 @@ const US_TN_CONFIG: TenantConfig = {
     "supervisionToLiberty",
   ],
   tableColumns: enabledTableColumns[pathways.US_TN],
+  workflowsOpportunities: ["compliantReporting"],
```
jovergaag, line 99: @Catacola I tried adding this directly to the navigation `workflows` key - but that navigation object is set up to control access to specific pages based on permissions, not necessarily to load specific content when on a specific page.

Do we think that we will ever want to show certain opportunities to specific users and not to others? @sychang

jovergaag: Actually - workflows uses the feature variants and not the Pathways permissions framework - so even if we do want to gate access to specific opportunities by users we can do that easily using feature variants.

I think this structure with `workflowsOpportunities` is the cleanest way to implement this rather than trying to embed it within the navigation `workflows` array.

src/tenants/US_ND.ts
```diff
@@ -33,6 +33,7 @@ const US_ND_CONFIG: TenantConfig = {
   enableVitalsCaseloadButton: true,
   navigation: {
     operations: [],
+    workflows: ["earlyTermination"],
```
Catacola, line 36: 😁

Catacola (approved):

src/WorkflowsStore/__tests__/WorkflowsStore.test.ts
```diff
@@ -335,52 +335,6 @@ test("don't subscribe to clients if no officers are selected", async () => {
...
-  runInAction(() => {
-    (workflowsStore.user as any).updates = { selectedOfficerIds: ["OFFICER2"] };
-  });
-
-  await when(
-    () => workflowsStore.opportunityCounts.compliantReporting !== undefined
-  );
-
-  expect(mockUnsub).toHaveBeenCalled();
-});
```
macfarlandian, line 382: this test unfortunately was just watching the deleted method to know when the async subscriptions were finished running; I think we do still want to maintain a test of this behavior so we should restore it and sub in a different computed value to watch (e.g. the clients returned by `mockSubscribeToCaseload`)

src/core/views.ts
```diff
@@ -253,6 +254,12 @@ export function getSectionIdForMetric(metric: MetricId): PathwaysSection {
   return PATHWAYS_SECTION_BY_METRIC_ID[metric];
 }
 
+export const WORKFLOWS_PATHS: Record<string, string> = {
```
macfarlandian, line 257: I think it's actually better ergonomics if you don't declare the type here ... that way Typescript will enforce the specific keys in this object instead of allowing arbitrary ones to be read and written

src/WorkflowsStore/__tests__/WorkflowsStore.test.ts
jovergaag: Good catch. Thank you!

src/WorkflowsStore/__tests__/WorkflowsStore.test.ts
```diff
@@ -335,50 +335,34 @@ test("don't subscribe to clients if no officers are selected", async () => {
...
-
-  await when(
-    () => workflowsStore.opportunityCounts.compliantReporting !== undefined
-  );
-
-  expect(workflowsStore.opportunityCounts.compliantReporting).toBe(mockCount);
-});
-
 test("clean up caseload subscriptions on change", async () => {
   workflowsStore.hydrate();
```
macfarlandian, line 339: you can remove this line since `waitForHydration` calls this method

src/WorkflowsStore/__tests__/WorkflowsStore.test.ts
```diff
@@ -335,50 +335,34 @@ test("don't subscribe to clients if no officers are selected", async () => {
...
+  runInAction(() => {
+    (workflowsStore.user as any).updates = {
+      selectedOfficerIds: ["OFFICER2"],
+    };
+  });
+
+  await when(() => workflowsStore.caseloadClients !== undefined);
 
-  expect(mockUnsub).not.toHaveBeenCalled();
+  expect(mockSubscribeToOfficers).not.toHaveBeenCalled();
```
macfarlandian, line 355: so this is not quite right, what we actually want to verify in this test is that it disposes of its old subscriptions when replacing them with new ones (they don't get garbage-collected automatically and will leak memory otherwise). 

I think you can still just use `mockUnsub` here because there is a default mock implementation of `subscribeToCaseloads` that returns it?

jovergaag: Ok. I clearly don't understand all of this very clearly...but I think it is right now?

macfarlandian: yeah don't feel too bad, i wrote these tests and I barely understand what is happening in them

macfarlandian (approved):
🙌🏻

PR #2094, opened 2022-07-26
[WIP] [Pathways] switch over to using the new backend

danawillow:
## Description of the change

Sending out a draft PR to get some initial feedback on the approach.

Preview app is at https://recidiviz-dashboard-stag-e1108--use-new-be-mxhvz3gc.web.app if you'd like to see what it actually looks like. The only pages that work at the moment are the 4 non-over-time Prison to Supervision pages.

I'm not sure that the autorun is the best place to be reading this data anymore because sometimes it reads data it doesn't need to. For example:
1. Open the network tab
2. Go to https://recidiviz-dashboard-stag-e1108--use-new-be-mxhvz3gc.web.app/system/prisonToSupervision/countByLocation
3. Go to https://recidiviz-dashboard-stag-e1108--use-new-be-mxhvz3gc.web.app/system/prisonToSupervision/personLevelDetail
4. Change a filter value
5. Notice that it sends a request both to `PrisonToSupervisionTransitionsPersonLevel` (what we want) _and_ `PrisonToSupervisionTransitionsCount?group=facility` (which we don't need at the moment)

From my understanding, this is happening because the `PathwaysMetric` is initialized when it gets visited, and then at that point it's in the application memory, so any change to a value read in the autorun (like changing a filter) means that it gets triggered, regardless of whether or not it's the active page.

Feel free to click around (on just those 4 pages) and see if you notice anything else funky, though it's not quite ready for a true QA yet. The last updated times not being there are a known issue (https://github.com/Recidiviz/recidiviz-data/issues/13809) and I'm seeing some diffs with staging that I'll take a look at.

## Type of change

## Related issues

> Closes [#XXXX]

### Development

### Code review


danawillow:
Just rebased on main and re-pushed the demo app, which now also supports the Prison page.

src/core/models/PathwaysMetric.ts
```diff
@@ -174,52 +177,50 @@ export default abstract class PathwaysMetric<RecordFormat extends MetricRecord>
...
-              }
-            } else if (val !== "ALL") {
-              queryParams.append(`filters[${queryKey}]`, val);
-            }
-          });
-        }
+      this.fetchNewMetrics(this.getQueryParams()).then((results) => {
+        runInAction(() => {
+          this.allRecords = results;
+        });
```
macfarlandian, line 189: I think you can address the redundant fetching issue by checking `CoreStore.page`/`CoreStore.section` and deciding whether or not to fetch based on that. I don't know specifically how those things are mapped to one another in this application but conceptually I think that is the info you need. I do still think that an `autorun` or `reaction` is the best way to do this, just have to make sure we are actually reacting to the right information

src/core/models/PathwaysMetric.ts
```diff
@@ -270,10 +271,11 @@ export default abstract class PathwaysMetric<RecordFormat extends MetricRecord>
   async hydrate(): Promise<void> {
     this.isLoading = true;
     try {
-      const fetchedData = await this.fetchAndTransform();
```
jovergaag, line 273: Where is the transform happening now?

src/core/models/utils.ts
```diff
@@ -70,11 +70,12 @@ const libertyDimensionDefaults = {
   judicialDistrict: "ALL",
 };
 
-const prisonDimensionDefaults = {
+export const prisonDimensionDefaults = {
```
jovergaag, line 73: I'm not quite sure why this would be needed anymore with the new BE. This was to default unknown values to ALL, but with the new BE we shouldn't have any unknown values anymore.

jovergaag (commented):
I made a few comments, but I will dig into this more once I can get it up and running locally to compare values locally. 

Also - the preview app that you posted is not getting any data. And in the PR description you said that this is applied to the Prison and Supervision pages, but I'm only seeing how it is applied to the Prison page. Am I missing something here or is the description incorrect maybe?

src/core/models/calculateAggregateMetrics.ts
```diff
@@ -47,6 +48,40 @@ interface AggregateMetricsInterface<T> {
...
+    (date) => {
+      const found = grouped[`${date.getFullYear()},${date.getMonth() + 1}`];
+      const val =
+        found && found[0]
+          ? found[0]
+          : {
+              year: date.getFullYear(),
+              month: date.getMonth() + 1,
+              count: 0,
+            };
```
Catacola, line 73: You can use `?.[0]` as a null-safe accessor on `found`. This then allows you to use the null-coalescing operator (`??`) instead of a ternary operator.

```suggestion
      const val =
        found?.[0]
          ?? {
              year: date.getFullYear(),
              month: date.getMonth() + 1,
              count: 0,
            };
```

src/core/models/PathwaysMetric.ts
macfarlandian: I am still looking into this (will be easier to debug once I can run it locally, which still is not working) but I am also thinking of another approach that might work, which would be to serve up the final filtered data from an `observable` or `computed` member that issues the fetch as a side effect ... there are various ways to trigger behaviors to only occur when an observable is actually being observed, which might help limit these kinds of unnecessary background fetches and maybe be a bit more explicit about it

macfarlandian: OK there were a few things going on here that it took me a while to parse out but I think I have a better handle on it now.

One issue is that this autorun leads to a redundant fetch when you first load a section, because you fetch during hydrate and then flip the loading/initialized flags, which triggers the autorun to immediately fetch again. (One way to manage this is to instead use a `reaction` that only observes the filter state, which you would also need to convert into an `observable` or `computed`; then it would only react to changes in the filter values and wouldn't be triggered when hydration completes.)

The other is the one you have pointed out, although I am not totally sure whether we should consider that a bug or a feature. Because the filter state persists across pages, if we didn't do all these background fetches then the data would drift out of sync with the filter state and would be incorrect if you navigated back to a previously viewed page. I guess the alternative would be to return them to an unhydrated state when a background change occurs so the data can be refetched lazily when it's needed? (Something like [this example](https://mobx.js.org/lazy-observables.html) maybe?) Unless we are concerned about all these fetches affecting performance somehow, it might be more trouble than it's worth? If we do want to go down that road we might think about ways to observe the navigation state more explicitly instead of relying on mobx magic, to keep the increasingly complex logic readable.

danawillow: Thanks Ian! I figured out how to get the `reaction` working, and also updated it to only create the reaction when the metric is observed, which together seem to fix the double-call issue and eliminate the need of `initialized` 🙌 

Also now that you've said it, I agree with you about the filter state across pages, so I've left that be.

src/core/models/PathwaysMetric.ts
danawillow: In the backend itself!

src/core/models/utils.ts
danawillow: Thanks for pointing this out! I was doing it for type reasons in calculateAggregateMetrics, but this made me realize I should try just changing the type to the one I actually need to use, so no need for this anymore.

src/core/models/calculateAggregateMetrics.ts
danawillow: 👍 thanks!

danawillow:
Thanks for the comments all! I'm going to spend some time working through the rest of the metrics and their associated tests. Feel free to continue reviewing if there's anything else approach-wise you'd like to add, but otherwise no rush.

src/core/models/PathwaysMetric.ts
```diff
@@ -171,55 +171,71 @@ export default abstract class PathwaysMetric<RecordFormat extends MetricRecord>
...
-              const timePeriod = getTimePeriodRawValue(val);
-              if (timePeriod) {
-                queryParams.append(`filters[${queryKey}]`, timePeriod);
-              }
-            } else if (val !== "ALL") {
-              queryParams.append(`filters[${queryKey}]`, val);
-            }
+    // Update the data in allRecords when a filter changes, but only after allRecords has been
+    // read the first time.
+    onBecomeObserved(this, "allRecords", () => {
```
macfarlandian, line 176: I _believe_ this hook can be called multiple times as observation stops and restarts (which we would expect to happen if you navigated away from a page and then back to it), which would result in redundant reactions and maybe a bit of a memory leak. I think the simplest alternative is just to create the reaction immediately in the constructor but short-circuit the effect function if the metric hasn't been hydrated yet?

danawillow: oh that makes sense. Done!

PR #2101, opened 2022-07-27
[Workflows] Use firestore rules to prevent cross-state data access

### LINKED ISSUE #2055: [Workflows] Update firestore to prevent cross-state access

Catacola:
## Description of the change

Update the firestore rules to add a bunch of security layers:
- All ETLs are read-only.
- ETL data can only be read by Recidiviz users or users whose stateCode matches the stateCode field in the record
- ClientUpdate (notV2) is read-only and restricted to Recidiviz and TN
- ClientUpdateV2 is read/write enabled and restricted to recidiviz users and users whose stateCode matches the prefix of the client record (e.g. `us_tn_345234523` can only be read by a user with `stateCode: "us_tn")
  - Additionally, it will ensure that all writes (even from Recidiviz users) have an appropriate prefix
- UserUpdates and FeatureVariants are restricted to only being read by the user who it applies to.
- Same rules are in place for all DEMO buckets

Tested via firestore rules, offline mode, and deploying rules to firestore staging and testing with both a recidiviz user and a TN test user.

## Type of change

## Related issues

Closes #2055

### Development

### Code review


firestore-config/firestore.rules
```diff
@@ -1,30 +1,40 @@
 rules_version = '2';
 service cloud.firestore {
   function allowedStateCode() {
-    return request.auth.token.stateCode in ["US_TN", "RECIDIVIZ"]
+    return request.auth.token.stateCode in ["US_ND", "US_TN", "RECIDIVIZ"]
+  }
+
+  function stateCodeMatchesRead() {
+    return request.auth.token.stateCode == "RECIDIVIZ"
+        || request.auth.token.stateCode == resource.data.stateCode
```
daschi, line 9: Trying to learn about the firestore rules, but what is the difference between `resource.data.stateCode` here and `request.resource.data.stateCode` on the write? Is `resource` available at the top level for reads?

firestore-config/reads.test.js
```diff
@@ -59,17 +59,73 @@ async function testAllReads(db, assertFn) {
...
 test.each([
   ["TN", getTNUser],
-  ["Recidiviz", getRecidivizUser],
-])("%s user can read", async (userType, getUserContext) => {
-  await testAllReads(getUserContext(testEnv).firestore(), assertSucceeds);
-});
+  ["ND", getNDUser],
+])(
+  "%s user can read their own state data",
+  async (userState, getUserContext) => {
```
daschi, line 124: Might be overkill but wanted to ask if we should add a test for "user cannot read other state data"

firestore-config/writes.test.js
```diff
@@ -73,22 +87,74 @@ test.each([
...
+    assertSucceeds,
+    "US_ND"
   );
 });
+
+test.each([
+  ["TN", getTNUser, "ND"],
+  ["ND", getNDUser, "TN"],
+])(
+  "%s user can write cross-state data",
```
daschi, line 151: nit: user can not write cross-state data
```suggestion
  "%s user can not write cross-state data",
```

daschi (changes requested):

firestore-config/firestore.rules
macfarlandian: IIRC `resource` is the existing document (for both reads and writes) and `request.resource` is the incoming write

firestore-config/reads.test.js
macfarlandian: +1

firestore-config/firestore.rules
Catacola: ^ Bingo

firestore-config/reads.test.js
Catacola: Not overkill. Totally meant to write this, but guess I only did it for the writes.

macfarlandian (changes requested):

firestore-config/firestore.rules
Catacola: Ohhhhhh, that just made me realize that we are open to a vulnerability where a TN user can overwrite a ND doc and vice versa as long as they set the new state code to their state code.

firestore-config/writes.test.js
```diff
@@ -73,22 +87,74 @@ test.each([
...
+test.each([["Recidiviz", getRecidivizUser]])(
+  "%s user can write without a stateCode",
+  async (userType, getUserContext) => {
+    await testWriteToCollectionsWithoutStateCode(
+      UPDATE_COLLECTION_NAMES,
+      getUserContext(testEnv).firestore(),
+      assertSucceeds
+    );
+  }
+);
```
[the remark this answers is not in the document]
macfarlandian: sorry I think i got sidetracked by something in the middle of this review and forgot to add this, but maybe we need to migrate the user-generated documents to include a state code field before making these security changes? I suspect that something is going to break (possibly with reads as well as writes)

firestore-config/firestore.rules
```diff
@@ -1,30 +1,40 @@
 rules_version = '2';
 service cloud.firestore {
   function allowedStateCode() {
-    return request.auth.token.stateCode in ["US_TN", "RECIDIVIZ"]
+    return request.auth.token.stateCode in ["US_ND", "US_TN", "RECIDIVIZ"]
+  }
```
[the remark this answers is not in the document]
Catacola: I've been debating that same thing. The only thing it may open us up to is someone in a state accessing data while we're still developing for that state. E.g. any authorized user can create a firebase token, even if workflows isn't enabled in their state. Say we're developing a new tool for ID. An ID user could, in theory, generate a valid firestore token and read ID data from firestore. I'm not sure this is a big issue, but it could happen with this change.

firestore-config/writes.test.js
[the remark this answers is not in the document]
Catacola: fine by me

firestore-config/firestore.rules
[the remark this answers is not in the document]
macfarlandian: ah I thought they would be prevented from generating a token by the permissions check in auth0 but I see that is not the case. I do agree this concern is valid if not hugely likely, but maybe we should solve for it in the token exchange endpoint instead? (e.g. by inspecting the user permissions or something)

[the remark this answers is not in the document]
Catacola: Agreed! Created a task to do so here: https://github.com/Recidiviz/pulse-dashboard/issues/2104

firestore-config/writes.test.js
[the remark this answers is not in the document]
Catacola: Sorry, missed your second message. That migration should be extremely easy. We can just slap `stateCode: "US_TN"` on everything.

firestore-config/writes.test.js
```diff
@@ -73,22 +87,130 @@ test.each([
...
+
+test.each([
+  ["TN", getTNUser, "ND", getNDUser],
+  ["ND", getNDUser, "TN", getTNUser],
+])(
+  "%s user cannot overwrite another state's data",
+  async (userState, getUserContext, otherUserState, getOtherUserContext) => {
+    await testWriteToCollectionsForState(
+      UPDATE_COLLECTION_NAMES,
+      getOtherUserContext(testEnv).firestore(),
```
daschi, line 160: I think I'm confusing myself on how this should work - I expected that this would be `getUserContext` and the state code would be `otherUserState` and that would fail, and then below `getUserContext` with `userState` would pass?

Catacola: With the first command, I'm building a bunch a document with the name "foo" in each of the collections with the state_code of the second state. I then make sure that the first state cannot override the other state's date, even if they try to set the state code to match theirs. This was a vulnerability in my original implementation of these rules.

daschi (approved):

firestore-config/firestore.rules
```diff
@@ -1,30 +1,54 @@
...
-    
 
     match /{document=**} {
-      allow read: if allowedStateCode()
+      allow read: if isRecidiviz() || stateCodeMatchesTarget();
     }
 
     match /userUpdates/{document=**} {
-      allow write: if allowedStateCode()
+      allow write: if allowWrite();
```
macfarlandian, line 35: maybe these `write` rules should all be `update` instead now that there is a separate `create` rule? probably makes no practical difference in its current state but if something changes this would clobber a more restrictive `create` rule

firestore-config/writes.test.js
```diff
@@ -73,22 +87,155 @@ test.each([
...
+    );
+
+    await testWriteToCollectionsForState(
+      UPDATE_COLLECTION_NAMES,
+      getRecidivizUser(testEnv).firestore(),
+      assertSucceeds,
+      `US_${newState}`
+    );
+  }
+);
```
macfarlandian, line 241: i'm not so sure we should allow this actually ... I can't think of a use case for it, seems like something we would only do by mistake and bring ruin upon ourselves

macfarlandian (commented):
this is looking great but a couple of small comments. also this is definitely going to break all writes now so we need to block it on both a data migration and a FE change to update the write methods

firestore-config/firestore.rules
```diff
@@ -1,30 +1,54 @@
...
+      allow create: if allowCreate();
```
macfarlandian, line 36: noticed something else here actually ... for this collection (as well as for `featureVariants`) I think we should instead check that the document ID matches the email address in the auth token, for both reads and writes. There is no real use case for letting a user access someone else's preferences (including Recidiviz users since we don't actually administer anything via the frontend)

firestore-config/writes.test.js
Catacola: I was thinking it would be for us to fix a mistake, but I guess we could always do that in the actual firestore UI

macfarlandian: yeah i think that would be the preferred method. i am not sure by what mechanism we would do it within the workflows app anyway? (if there is one it's probably not there on purpose)

firestore-config/firestore.rules
```diff
@@ -1,38 +1,96 @@
...
   }
 
   match /databases/{database}/documents {
-    
 
-    match /{document=**} {
-      allow read: if allowedStateCode()
+    match /featureVariants/{record} {
+      allow read: if usernameMatches(record);
+      allow write: if false;
```
macfarlandian, line 53: you can just drop this line, I believe nothing is allowed by default

firestore-config/firestore.rules
```diff
@@ -1,38 +1,96 @@
...
     }
 
-    match /clientUpdatesV2/{document=**} {
-      allow write: if allowedStateCode()
+    match /clientUpdatesV2/{clientId}/{document=**} {
+      // documentIDs for the client updates are of the form us_xx_${client_id}
+      // check to make sure the user's state code matches the prefix on the record
+      allow read: if isRecidiviz() || stateCodeMatchesClientId(clientId);
+      allow write: if hasProperStateCodePrefix(clientId) && (isRecidiviz() || stateCodeMatchesClientId(clientId));
```
macfarlandian, line 79: I did not actually believe it was possible to do this without a migration until I reached this line. incredible

firestore-config/testUtils.js
```diff
@@ -27,13 +27,21 @@ export const ETL_COLLECTION_NAMES = [
...
  */
 export const ADMIN_COLLECTION_NAMES = ["featureVariants"];
 /**
- * Update collections are writeable
+ * Update collections are readable and writeable by all users within a state
  */
-export const UPDATE_COLLECTION_NAMES = ["userUpdates", "clientUpdates"];
+export const SHARED_UPDATE_COLLECTION_NAMES = ["clientUpdatesV2"];
+/**
+ * Update collections are readable and writeable only by TN or Redidiviz
```
macfarlandian, line 38: reviditiz

server/workflows/firebaseToken.js
```diff
@@ -42,7 +44,7 @@ async function getFirebaseToken(req, res) {
     stateCode = getAppMetadata({ user }).state_code;
     uid = user.email;
   } else {
-    uid = req.user.sub;
+    uid = req.user[`${METADATA_NAMESPACE}email_address`];
```
macfarlandian, line 47: is this ... broken in the wild? or do these values just happen to be the same but this is the more reliable source?

firestore-config/firestore.rules
```diff
@@ -1,38 +1,96 @@
...
+  }
+
+  function usernameMatches(id) {
+    return request.auth.token.user_id.lower() == id.lower();
+  }
+
+  function hasProperStateCodePrefix(clientId) {
+    // Check to make sure an ID is properly prefixed by the state code
+    // e.g. "us_xx_someID"
+    return clientId.size() > 7 && clientId.lower()[0:3] == "us_" && clientId[5] == "_";
```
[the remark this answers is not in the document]
Catacola: It could totally be done in regex. I even use a regex like 4 lines below.

server/workflows/firebaseToken.js
Catacola: Auth0 supplies a string that may or may not include your email address. I changed it to be the user's email address assuming we got to specify the uid to be whatever we wanted as long as it was unique. It worked against the staging firestore for me. I can revert this and add a new field with the user's email if that would be better.

src/WorkflowsStore/__tests__/WorkflowsStore.test.ts
```diff
@@ -90,7 +90,7 @@ function doBackendMock() {
 }
 
 async function waitForHydration(): Promise<void> {
-  workflowsStore.hydrate();
+  await workflowsStore.hydrate();
```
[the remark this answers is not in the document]
Catacola: I think it happened because pycharm yelled at me so it did it to make it happy, but it looks like `waitFoyHydration()` might do the trick as well.

firestore-config/firestore.rules
Catacola: I'm pretty sure that's true. I was being pedantic, but I guess that's overkill.

PR #2107, opened 2022-07-28
[Workflows] recommended language modal for almost eligible criteria

macfarlandian:
## Description of the change

Constructs a blob of recommended language for the PO to send to the client regarding how to attain eligibility, when applicable, displays said blob in a modal, and lets you copy it to your clipboard. (I used a library for this, which we have also used in Spotlight for similar functionality. I didn't see anything existing in this repo that would do it but correct me if I'm wrong!)

## Type of change

## Related issues

> Closes https://github.com/Recidiviz/recidiviz-data/issues/13263

### Development

### Code review


package.json
```diff
@@ -140,7 +140,7 @@
     "react-dom": "^17.0.0",
     "react-helmet": "^6.1.0",
     "react-intersection-observer": "^9.1.0",
-    "react-modal": "^3.14.3",
+    "react-modal": "^3.12.1",
```
macfarlandian, line 143: Even though the ranges overlapped, I needed to match this to exactly what the design-system requires to prevent Yarn from installing multiple versions of this package, which was causing an error because the library maintains some kind of global state for accessibility purposes (related to declaring the application root in `index.js`)

src/WorkflowsStore/Opportunity/CompliantReportingOpportunity.ts
```diff
@@ -642,6 +682,66 @@ export class CompliantReportingOpportunity implements Opportunity {
     );
   }
 
+  get almostEligibleRecommendedNote():
+    | { title: string; text: string }
+    | undefined {
+    // note functionality only supports a single missing criterion
```
jovergaag, line 688: Can a client be almost eligible even if they are missing more than 1 criterion?

jovergaag (approved):
Super cool! I did have to delete my node modules to get it to build with the `react-modal` downgrade but thats no biggie.

src/WorkflowsStore/Opportunity/CompliantReportingOpportunity.ts
macfarlandian: they will be in the data, yes, but we are suppressing them in the app for now (there is logic in the opportunity creation flow to enforce this) ... this is just a failsafe in case we change that without updating this so it doesn't output something inaccurate

PR #2115, opened 2022-07-29
[Workflows] Support reading/writing to clientUpdatesV2 with globally unique id

### LINKED ISSUE #2090: [Workflows] Support reading client information by state
In #2053 and #2054, we're converting the backend to store client information in firestore prefixed by the state code. (e.g. `us_xx_0012345`). The goals of this task is to update the frontend to pull from the state-prefixed document. If the state-prefixed document doesn't exist, fall back to pulling the document at the plain client id for backwards compatability.

daschi:
## Description of the change

This updates the subscription to `clientUpdates` to first subscribe to `clientUpdatesV2`, and to then subscribe to `clientUpdates` collection if the document does not exist. 

It also updates the write functions so that we are merging the old document and the new changes and saving it to the new `clientUpdatesV2` collection.

This adds a new rule for the new collection, which I can update once the other firestore rules PR is merged.

## Type of change

## Related issues

> Closes #2090 

### Development

### Code review


daschi:
I’ll look into the failing test on Monday! I think I am checking for an object as falsy and that’s failing, but was having by TS issues with using ‘Object.keys’

src/WorkflowsStore/Client.ts
```diff
@@ -244,7 +262,15 @@ export class Client {
   }
 
   get updates(): ClientUpdateRecord | undefined {
-    return this.fetchedUpdates.current();
+    // TODO(#2108): Clean up requests to `clientUpdates` after fully migrating to `clientUpdatesV2`
+    if (
+      !this.fetchedUpdatesV2.current() ||
```
macfarlandian, line 267: I think you don't want this test here because it will only be falsy (i.e. `undefined`) while the initial fetch is pending. So it would result in always using the value from the old collection

macfarlandian (changes requested):

src/WorkflowsStore/Client.ts
daschi: Once the initial fetch completes though, would this not get re-computed?

daschi: I am seeing v2 update on the frontend, but when I refresh I see it flash from v1 to v2 which is not great....I'll think about this a bit to see if we can wait until both are loaded before returning something

daschi: Okay I switched this around to return v2 from the `if` statement...let me know if that works better! I tested this with the staging firestore and it seemed to be doing all the right things 🤞

daschi: also happy to switch this to use `compliantReportingUpdates` as the collection name instead

macfarlandian (approved):
I realized that if we are going to move to subcollections as discussed in https://github.com/Recidiviz/pulse-dashboard/pull/2149#discussion_r938333947 then there isn't really any benefit to doing it as part of this PR, as we still have to do this migration to implement the new ID scheme. I'm going to go ahead and get this merged to unblock further development

Issue #2126, opened 2022-08-01
[Workflows][Multi-tenant] Add state_code subcollection for recidiviz users in userUpdates collection


daschi:
**What needs to be done? Why does it need to be done?**
Once we have multiple states in the app we'll need to divide the `userUpdates` selectedOfficerIds by state code for recidiviz users. 

Question: Should we add the subcollection for all users or just recidiviz users?

**Additional context**
https://paper.dropbox.com/doc/Design-Doc-Workflows-Supporting-multiple-tenants--BmLqaOHcOSPFOeUVj7Z3eCoRAg-mlLZIK8OYPuALCliBbXdo

macfarlandian:
had a somewhat random thought about this: since it is a special and somewhat rare case (that possibly will NEVER affect an external user), why don't we just add an extra optional `map` field to the `userUpdates` document instead of creating subcollections? It's probably easier to change which field we check on the fly than to get a whole new document or subscription (especially if we at some point introduce fields that we don't want to segment by state)

macfarlandian:
pretty much yeah

PR #2128, opened 2022-08-01
[Workflows] EarlyTerminationOpportunity skeleton

### LINKED ISSUE #2091: [US_ND] Create EarlyTerminationOpportunity.ts
We now define much of the config around an opportunity in a single class. See `src/PracticesStore/Opportunity/CompliantReportingOpportunity.ts` for an example.

The goal of this task is to create the `EarlyTerminationOpportunity` class.

Assume the earlyDischargeRecord will look like this and only exist for eligible clients:
```
{
  stateCode: str;
  externalId: str;
  formInformation: {
    judgeName: str;
    sentencingDate: date;
    sentenceLengthYears: num;
    chargeName: str;
  };
  reasons: {
    pastEarlyDischarge: {eligibleDate: date};
    eligibleSupervisionLevel: {supervisionLevel: str};
    eligibleSupervisionType: {supervisionType: str};
    notActiveRevocationStatus: {};
  };
} 
```

### LINKED ISSUE #2098: [Workflows] Update caseloadViews.tsx to display state-specific opportunities
The all-client view hosted at `caseloadViews.tsx` currently is hardcoded to show an opportunity list for every possible opportunity.

Change it to only iterate over the opportunities enabled for that state.

### LINKED ISSUE #2103: [Workflows] Create Early Termination page

jovergaag:
## Description of the change

Skeleton of the `EarlyTerminationOpportunity` along with updated fixture data to enable offline mode. I also added a `staff` record with an associated client ans associated earlyTerminationReferral in Firestore to enable the officer search on the "Early Termination" opportunity page.

Most of the methods in the `EarlyTerminationOpportunity` class are empty placeholders for now. There are tickets to continue to build out each of these methods individually, instead of holding up this PR on that logic since so many things depend on this.

## Type of change

## Related issues

> Closes #2091
> Closes #2098
> Closes #2103

### Development

### Code review


src/WorkflowsStore/Opportunity/EarlyTerminationOpportunity.ts
```diff
@@ -0,0 +1,160 @@
...
+
+  // eslint-disable-next-line class-methods-use-this
+  get requirementsMet(): OpportunityRequirement[] {
+    return [];
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  get requirementsAlmostMet(): OpportunityRequirement[] {
+    return [];
+  }
```
jovergaag, line 136: I left all of these class methods essentially blank because I'm not clear on the requirements for them. I am debating whether to try to merge this as is and start to flesh out these methods when we have more details on the opportunity requirements, or try to dig for these requirements now. Thoughts are appreciated.

src/WorkflowsStore/Opportunity/EarlyTerminationOpportunity.ts
```diff
@@ -0,0 +1,188 @@
...
+    sentenceLengthYears: number;
+    chargeName: string;
+    remainingFees: number;
+  };
+  reasons: {
+    pastEarlyDischarge?: { eligibleDate?: Date };
+    eligibleSupervisionLevel?: { supervisionLevel?: string };
+    eligibleSupervisionType?: { supervisionType?: string };
+    notActiveRevocationStatus?: { status?: string };
+  };
```
jovergaag, line 47: @Catacola I'm a little unsure of this `reasons` object. It seems like an array would make more sense here. The ticket requirements listed this as an object, but I'm wondering if an array would be easier to work with. LMK what you think.

src/WorkflowsStore/Opportunity/EarlyTerminationOpportunity.ts
```diff
@@ -0,0 +1,188 @@
...
+};
+
+export class EarlyTerminationOpportunity implements Opportunity {
```
macfarlandian, line 50: i think it's best to avoid exporting this class so we don't instantiate it outside of the creator function by mistake

src/WorkflowsStore/Opportunity/EarlyTerminationOpportunity.ts
macfarlandian: i think in practice it would be kind of annoying at the type level to have the methods be optional; it's fine for them to just always return `false` or an empty array or whatever because, at least for now, if there aren't any almost-eligible clients then none of the related UI elements will be visible. And that way we have a single consistent interface and don't have to add an extra layer of checks for `undefined` throughout the UI.

src/WorkflowsStore/Opportunity/EarlyTerminationOpportunity.ts
```diff
@@ -0,0 +1,188 @@
...
+
+  // eslint-disable-next-line class-methods-use-this
+  get almostEligible(): boolean {
+    return false;
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  get rank(): number {
+    return 0;
+  }
```
macfarlandian, line 139: assuming the statuses will be the same across opportunities, the `rankByReviewStatus` util function is boilerplate and you can just include it here now

src/WorkflowsStore/Opportunity/EarlyTerminationOpportunity.ts
```diff
@@ -0,0 +1,188 @@
...
+    return defaultOpportunityStatuses[this.reviewStatus];
+  }
+
+  get statusMessageLong(): string {
+    return defaultOpportunityStatuses[this.reviewStatus];
+  }
+
+  // eslint-disable-next-line class-methods-use-this
+  get requirementsMet(): OpportunityRequirement[] {
+    return [];
```
macfarlandian, line 158: the `reasons` correspond to fields in the [copy sheet](https://docs.google.com/spreadsheets/d/1wvNWKxcvDAJyi1J71JD87cnJa1toBGkkP8N0GuQJlWk/edit#gid=470391875) so you could probably throw those in here at this point also

macfarlandian (commented):
sorry, I inadvertently published my review while I was only halfway done, had a few more comments here. 

Overall I think that based on the available information you can flesh this out a little bit more and it would probably be worth merging at that point. I don't know (and I don't know if we have really yet figured out) how the user-generated portion is going to be structured but I don't think we necessarily have to block on that ... e.g. that could happen in parallel with the form modeling and UI being developed, both of which would be blocking on having the skeleton in place

src/WorkflowsStore/Opportunity/EarlyTerminationOpportunity.ts
Catacola: Would it be possible to make default implementations of these "optional" functions in the base abstract class?

src/WorkflowsStore/Opportunity/EarlyTerminationOpportunity.ts
Catacola: How so? Do you mean like `[{type: 'pastEarlyDischarge', details: { eligibleDate?: Date }}, ...]`?

src/firestore/types.ts
```diff
@@ -117,6 +117,7 @@ export type ClientRecord = {
   lastSpecialConditionsNote?: string;
   specialConditionsTerminatedDate?: string;
   compliantReportingEligible?: CompliantReportingEligibleRecord;
+  earlyTerminationEligible?: EarlyTerminationEligibleRecord;
```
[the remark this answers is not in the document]
Catacola: Whoops, sorry, this will just be a bool

daschi:
Just a general comment that I think we can make `featureVariants` collection have a stateCode subcollection. I'll add it to this ticket if people agree: https://app.zenhub.com/workspaces/polaris-614c9f27fa19b10015ddd495/issues/recidiviz/pulse-dashboard/2126

src/WorkflowsStore/Client.ts
```diff
@@ -192,6 +193,10 @@ export class Client {
         record.compliantReportingEligible,
         this
       ),
+      earlyTermination: createEarlyTerminationOpportunity(
+        record.earlyTerminationEligible,
+        this
+      ),
```
daschi, line 199: I'm curious if it's already been discussed, but I was wondering about subclassing by state, like `UsNdClient` or something, so that we can keep state-specific logic separate from shared properties?

src/WorkflowsStore/Opportunity/EarlyTerminationOpportunity.ts
jovergaag: It looks to me like the statuses are different. From the PRD:

The following statuses will be present in the list view:
Early termination date in X days
Past early termination date
Early termination paperwork complete

I'm still trying to chase down what the enum values are going to be...

src/WorkflowsStore/Opportunity/EarlyTerminationOpportunity.ts
jovergaag: I've spent quite a bit of time this morning chatting with Justine about this reasons object. I better understand now what each of the keys mean, and it does make sense to leave it as an object. 

Basically, we will need to validate the opportunity by checking that there is a valid eligible_date, valid supervision_level, valid supervision_type and null revocation_date for eligible opportunities. I will add that validation in this PR.

jovergaag:
> What do you think about adding stateCode subcollections to featureVariants? I can add it to this ticket if people agree: https://github.com/Recidiviz/pulse-dashboard/issues/2126

I can see that it could be helpful! I like the idea.

src/firestore/types.ts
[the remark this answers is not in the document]
Catacola: Sorry, I meant to finish this. The plan was that the record with the form info and reasons would go in its own collection (e.g. `EarlyTerminationReferrals`) to keep the client object from ballooning. For now, I was planning on using a bool on the client record to indicate eligibility. In the future, we could change it to a list of eligible opportunities once we have more than a few of them.

src/WorkflowsStore/Client.ts
jovergaag: I suppose eventually a second state could also have the "earlyTermination` opportunity enabled, in that case I would hope that we will try to keep the logic the same across all states with that opportunity so that it can be supported by the same views and FE logic. If we don't think that that is the case then I think tenant specific opportunity classes probably make more sense than tenant specific client classes. I'm not sure how much tenant specific client logic there is going to be other than which opportunities are enabled.  Definitely open to other ideas around that though!

src/WorkflowsStore/Opportunity/EarlyTerminationOpportunity.ts
Catacola: Currently, the ETL will only load records for eligible clients, so you only need to validate the shape, not that the client is eligible. However, in the future when we do almost eligible clients, we will need to do some sort of counting/validation

Catacola:
> What do you think about adding stateCode subcollections to `featureVariants`? I can add it to this ticket if people agree: #2126

I think that's a fine idea

macfarlandian:
> What do you think about adding stateCode subcollections to `featureVariants`? 

I'm not sure I really see the use case for it actually? I don't really see us having a complex matrix of generic feature x state that we would need to manage at such a granular level (and only for ourselves because external users are already limited to a single state)

src/WorkflowsStore/Client.ts
macfarlandian: I think `Client` in its current state owns a lot of logic that we probably want to push down to the opportunity level (e.g. dealing with the form data), so I agree with Jen that we will hopefully get to a point where the Client object does not do much of anything except maybe fetch and distribute data. There is probably going to have to be state-specific configuration of some kind at the client level, though ... could be some bits of copy, could be different fields we need to include in the client profile, stuff like that. I wouldn't necessarily do it as part of this PR but it's something to keep in mind as we build out ND stuff

src/WorkflowsStore/Opportunity/EarlyTerminationOpportunity.ts
macfarlandian: ha ok sorry i guess i didn't read the spreadsheet carefully enough. I think you could just add OVERDUE as an option to the generic function though (e.g. it's probably fine if TN just ignores it as long as the relative order of the values it does use stays the same)

jovergaag:
> I'm not sure I really see the use case for it actually? I don't really see us having a complex matrix of generic feature x state that we would need to manage at such a granular level (and only for ourselves because external users are already limited to a single state)

The use case that I see is that if we have a feature variant enabled, like we did for US_TN for almost eligible, and we need to manually manipulate clients for another state like I did for ND early termination, it is tedious to manipulate the client record in both  `SANDBOX_clients` and `clients`.  This is probably a rare case, and not a ton of work, but state specific feature variants would allow us to keep manual updates contained to the `clients` collection.

But I do see your point @macfarlandian that this may introduce more complexity than it is solving for.

macfarlandian:
>  state specific feature variants would allow us to keep manual updates contained to the `clients` collection.

hmm interesting ... FWIW the SANDBOX_clients collection is no more. Hopefully we don't have to resort to such a thing again, I did it because it was an easy way to unblock development at the time but if it will cause other downstream problems there is no need to repeat it. (really i was just too lazy to create a bunch of data fixtures for almost eligible)

src/WorkflowsStore/Opportunity/EarlyTerminationOpportunity.ts
jovergaag: It turns out that the copy spreadsheet had different statuses than the PRD - and these statuses line up with the CR statuses so I just left them as is, with custom messages.

src/WorkflowsStore/Opportunity/EarlyTerminationOpportunity.ts
jovergaag: There is a [ticket](https://app.zenhub.com/workspaces/polaris-614c9f27fa19b10015ddd495/issues/recidiviz/pulse-dashboard/2102) specifically for this so I will not add anything additional in this PR.

src/WorkflowsStore/Opportunity/EarlyTerminationOpportunity.ts
jovergaag: There is a ticket to validate the opportunity (#2144) so I will not add any logic in this PR.

src/firestore/types.ts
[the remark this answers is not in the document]
jovergaag: I moved everything into `EarlyTerminationReferrals`

jovergaag:
@macfarlandian @Catacola @daschi  This is ready for another review. I ended up moving all of the early termination data into the `earlyTerminationReferrals` collection instead of as an attribute on the `clients` collection. The only thing still on `clients` is the `earlyTerminationEligible` boolean.

I also left many of the methods in the opportunity blank or basic and created tickets for each of them specifically. There is so much blocked by this PR, I wanted to get it in asap so we can start working on other tickets.

macfarlandian (approved):

src/WorkflowsStore/Opportunity/__fixtures__/index.ts
```diff
@@ -95,3 +96,52 @@ export const CompliantReportingAlmostEligibleCriteria: Required<
...
   recentRejectionCodes: ["TEST1"],
 };
+
+export const earlyTerminationEligibleClientRecord: RequireKeys<
+  ClientRecord,
+  "earlyTerminationEligible"
+> = {
+  personName: {
+    givenNames: "JAMIE",
+    surname: "JONES",
```
Catacola, line 106: I have a long-time friend and former roommate with this exact name.

src/WorkflowsStore/Opportunity/EarlyTerminationOpportunity.ts
```diff
@@ -0,0 +1,175 @@
...
+class EarlyTerminationOpportunity implements Opportunity {
+  client: Client;
+
+  readonly type: OpportunityType = "earlyTermination";
+
+  private record: EarlyTerminationReferralRecord;
+
+  constructor(record: EarlyTerminationReferralRecord, client: Client) {
+    makeAutoObservable<
+      EarlyTerminationOpportunity,
```
Catacola, line 46: Thanks!

Catacola (approved):

PR #2132, opened 2022-08-02
[Workflows] ND Early Termination with routing fix

jovergaag:
## Description of the change

Reverts the ND Early Termination revert, and adds the routing fix.

The fix is to add "client" to the `workflows` navigation array. The navigation object has a key for every view, and each of these view keys lists out the enabled pages for that view. Since we have a "client" page, we need to list it here otherwise it will never be allowed for users. However, in the case of workflows, the `workflows` key also lists opportunity types because they are BOTH opportunity types and pages. This means that when accessing navigation object in `WorkflowsStore` to parse out the enabled opportunity types for a given tenant, we need to filter out the "client" page. It's not the most graceful thing, but assuming "client" is the only non-opportunity page that we will support I think it is not an awful solution.

I did play around with trying to structure the routes/pages/permissions in a way that the navigation object was designed for, which would be something like 
` { workflows: ["opportunityType", "client"], opportunityType: ["compliantReporting", "earlyTermination"] }

But that means that I would need to change the routes to `/workflows/opportunityType/compliantReporting` instead of what TN is currently using `/workflows/compliantReporting" which would be disruptive for users and create more of a mess. It means we would also need to update our user permissions for TN which is also a little risky and messy.

## Type of change

## Related issues

> Closes [#XXXX]

### Development

### Code review


src/WorkflowsStore/WorkflowsStore.ts
```diff
@@ -390,8 +390,9 @@ export class WorkflowsStore implements Hydratable {
   get opportunityTypes(): OpportunityType[] {
     return (
       (this.rootStore.currentTenantId &&
-        (tenants[this.rootStore.currentTenantId].navigation
-          ?.workflows as OpportunityType[])) ??
+        (tenants[this.rootStore.currentTenantId].navigation?.workflows?.filter(
+          (opp) => opp !== "client"
```
macfarlandian, line 394: I think instead of this it might be cleaner to maintain the opportunity mapping as a separate config object instead of building it directly into navigation? we can spread it into the navigation config alongside any non-opportunity routes but then we could access it directly here without having to keep this filter in sync with future unrelated route changes

macfarlandian (changes requested):

macfarlandian (approved):
🙌🏻

Issue #2152, opened 2022-08-05
[Workflows] Move clientUpdates into subcollections per tenant


jovergaag:

jovergaag:
https://github.com/Recidiviz/pulse-dashboard/pull/2149#discussion_r938333947

macfarlandian:
I think what we want here is a subcollection for opportunities. The structure I would propose is this: 
- **collection:** `clientUpdatesV2`
  - **id:** `<client GUID>`
    - **document**: `{ ...whatever }` (probably nothing for now but in Case Triage we had some generic client-level updates so presumably that will be supported here eventually as well)
    - **collection:** `opportunities`
      - **id:** `compliantReporting`
        -  **document:** `CompliantReportingUpdateRecord`
      - **id:** `earlyTermination`
        -  **document:** `EarlyTerminationUpdateRecord`
      - etc

PR #2156, opened 2022-08-08
[Workflows] Allow client full profile to support multiple opportunity types

### LINKED ISSUE #2099: [Workflows] Update `FullProfile.tsx` to iterate over all enabled opportunities in a state
Currently, the full profile is hardcoded to only check for the compliant reporting opportunity. Instead, we should iterate over all opportunities enabled for that state.

Feel free to return a stub component for the Early Termination opportunity if the preview component isn't built yet.

### LINKED ISSUE #2172: [Workflows] Update denial reasons codes in dropdown
These are the reason codes that are currently showing up in the drop down - they are made up by me - it would be nice to get a code or string to display in the dropdown instead.

This is where they show up:

jovergaag:
## Description of the change

Changes `CompliantReportingDenial` to `OpportunityDenial` and `CompliantReportingPreview` to `OpportunityPreview` to make them opportunity type agnostic, which in turn allows the client full profile to support multiple opportunity types. Also updates the updates the client update functions to include the opportunity type to handle multiple opportunity types.

## Type of change

## Related issues

> Closes #2099
> Closes #2172

### Development

### Code review


src/core/WorkflowsClientProfile/EarlyTerminationModule/EarlyTerminationModule.tsx
```diff
@@ -18,28 +18,19 @@
...
-    return "Printing PDF...";
-  }
-
-  if (client.updates?.earlyTermination?.completed) {
-    return "Reprint PDF";
-  }
-
-  return "Print PDF";
-};
+const OPPORTUNITY_TYPE = "earlyTermination";
```
jovergaag, line 33: Note: this component is nearly identical to `CompliantReportingModule`. I considered making a generic component to be used for both, but somehow if felt right to keep them separate for now until we know what the form is going to look like. If anyone thinks otherwise lmk and I can make the generic component.

src/core/WorkflowsClientProfile/EarlyTerminationModule/EarlyTerminationModule.tsx
```diff
@@ -65,11 +56,15 @@ export const EarlyTerminationModule = observer(
               kind="primary"
               shape="block"
               buttonFill={colors.buttonFill}
-              onClick={() => client.printReferralForm("earlyTermination")}
+              onClick={() => client.printReferralForm(OPPORTUNITY_TYPE)}
             >
-              {getPrintText(client)}
+              {getPrintText(client, OPPORTUNITY_TYPE)}
```
jovergaag, line 61: There is still an issue when printing the form - the print button gets stuck on "Printing...". The 
'COMPLETED' update is registered in firestore but I'm not sure whats going on in `WorkflowsStore` such that it is reflected here in the component. There is a [ticket](https://app.zenhub.com/workspaces/polaris-614c9f27fa19b10015ddd495/issues/recidiviz/pulse-dashboard/2158) to sort this out.

src/WorkflowsStore/__tests__/Client.test.ts
```diff
@@ -166,118 +167,145 @@ test("fetch client updates on demand", async () => {
...
-    client.recordId,
-    "compliantReporting",
-    true
-  );
-  expect(trackSetOpportunityStatus).toHaveBeenCalledWith({
-    clientId: client.pseudonymizedId,
-    status: "DENIED",
-    opportunityType: "compliantReporting",
-    deniedReasons: reasons,
+  OPPORTUNITY_TYPES.forEach(async (opportunityType) => {
```
macfarlandian, line 175: could make for easier debugging if you moved this into a `test.each` instead

src/core/WorkflowsClientProfile/Details.tsx
```diff
@@ -108,7 +111,7 @@ function getProbationSpecialConditionsMarkup(client: Client): JSX.Element {
   return (
     <>
       {!conditionsToDisplay.length &&
-        "None according to judgment orders in TOMIS"}
+        `None according to judgment orders in ${contentManagementSystem}`}
```
macfarlandian, line 114: i found this naming a little confusing ... should maybe be "caseManagementSystem" instead? 

Also I think "judgment orders" and "Board Actions" in these bits of copy are probably state-specific language for TN also ... something to flag for @sychang maybe to extend the copy template

src/core/WorkflowsClientProfile/EarlyTerminationModule/EarlyTerminationModule.tsx
macfarlandian: yeah I think this makes sense for now ...can revisit when we implement the next opportunity

src/firestore/firestore.ts
```diff
@@ -408,14 +408,15 @@ export const updateCompliantReportingDraft = async function (
...
-export async function updateCompliantReportingDenial(
+export async function updateOpportunityDenial(
   userEmail: string,
   clientId: string,
   recordId: string,
   fieldUpdates: {
     reasons?: string[];
     otherReason?: string;
   },
+  opportunityType: OpportunityType,
```
macfarlandian, line 419: love it ... I was implementing something similar as part of #2152 that will dovetail nicely with this

macfarlandian (changes requested):

src/core/WorkflowsClientProfile/Details.tsx
jovergaag: Oh yep, it is totally a case management system not content :) 

I have asked Serena and Lily for clarification on these strings in the copy doc. I'll update them when I hear back.

src/firestore/firestore.ts
```diff
@@ -431,8 +432,8 @@ export async function updateCompliantReportingDenial(
   );
 
   const changes: PartialWithFieldValue<ClientUpdateRecord> = {
-    compliantReporting: {
-      ...(oldDocument?.compliantReporting ?? {}),
+    [opportunityType]: {
+      ...(oldDocument?.[opportunityType] ?? {}),
```
daschi, line 436: I don't think we need this - this was just to be able to migrate the old compliantReporting updates.

jovergaag: When you say we don't need this, do you mean spreading the `oldDocument`? Is that migration already over and done with? I will delete line 436 if so.

daschi: No not yet, I think we will need to write a separate migration script to make sure we've copied over all the old documents and then delete them. We just don't need to replace `compliantReporting` with `opportunityType`, because we only need to migrate compliant reporting. Does that make sense?

jovergaag: I think there are some `earlyTermination` documents in the old client updates collection. I don't think its doing any harm to keep it referencing the correct opportunity type here, but let me know if I'm missing something.

src/core/WorkflowsClientProfile/Details.tsx
sychang: Just responded to the comment saying we can use "None according the DOCSTARS" for this case. 

I think we can reduce configuration complexity in generalizing this by just saying "None according to ${contentManagementSystem}" so it's one less thing that needs to be configured. We lose out on a tiny bit of information, but not a huge amount and it can be explained in the methodology.

src/core/WorkflowsClientProfile/OpportunityDenial.tsx
```diff
@@ -0,0 +1,236 @@
...
+      "Under active intermediate measure as a result of 1+ violations",
+    INCOMPLETE_CASE_PLAN_GOALS: "Has not completed case plan goals",
+    REVOCATION: "Under active revocation status",
+    SEX_OFFENSE: "Being supervised for sex offense",
+    NONPAYMENT: "Willfull nonpayment of fines/fees despite ability to pay",
+    OTHER_OFFENSE: "Incarcerated on another offense",
+    DECLINED: "State's Attorney permanently declined consideration",
+    [OTHER_KEY]: "Please specify a reason",
+  },
+};
```
Catacola, line 57: It feels to me like the `Opportunity` classes should define a method that returns these. Since the below component already gets the opportunity class, it can just call this method directly to get the allowed reasons. This will prevent us from needing to remember to update this component every time we add a new opportunity.

src/core/WorkflowsClientProfile/OpportunityPreview.tsx
```diff
@@ -72,47 +79,48 @@ const FormPreview = styled.img<Pick<StatusPalette, "border">>`
...
 
-export const CompliantReportingPreview = observer(
-  ({ client }: ClientProfileProps) => {
-    if (!client.opportunities.compliantReporting) return null;
+export const OpportunityPreview = observer(({ client }: ClientProfileProps) => {
+  // We will choose the first opportunity for now,
+  // until we have a design for how to preview multiple opportunities per client
+  const opportunity = values(client.opportunities).filter(
+    (opp) => opp !== undefined
+  )[0];
```
Catacola, line 87: If you just want the first match, you can user `arr.find()` instead of `arr.filter()[0]`. This has the advantage of both being more terse and faster (on average).

```suggestion
  const opportunity = values(client.opportunities).find(
    (opp) => opp !== undefined
  );
```

Catacola (changes requested):

src/core/WorkflowsClientProfile/Details.tsx
jovergaag: @sychang I just added a state specific copy object that allows us to use the state specific language, and it is structured in a way that it will be easy to pull from the copy doc whenever we get around to that. I think its fine to keep the extra detail in there!

src/core/WorkflowsClientProfile/OpportunityDenial.tsx
jovergaag: I was imagining that this copy would live in the copy doc at some point, and this component would just load the copy from the content system whenever we get around to that. But I suppose the opportunity could also load the copy. I'll update it to see what that looks like.

Catacola: Actually, I see that the `Opportunity` below is a record and not the opportunity class, but it has the `opportunityType`, so we should be able to fetch the class somehow.

src/WorkflowsStore/__tests__/Client.test.ts
jovergaag: Much nicer! Thank you.

src/core/WorkflowsClientProfile/Details.tsx
```diff
@@ -53,6 +54,22 @@ const SpecialConditionsCopy = styled.div`
   ${typography.Body12}
 `;
 
+type SupervisionTypeCopy = {
+  parole: string;
+  probation: string;
+};
```
macfarlandian, line 60: i like the pattern here but i think the names are maybe too generic ... something should indicate that these strings are specifically for special conditions empty states

src/core/WorkflowsClientProfile/OpportunityPreview.tsx
```diff
@@ -72,47 +79,48 @@ const FormPreview = styled.img<Pick<StatusPalette, "border">>`
...
-export const CompliantReportingPreview = observer(
-  ({ client }: ClientProfileProps) => {
-    if (!client.opportunities.compliantReporting) return null;
+export const OpportunityPreview = observer(({ client }: ClientProfileProps) => {
+  // We will choose the first opportunity for now,
+  // until we have a design for how to preview multiple opportunities per client
+  const opportunity = values(client.opportunities).find(
+    (opp) => opp !== undefined
+  );
+  if (!opportunity) return null;
```
macfarlandian, line 88: looking at this again I feel like this logic should move up to the client profile and this component should just take an `Opportunity` as a prop? That way at least we have one generic component and the temporary logic is not spread across two different locations

jovergaag:
@daschi @Catacola This is ready for another look!

src/WorkflowsStore/Opportunity/EarlyTerminationOpportunity.ts
```diff
@@ -36,6 +38,17 @@ import {
   rankByReviewStatus,
 } from "./utils";
 
+const DENIAL_REASONS_MAP = {
+  VIOLATIONS: "Under active intermediate measure as a result of 1+ violations",
+  INCOMPLETE_CASE_PLAN_GOALS: "Has not completed case plan goals",
```
daschi, line 43: Is this how it should appear in the side bar, or is this just a placeholder till we have the copy?

daschi (approved):
looks good to me ⭐

src/WorkflowsStore/Opportunity/EarlyTerminationOpportunity.ts
jovergaag: The copy itself is final - but the codes are just made up. What do you mean by side bar?

daschi: Oh because it looks like the keys in this map are what we're displaying on the frontend in the denial reasons drop down when you select one, just wondering if there is a different format it should be?

jovergaag: Oh yes, I had seen that a while ago and forgotten about it. I'm not sure what to do about that. 

@sychang Thoughts on what code we should display in the drop down when a client is marked as ineligible for ND? The codes currently being displayed are ones that I just made up and are the following:

Screen Shot 2022-08-12 at 10 10 32 AM

Which show up like

Screen Shot 2022-08-12 at 10 08 55 AM

jovergaag:
@Catacola Would love a re-review from you on this today, I would like to get it merged and deployed so we can start getting some `earlyTermination` events flowing into prod views to develop the impact metrics against.

I think I would like to merge this as is without updating the codes (discussion above), and wait until next week when Serena is back to update the codes in the dropdown.

Catacola (approved):

macfarlandian (approved):
🙌🏻

PR #2160, opened 2022-08-10
[Workflows] move opportunity updates into subcollections

### LINKED ISSUE #2152: [Workflows] Move clientUpdates into subcollections per tenant

macfarlandian:
## Description of the change

Implements yet another in-place migration to break the `ClientUpdates` document into subcollections. 

Highlights: 

- Migrates the update document upon first read rather than first write, which simplifies everything downstream considerably (the Client no longer knows about the various different data sources, it is all encapsulated in the Firestore subscribe function)
- Subscriptions for each opportunity type are enumerated in a mapping; while this is less generic and scalable in the short term, I thought it would actually make it easier to do the next step of refactoring, which would be to move each subscription along with all form data handling into the respective Opportunity classes. (It seemed tricky to do this partially, given the interdependencies between the two, and I didn't want to add complexity to this PR by doing that all now.)
- the new `ClientUpdate` documents themselves will not actually contain any fields anymore; we can clean that up when data is fully migrated

Pretty difficult to test all of the subscription behavior reliably in unit tests but I have verified this in offline mode as well as against staging data (migrating successfully from both legacy data sources, writing changes to the new collections) ... the existing Firestore security rules already covered this use case but we should take care that remains true when we update them for multi-tenancy

## Type of change

## Related issues

> Closes #2152 

### Development

### Code review


src/firestore/firestore.ts
```diff
@@ -258,6 +258,75 @@ export function subscribeToClientUpdatesV2(
...
+export function subscribeToOpportunityUpdate<
+  UpdateType extends OpportunityUpdateRecord
+>(
+  clientId: string,
+  clientRecordId: string,
+  opportunityType: OpportunityType,
+  handleResults: (results?: UpdateType) => void
+): Unsubscribe {
+  const updateDocRef = doc(
+    collections.clientUpdatesV2,
```
daschi, line 311: I'm curious about the new path, are we preparing for other possible subcollections a client could have alongside `clientOpportunityUpdate`? Or is this structure more so we can group all the opportunities in a subcollection?

I'm wondering about clients who might have multiple forms filled out for them, which I think might become the case for early termination in ND, but I think this works because each client/form combo will have a unique recordId?

src/firestore/firestore.ts
```diff
@@ -382,35 +451,34 @@ const getClientUpdatesV2DocRef = async function (
   return { docRef, oldDocument };
 };
 
+async function updateOpportunity(
```
daschi, line 454: 🥳

daschi (approved):

src/firestore/firestore.ts
macfarlandian: I guess the thinking was that a separate subcollection per opportunity type would be cumbersome, since they would each only have one document in them? Instead of creating duplicate records when there are multiple forms or whatever other such situations might arise related to opportunities, we would encapsulate all of that within the structure of a single document. There isn't any particular reason that different types of opportunity documents need to share structure with one another (nor is it a realistic goal IMO given all the state-specificity that goes into them) ...  since each one is backed by a separate class implementation we can just deal with all the mess there to support the generic Opportunity interface.

I guess it's plausible that there could be other subcollections besides opportunities but there wasn't anything specific I had in mind ... it's a pattern to consider as an alternative to adding lots of conditional or state-specific properties to the main client document

src/firestore/firestore.ts
```diff
@@ -258,6 +258,75 @@ export function subscribeToClientUpdatesV2(
...
+    if (!v1UpdatesDoc) {
+      legacyRecord = (await getDoc(v2UpdatesDocRef)).data();
+    }
+
+    dataToMigrate = legacyRecord?.[opportunityType];
+  }
+
+  if (dataToMigrate) {
+    // write old + new data to new destination
+    setDoc(updateDocRef, { ...dataToMigrate });
```
Catacola, line 293: I'm not sure if you're making a copy on purpose, but wanted to flag this just in case.

```suggestion
    setDoc(updateDocRef, dataToMigrate);
```

src/firestore/firestore.ts
```diff
@@ -425,58 +493,36 @@ export async function updateCompliantReportingDenial(
...
-    compliantReporting: {
-      ...(oldDocument?.compliantReporting ?? {}),
-      denial: {
-        ...filteredUpdates,
-        ...fieldsToDelete,
-        updated: {
-          by: userEmail,
-          date: serverTimestamp(),
-        },
+  const changes = {
```
jovergaag, line 496: Oh now nice!

src/WorkflowsStore/Client.ts
```diff
@@ -213,35 +215,45 @@ export class Client {
...
-        if (r) {
-          runInAction(() => {
-            handler(r);
-            const data = r.compliantReporting?.referralForm?.data ?? {};
-            set(this.compliantReportingReferralDraftData, data);
-          });
-        } else {
-          // empty object will replace undefined, signifying completed fetch
-          handler({});
+    set(this.opportunityUpdateSubscriptions, {
```
jovergaag, line 218: So much better now. Thank you.

jovergaag (approved):

src/WorkflowsStore/Opportunity/CompliantReportingOpportunity.ts
```diff
@@ -316,8 +316,10 @@ class CompliantReportingOpportunity implements Opportunity {
...
       if (updates.completed) {
         status = "COMPLETED";
-      } else {
+      } else if (updates.referralForm) {
         status = "IN_PROGRESS";
+      } else {
+        status = "PENDING";
       }
     } else {
       status = "PENDING";
```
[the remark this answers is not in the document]
macfarlandian: wow so much better ... what a mess

PR #2166, opened 2022-08-11
[Workflows] Remove mobx warnings from client test

### LINKED ISSUE #2155: [Workflows] Clean up mobx warnings in Client.test.ts

daschi:
## Description of the change

Removes mobx warnings from client test by adding actions/autorun.  

## Type of change

## Related issues

> Closes #2155

### Development

### Code review


src/WorkflowsStore/Client.ts
```diff
@@ -268,17 +269,19 @@ export class Client {
...
-        record.compliantReportingEligible,
-        this
-      ),
-      earlyTermination: createEarlyTerminationOpportunity(
-        record.earlyTerminationEligible,
-        this.fetchedEarlyTerminationReferral.current(),
-        this
-      ),
-    };
+    autorun(() => {
```
macfarlandian, line 272: we have a testUtils funtion called `reactImmediately` that will do this and immediately dispose of the autorun to prevent memory leaks ... just been lazy about using it 

I feel like we could probably just disable these lint warnings in the test environment also? I don't think it really matters that much if we do this or not unless we are testing something that specifically depends on some mobx functionality. Generally speaking I think accessing these values outside of a reaction shouldn't be _broken_, it may just be less efficient or not fully reactive ... and it makes tests less annoying to write 😎

daschi: This was also causing a warning in the console....do you mean it's prob okay to ignore the warning in the console and I should remove this `autorun` here, and just silence the test warnings?

jovergaag: I agree that its probably ok to have the warnings...we should just silence them in the tests because the test output has gotten to be SUPER noisy and hard to find the stuff that really matters. That was essentially what the ticket was for, just to quiet the warnings, whether by fixing what is causing the warning, or ignoring them in tests is fine with me!

daschi: Ok! I just added a config to the `setupTests` to remove all warnings...not sure if this will cause issues down the line but it at least didnt' break any tests 🤞 also fixed some of the other warnings

jovergaag (approved):
Oh that's much nicer! Thank you. 

There is a linting error, but looks good after that.

src/WorkflowsStore/Client.ts
macfarlandian: yes that is what i meant ... I think a reasonable policy going forward would be to be vigilant about warnings that we see in the dev environment and to generally disregard them in the test environment. Maybe we can selectively turn them on sometimes for debugging but agreed with Jen that they just generate too much noise which probably contributes to warning fatigue. 

(The [MobX docs](https://mobx.js.org/configuration.html#linting-options) suggest that the linting feature is mainly intended as "traning wheels" rather than a permanent diagnostic tool ... we could also consider adopting the [ESLint plugin](https://github.com/mobxjs/mobx/blob/main/packages/eslint-plugin-mobx/README.md) as a replacement/supplement)

daschi: Hmm..it looks like the warnings still show up when I remove the `autorun` even with that `configureMobx` set to never in `setupTests`...I can keep digging into why that isn't working, or just merge this as is for now?

macfarlandian: ah sorry did you mean "in the console" as in, in the browser console when you run the app? looking at this again I realized that this change was in the actual Client object and not the test file so I think everything I said was irrelevant 😎 ... I actually don't yet understand why this was throwing a warning but I'm pretty sure `autorun` is the wrong solution here

daschi: Oh oops...sorry I merged too soon. Why is `autorun` the wrong solution?

PR #2251, opened 2022-08-22
Fix offline mode

jovergaag:
## Description of the change
When we were working to fix the cross view security issue uncovered by the pentesters, the api response for offline users was broken. This commit fixes that by making requests from offline users 'allowed'.

## Type of change

## Related issues

> Closes [#XXXX]

### Development

### Code review


server/routes/api.js
```diff
@@ -198,6 +198,7 @@ function vitals(req, res) {
   const appMetadata = getAppMetadata(req);
   const allowed =
     appMetadata.state_code === "recidiviz" ||
+    isOfflineMode ||
```
jovergaag, line 201: Since w don't send a token to the server in offline mode, the app_metadata with the permissions is not available in the server. We should skip the permissions check for offline mode.

src/core/models/backendDiff/Differ.ts
```diff
@@ -48,7 +48,7 @@ export abstract class Differ<DataRecord, DiffValueType> {
     });
 
     const newResults = new Map<string, DiffValueType>();
-    newData.forEach((data) => {
+    newData?.forEach((data) => {
```
jovergaag, line 51: @danawillow To get Ilya up and running in offline mode, I had to guard for undefined here for the new BE response. Are you relying on this to throw an error in cases when there is no data from the new BE or is this safe?

server/routes/__tests__/server.test.js
```diff
@@ -287,13 +287,13 @@ describe("Server tests", () => {
...
 
-    it("responds with a formatted error resposne", () => {
+    it("responds with a formatted error response", () => {
       return request(app)
         .get("/api/US_MO/newRevocations/revocations_matrix_events_by_month")
         .then((response) => {
           expect(response.statusCode).toEqual(500);
           expect(response.body.errors).toEqual([
-            "No authorization token was found",
+            "Cannot read properties of undefined (reading 'flattenedValueMatrix')",
```
jovergaag, line 296: This is a different error than before, but the test is just checking that there is a formatted error so this feels safe to me.

server/app.js
```diff
@@ -144,7 +144,7 @@ function errorHandler(err, _req, res, next) {
   }
 }
 
-if (!isOfflineMode) {
+if (!isOfflineMode && !authEnv === "test") {
```
jovergaag, line 147: In order to test the permissions in the API - we have to run tests with `isOfflineMode === false` so in this case, we also don't want to check the jwt in the test auth environment.

src/core/models/backendDiff/Differ.ts
danawillow: This is safe! If this is undefined (and `oldData` isn't empty), then we'll just log a diff. The actual check for whether the call to the new BE succeeded is in `metricsClient.js`

danawillow (approved):
Thanks for doing this!

jovergaag:
@nasaownsky I won't have time to get these tests fixed and merged by the end of my day, but if you rebase on this branch, it should get you up and running in offline mode!

PR #2254, opened 2022-08-23
[Workflows] Allow officers from all districts for US_ND

### LINKED ISSUE #2213: [US_ND] [Workflows] Disable district-level officer filter for entire state
If a user is marked as having a caseload and being assigned to a district, we currently restrict their list of officers to the other officers in their district.

This feature is not desired in ND.

We should add a flag in the `TenantConfig` objects to control whether this district-level filter is enabled on a state-by-state basis. Feel free to propose other ways to configure this.

daschi:
## Description of the change

This adds a flag to US_ND tenant `workflowsEnableAllDistricts` which will enable officers from all districts in the officer dropdown and caseloads lists. 

Another idea I had was to add a  `workflowsConfiguration` property to tenant config that's an array of strings like ["allowAllDistricts"], in the case that we think workflows will have more of these very specific configurations. 

The way it's currently set up means any new tenant will automatically be restricted until `workflowsEnableAllDistricts` is added to the config.

## Type of change

## Related issues

> Closes #2213

### Development

### Code review


src/RootStore/__tests__/TenantStore.test.ts
```diff
@@ -119,6 +130,51 @@ describe("TenantStore", () => {
...
+      jest.clearAllMocks();
+      jest.resetAllMocks();
+    });
+
+    afterEach(() => {
+      jest.clearAllMocks();
+      jest.resetAllMocks();
+    });
+
+    it("returns true when a current tenant enables viewing all districts", () => {
```
daschi, line 144: I'm kinda stuck trying to figure out why this test passes when run alone but fails as a unit. It prints out US_MO as the current tenant ID, even though it should use the new mock rootStore? @macfarlandian @jovergaag have either of you dealt with this before?

macfarlandian: weird ... usually that would be a case of state leakage between tests (e.g. a jest mock not being reset) but I don't see where that could be happening here. I'm stumped!

Catacola:
This looks good! I like the idea of a list of options we could explicitly enable, but let's wait until we have 2 or 3 of them before building something more complex than just a boolean like this.

And I'm with Ian on the error.

src/RootStore/__tests__/TenantStore.test.ts
macfarlandian: was digging some more this morning and noticed that if I run just the local describe block this test passes and the second one fails. I think possibly the state is leaking via session storage which the tenant store reads from [here](https://github.com/Recidiviz/pulse-dashboard/blob/daschi-2213-district-filter/src/RootStore/TenantStore/TenantStore.ts#L36) so probably you need to clear it between tests?

daschi: I found it! I needed to clear session storage

daschi: Oh nice! I didn't see your comment before posting mine 😆

macfarlandian: great minds

macfarlandian (approved):

PR #2284, opened 2022-08-26
[WIP] Subscription class exploration

macfarlandian:
## Description of the change

This is by no means a merge candidate, just a way to test out the ergonomics of a new subscription interface and get some feedback on whether this is worth pursuing. 

I started with probably the most trivial example, subscribing to the list of officers in your state after the WorkflowsStore is hydrated, but it does demonstrate some interesting stuff: 

- a mostly generic subscription interface that encapsulates the Firestore and MobX weirdness and gives us more control over the subscription's lifecycle (subscribe and unsubscribe handles, a persistent `data` observable that doesn't get wiped out automatically when we create a new subscription) and replicates the auto-subscribe functionality of `mobx-utils.fromResource` ... I didn't follow this too far but I imagine it could be factored out to separate the boilerplate from the specific (which is mainly just how it constructs the query or whatever it's going to watch)
- the subscription functionality is more testable but also easier to isolate and mock in tests that don't really care about it
- it can directly respond to other observables in the datastore, which could mean less need to manually create and destroy them in `WorkflowsStore`, which is one of the more cumbersome aspects of the current setup. I also didn't follow this too far but you could imagine, e.g., reactively stopping/starting the subscription when the tenant ID changes, and extrapolate that to other cases more relevant to user-land
- alternatively, if you look at just the first commit you can see an example of how this would work without directly hooking into the rest of the datastore (taking arguments in the constructor, and presumably we could change those values on the fly and also have the underlying subscription react to that)

Curious to hear what people think! If this seems promising the next step would probably be to scope out the full change and maybe create a design doc

## Type of change

## Related issues

n/a

### Development

### Code review


src/WorkflowsStore/subscriptions/OfficerSubscription.ts
```diff
@@ -0,0 +1,82 @@
...
+
+  private cancelSnapshotListener?: Unsubscribe;
+
+  constructor(rootStore: RootStore) {
+    this.rootStore = rootStore;
+
+    makeAutoObservable(this);
+
+    onBecomeObserved(this, "data", () => this.subscribe());
+    onBecomeUnobserved(this, "data", () => this.unsubscribe());
```
jovergaag, line 42: Super cool, I didn't know this existed!

src/WorkflowsStore/__tests__/WorkflowsStore.test.ts
```diff
@@ -244,47 +244,30 @@ test("receive feature variants from subscription", async () => {
...
-  testObserver = keepAlive(computed(() => workflowsStore.availableOfficers));
+  expect(workflowsStore.districtFilter).toBe(mockOfficer.info.district);
+});
 
-  await when(() => workflowsStore.availableOfficers.length > 0);
+test("officers from subscription", async () => {
+  await waitForHydration();
 
-  expect(mockSubscribeToOfficers).toHaveBeenCalled();
+  OfficerSubscriptionMock.mock.instances[0].data = mockOfficers;
```
daschi, line 270: 🙌

daschi (commented):
I think it's great, and the testing will be easier! I'm not sure if we lose anything by not using `fromResource`, but I think the `onBecomeObserved` pattern makes sense. 

Re: making it reactive, would that come from adding `currentTenantId` as a computed value in the subscription class?

macfarlandian:
> Re: making it reactive, would that come from adding `currentTenantId` as a computed value in the subscription class?

Yes I didn't go so far as to write a test to demonstrate this, but that's how it would work in this case. It would probably need a little additional logic to figure out if there was an active subscription that needed to be replaced. But the general idea is that we can make the subscriptions themselves more context-aware rather than requiring some other class to manage their lifecycles

PR #2296, opened 2022-08-30
[Workflows] Add EarnedDischargeOpportunity skeleton and routes

### LINKED ISSUE #2227: [Workflows] Create `EarlyDischargeOpportunity.ts`

daschi:
## Description of the change

This is a suuuuper skinny skeleton of the `EarnedDischargeOpportunity`, mostly because all of the subscription stuff will be changing soon. Let me know if anything seems like we won't need it, I moved the form properties to form-specific types for the opportunity since this opportunity doesn't seem to have the same type of form data. 

## Type of change

## Related issues

> Closes #2227

### Development

### Code review


src/WorkflowsStore/Opportunity/types.ts
```diff
@@ -23,13 +23,14 @@ import {
 } from "./EarlyTerminationReferralRecord";
 
 export type OpportunityCriterion = {
-  // TODO: text as templates?
   tooltip?: string;
+  text?: string;
```
macfarlandian, line 27: not sure this should really be optional in practice?

src/WorkflowsStore/Opportunity/EarnedDischargeOpportunity.ts
```diff
@@ -0,0 +1,160 @@
...
+
+// This could be configured externally once it's fleshed out
+// to include all copy and other static data
+// TODO: Update the keys in this mapping once we know what the data looks like
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+const CRITERIA: Record<string, OpportunityCriterion> = {
+  minimumSentenceServed: {
+    tooltip:
+      "If on probation, served minimum sentence according to the court; if on parole for a nonviolent crime, served at least one year; if on parole for a sex/violent offense, served at least one-third of remaining sentence; if on parole for a life sentence, served at least five years on parole",
+  },
```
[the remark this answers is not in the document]
daschi: I left it off because it will be interpolated with sentence length, and I think the pattern we did for ET added that in the `requirementsMet` method ( i.e. `Served ${sentence length served}`)

src/WorkflowsStore/Opportunity/EarnedDischargeReferralRecord.ts
```diff
@@ -0,0 +1,49 @@
...
+
+type EarnedDischargeReason = {
+  minimumSentenceServed: string;
+  noNewOffenses: string;
+  courtOrderConditionsMet: string;
+};
+
+export type EarnedDischargeCriteria = {
+  criteriaName: string;
+  reason: EarnedDischargeReason;
```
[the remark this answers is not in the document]
daschi: Hmm...I used the word reason because that's what the record uses, curious what other part of the codebase is using reason in another way?

src/WorkflowsStore/Opportunity/types.ts
daschi: true - it's only really optional because of how it's constructed, half with the `CRITERIA` object and half with the `requirementsMet` method. Now that I'm looking more closely it's the exact same type as `OpportunityRequirment` on line 48. We should probably just have one of these..

src/WorkflowsStore/Opportunity/EarnedDischargeReferralRecord.ts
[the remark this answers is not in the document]
macfarlandian: there is a `reasons` field in the opportunity updates that holds the denial reasons .... this was something @jovergaag flagged to me last week actually, that it's confusing to use it in both places and one of them should migrate away from it. Personally I think "reasons" makes more sense for the denials because it's an expression of the user's intent, but I'm sure that's arguable.

I don't know if "requirements" is really right here either, elsewhere in the data models we distinguish somewhat arbitrarily between "requirements" and "criteria" ... e.g. the record for Compliant Reporting calls them "criteria" whereas the Opportunity interface itself has "requirements" (which is the generic form that's just a list of text items to display). That is not obviously better really since it's not a meaningful distinction? Here and in ND we call them "reasons" and "requirements", which is equally meaningless. 🙃 Any suggestions? Thinking about this is giving me brain fog

src/WorkflowsStore/Opportunity/types.ts
macfarlandian: yeah i think I was getting the two mixed up, I know this type was originally created as a partial refactor so I just want to make sure it is continuing in a sensible direction

src/WorkflowsStore/Opportunity/EarnedDischargeReferralRecord.ts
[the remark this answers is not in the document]
daschi: Hmm...I see what you mean. Trying to think of another word for "reason they are meeting criteria", how about `eligibilityReason`? I know reason is still there but maybe it will help differentiate it from denialReason. I'm not sure how far from the terminology in the exported data we would want to go. I could see these also being called `requirement` or `metRequirement` though.

src/WorkflowsStore/Opportunity/EarnedDischargeReferralRecord.ts
```diff
@@ -0,0 +1,49 @@
...
+};
+
+export type EarnedDischargeReferralRecord = {
+  stateCode?: string;
+  externalId?: string;
+  formInformation?: {
+    clientName: string;
+  };
+  reasons?: EarnedDischargeCriteria[];
```
macfarlandian, line 35: I think something that has contributed to my confusion around 'reasons" versus "criteria" etc is that I don't really understand this data structure? How is there going to be an array of `Criteria` that all contain the same mapping of `reasons`? I guess I would expect the three fields in `EarnedDischargeReason` to just be at the top level of the record rather than being repeated. 

(also we are structuring this as `reason -> criteria -> reasons` which does not help either ... maybe the one at this level should be called `criteria`)

daschi: yeah I think having `criteria` at the top level would make more sense. Something like:

`criteria: { name: "", criteriaMet: { minimumSentenceServed: any } }`

But I would imagine this would be a backend change - is that something we want to try to bring up as ED is being built out? I assume we would want to also have it match in CR and ET.

daschi: `name` would be the type of criteria (i.e. `SUPERVISION_EARLY_DISCHARGE_DATE_WITHIN_30_DAYS`), which is being used to build the reasons structure in `transformedRecord`

PR #2298, opened 2022-08-30
[Workflows][Fast-follow] Add the edit and collab and date status to early term form

### LINKED ISSUE #2291: [Workflows] Add the "Edit and collaborate" text and default timestamp at the top of the form
In TN, there is text on the top left of the form page that says "Edit and collaborate on the document below." And by default, it shows when the data was prefilled by TDOC data. If a user updates the form, that timestamp updates.

In ND, the top of the form does not have that text on the left. And it has no default timestamp on the right that shows when the data was prefilled. Can we add both of those to the top of the form to match between the states?

daschi:
## Description of the change

This adds the Edit and collaborate form viewer status to the early termination form. It also updates the "edited on" status to match what we have for CR. It's using the current date until we have the `dateToday` field added to the ET query.

I also updated the CR transformer because I think it was using `new Date()` instead of the `dateToday` field. Let me know if that shouldn't be the case!

## Type of change

## Related issues

> Closes #2291 

### Development

### Code review


src/core/Paperwork/US_TN/Transformer.ts
```diff
@@ -49,7 +50,7 @@ export const transform = (
   return {
     ...data,
     allDockets,
-    dateToday: formatWorkflowsDate(new Date()),
+    dateToday: formatWorkflowsDate(parse(data.dateToday, "y-M-d", new Date())),
```
macfarlandian, line 53: I have never quite understood the utility of this field, to be honest. Why do we need the ETL data to tell us what today's date is? Especially given that it is more likely to be wrong than `new Date()` ... personally I don't think we should use it here and I don't think we should add it to any new places either. kill it 🪦

daschi: hahaha okay i can see that. I think I was seeing it as a "last updated date" that would tell us the freshness of the prefill data, but if it's always going to be today then yeah I guess we don't need it!

macfarlandian: there would be potential value in that ... but in this case it is literally used to fill in a field that says "today's date" lol

src/core/WorkflowsCompliantReportingForm/WorkflowsCompliantReportingForm.tsx
```diff
@@ -57,9 +56,9 @@ const WorkflowsCompliantReportingForm: React.FC = () => {
       draft.updated.date.seconds * 1000
     ).fromNow()}`;
   } else {
-    lastEdited = `Prefilled with data from TDOC on ${
-      compliantReporting?.formData.dateToday ?? moment().format("MM-DD-YYYY")
-    }`;
+    lastEdited = `Prefilled with data from TDOC on ${moment().format(
+      "MM-DD-YYYY"
```
macfarlandian, line 60: huh I didn't realize this field was actually doing double duty as as today's date and the last export date ...  this is still kind of bogus honestly but fixing that is probably not in scope here since it was already like this. If we wanted to start adding an export date field to all opportunity records to support this display I would be for that, just please don't call it `dateToday` lol

macfarlandian (approved):

src/core/WorkflowsCompliantReportingForm/WorkflowsCompliantReportingForm.tsx
daschi: hahah...sounds good to me. I just closed an issue to add not-date-today-but-also-still-today field but I can re-open it and update the requirements so we can add it to all new opportunities. Probably as `lastUpdatedDate` or something

